PR #410, opened 2013-11-06
Update kaltura.conf.template

octycho:
Setting VirtualHost directive to default,because there is a duplicate virtualhost in the included files, therefore the apache server does not function properly.


jessp01:
Hello @octycho,

This was fixed a while back. We could not merge your contrib because you have not signed the CLA.
We would love for you to sign it and submit additional contribs in the future.

Thanks,

PR #1267, opened 2014-05-29
Panda flavor profiles add page

dudyk:
A page to add flavor profiles, it's added as part of the internal tools admin console plugin.
It supports adding to specific partners, and as the default partner (when set to 0).
Thanks @jessp01 for spotting my bug.


dudyk:
I signed the CLA

jessp01:
@dudyk, indeed signed. The status has not changed to 'success' because there is a 1000 status update limitation enforced by GitHub's API. The original github hooks code would change status to 'pending' each time it ran and so, since it took a relatively long time before you signed, it exceeded 1000. I have later changed to code to check status prior to trying to set it so that, if it is already in 'pending' it will not be reset.

@tan-tan-kanarek, please review the pull so it can be merged.

Thanks,
Alas, there is nothing I can do to set this status to 'success' now

tan-tan-kanarek:
Merged in https://github.com/kaltura/server/pull/2927

Issue #1812, opened 2014-10-24
Unicode Improperly Truncated during bulk upload - Makes XML reply not well formed


flipmcf:
we just uploaded a bunch of media entries via CSV file from the KMC.
One of these entries has "Bad Content" - the Title is over 256 bytes.
The Title is also in unicode.

When the Title is truncated to 256 Bytes, the truncation happens in the middle of a unicode character.

Here's the CSV entry:

* title,description,url,contentType,category
"លទ្ធផលរបាយការណ៍ពីស្ថានភាពសិទ្ធិមនុស្សកម្ពុជាក្នុងសន្និសីទនៃអង្គការសហប្រជាជាតិ (វគ្គ បញ្ចប់)-H100114RSV3-WED","",http://url.to/thevideo.mp4,video,"Khmer>1",,,,,,,

The resulting KalturaMediaListResponse contains:

[...]

PR #2029, opened 2014-12-16
 Re-pull #1061 into Jupiter-10.1.0 + #1813 #1823 

flipmcf:
First 4 commits are to bring Jupiter-10.1.0 up-to-date with 9.14.0: See pull #1061

36b6ee0 and 75216a1 are for ticket #1812 (re-opening Pull #1813)


erankor:
Hi @flipmcf,

My preference is not to strip invalid utf8 characters in the client, I would prefer to fix it on the server side. Currently, the server is validating / stripping the incoming utf8 strings, however the strings can indeed become corrupt in case they get truncated by the database. We should fix it at some point, but it's currently not high priority.
Please remove the stripping of bad utf8 characters from the response and reopen against the default branch.

Thanks !

Issue #1926, opened 2014-11-26
thumbasset generate action fails when file on remote storage and no local copy exists


lelikg:
When exporting to remote storage + deleting local file - trying to create a thumbnail via thumbasset generate action outputs the following error:

<?xml version="1.0" encoding="utf-8"?><xml><result><error><code>SOURCE_FILE_REMOTE</code><message>The source file is remote, no local file found</message><objectType>KalturaAPIException</objectType><args></args></error></result><executionTime>0.091269016265869</executionTime></xml>

jessp01:
Se comment here: https://github.com/kaltura/server/pull/2037#issuecomment-69490178

PR #2037, opened 2014-12-19
Generate thumbnail for LOCAL Remote Storages

visomar:
As it is now, Kaltura can only generate thumbnails for entries which are in the main DC, i.e. using the default storage space that comes with kaltura (/web/content/entry/data...).
In order to generate thumbnails for entries which are stored in a remote storage, there has to be a flavor that remains in that main DC.
Nevertheless, if you are using a Remote Storage configured as LOCAL, you can access locally to it, so I've made this modification which checks if this is the case, and then builds the source path for the video to generate correctly the thumbnail.

### DIFF
--- a/alpha/apps/kaltura/lib/batch2/bcdl/kBusinessPreConvertDL.php
+++ b/alpha/apps/kaltura/lib/batch2/bcdl/kBusinessPreConvertDL.php
@@ -279,11 +279,18 @@ class kBusinessPreConvertDL
 		
 		if(!$fileSync || $fileSync->getFileType() == FileSync::FILE_SYNC_FILE_TYPE_URL)
 		{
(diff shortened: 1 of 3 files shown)


erankor:
@visomar, thanks for you contribution !
We have decided not to merge this change at this point since we believe it represents a wrong usage of our system. 
As we discussed in https://github.com/kaltura/server/issues/1926#issuecomment-69024731, we believe that remote storage should not be used when the files are accessible via NFS from the Kaltura server. When the primary storage is NFS-mounted to the Kaltura server it should be defined as the storage root (/web by default). When remote storage is avoided, the system does not to copy the files after they are transcoded (they can simply be moved to their final location), and also, the source files can be used for transcoding additional flavors without the need to import them first.

PR #2108, opened 2015-01-05
PLAT-2289 Live to VOD offset fix

AviBueno:
1. Added code to copy cuepoints from live to VOD within the barriers of the VOD duration.
2. Changed lastElapsedRecordingTime to lastElapsedLiveDuration and made lastElapsedRecordingTime deprecated.
3. Accumulating recorded entry duration in appendRecording via recordedLengthInMsec
4. Added kLiveRecordingSegmentInfo
5. Setting cuepoint status to HANDLED immediately if recording is disabled, otherwise, after copying cuepoints to VOD entry.
6. Copying only cuepoints with READY state (skipping already HANDLED cuepoints), which replaces the maxCopiedVodCuePointStartTime condition.

### DIFF
--- a/api_v3/lib/types/entry/KalturaLiveEntry.php
+++ b/api_v3/lib/types/entry/KalturaLiveEntry.php
@@ -36,15 +36,8 @@ abstract class KalturaLiveEntry extends KalturaMediaEntry
 	/**
 	 * Elapsed recording time (in msec) up to the point where the live stream was last stopped (unpublished).
 	 * @var int
-	 * @deprecated
 	 */
 	public $lastElapsedRecordingTime;
-
-	/**
-	 * Elapsed live streaming time (in msec) up to the point where the live stream was last stopped (unpublished).
-	 * @var int
-	 */
-	public $lastElapsedLiveDuration;
 
 	/**
 	 * Array of key value protocol->live stream url objects
--- a/api_v3/lib/types/entry/KalturaLiveEntry.php
+++ b/api_v3/lib/types/entry/KalturaLiveEntry.php
@@ -36,8 +36,15 @@ abstract class KalturaLiveEntry extends KalturaMediaEntry
 	/**
 	 * Elapsed recording time (in msec) up to the point where the live stream was last stopped (unpublished).
 	 * @var int
+	 * @deprecated
 	 */
 	public $lastElapsedRecordingTime;
+
+	/**
+	 * Elapsed live streaming time (in msec) up to the point where the live stream was last stopped (unpublished).
+	 * @var int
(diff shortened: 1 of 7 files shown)


plugins/cue_points/thumb_cue_point/model/ThumbCuePoint.php
```diff
@@ -36,10 +36,44 @@ public function getMetadataObjectType()
...
+					return $vodToLiveDeltaTime;
+				}
+				else
+				{
+					return null;
+				}
+			}
+			else
+			{
+				$vodToLiveDeltaTime = $liveRecordingSegmentInfo->getVodToLiveDeltaTime();
```
tan-tan-kanarek: @AviBueno, shouldn't it by +=?

api_v3/lib/types/entry/KalturaLiveEntry.php
```diff
@@ -36,10 +36,17 @@
 	/**
 	 * Elapsed recording time (in msec) up to the point where the live stream was last stopped (unpublished).
 	 * @var int
+	 * @deprecated
 	 */
 	public $lastElapsedRecordingTime;
```
tan-tan-kanarek: @AviBueno, I think this field is used internally by Kaltura servers and you can remove it and break compatibility, it won't affect anyone.

plugins/cue_points/thumb_cue_point/model/ThumbCuePoint.php
AviBueno: No, every value reflects the total drift from the beginning of the stream.

api_v3/lib/types/entry/KalturaLiveEntry.php
```diff
@@ -36,10 +36,17 @@
...
 
 	/**
+	 * Elapsed live streaming time (in msec) up to the point where the live stream was last stopped (unpublished).
+	 * @var int
+	 */
+	public $lastElapsedLiveDuration;
```
tan-tan-kanarek: @AviBueno, are you going to set this attribute from the media-server? Why can't you use the duration as passed to the append recording?

AviBueno: lastElapsedLiveDuration replaces lastElapsedRecordingTime, which is set on the server in KalturaLiveEntryService::appendRecording(). 
This var is the live stream duration anchor, which is set upon a stop-stream (unpublish) event and will not change during a streaming session even when appendRecording is called periodically. This is opposed to LiveEntry::msDuration, which is changed each call to appendRecording via setLengthInMsec.

PR #2110, opened 2015-01-06
Jupiter 10.2.0 sphinx plays views script

shpiNi:


shpiNi:
@etameran please double-check me before merging - 
1. removed the partnerId from input parameters.
2. made the reset mode as default.

PR #2128, opened 2015-01-13
add source tag when import finished

shpiNi:


alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -38,6 +38,8 @@ public static function createOriginalFlavorAsset($partnerId, $entryId, &$msg = n
 		$flavorAsset = flavorAsset::getInstance();
 		$flavorAsset->setStatus(flavorAsset::FLAVOR_ASSET_STATUS_QUEUED);
 		$flavorAsset->incrementVersion();
+		if ($shouldAddSourceTag)
```
tan-tan-kanarek: @shpiNi, please remove the condition and add the SOURCE tag in all cases.

PR #2172, opened 2015-01-27
Jupiter 10.4.0 sup 3451

shpiNi:


tan-tan-kanarek:
We’re not supposed to change Zend framework code.
Please consult google how to define default using the configuration without changing the framework.

T.

PR #2177, opened 2015-01-28
relocating maxAttempts functions to correct object SUP-2485 #time 3h

shpiNi:


plugins/admin_console/system_partner/lib/api/KalturaSystemPartnerLimit.php
```diff
@@ -38,6 +38,10 @@ public static function fromPartner($type, Partner $partner)
 			case KalturaSystemPartnerLimitType::LIVE_STREAM_OUTPUTS:
 				$limit->max = $partner->getMaxLiveStreamOutputs();
 				break;
+
+			case KalturaSystemPartnerLimitType::USER_LOGIN_ATTEMPTS:
+				$limit->max = $partner->getMaxLoginAttempts();
```
shpiNi: @tan-tan-kanarek  'USER_LOGIN_ATTEMPTS' limit's input in admin console is only 'max' so it needs to stay a 'KalturaSystemPartnerLimit' object with a single 'max' field and it's handling should be here. please advise if you think otherwise.

PR #2189, opened 2015-02-01
Jupiter 10.3.0 add drop folder event notification plugin

hilak:


hilak:
@yuly-roberman  Please review

PR #2197, opened 2015-02-04
Copying cuepoints on RECORDING_ANCHOR VOD flavor only

AviBueno:
1. Ensuring all LIVE asset tags are copied to VOD,
2. Copying cuepoints from LIVE to VOD only when the mediaInfo was created for the VOD asset with the RECORDING_ANCHOR tag.

PLAT-2451 #time 2d

### DIFF
--- a/plugins/cue_points/base/lib/kCuePointManager.php
+++ b/plugins/cue_points/base/lib/kCuePointManager.php
@@ -57,8 +57,7 @@ class kCuePointManager implements kObjectDeletedEventConsumer, kObjectChangedEve
 			return null;
 		}
 
-		$flavorAssetId = $mediaInfo->getFlavorAssetId();
-		$flavorAsset = assetPeer::retrieveById( $flavorAssetId );
+		$flavorAsset = $mediaInfo->getasset();
 		if ( ! $flavorAsset || ! $flavorAsset->hasTag(assetParams::TAG_RECORDING_ANCHOR) )
 		{
 			return null;
--- a/plugins/cue_points/base/lib/kCuePointManager.php
+++ b/plugins/cue_points/base/lib/kCuePointManager.php
@@ -36,12 +36,41 @@ class kCuePointManager implements kObjectDeletedEventConsumer, kObjectChangedEve
 	*/
 	public function shouldConsumeCreatedEvent(BaseObject $object)
 	{
-		if($object instanceof mediaInfo)
+		if ( self::getVodEntryBasedOnMediaInfoFlavorAsset($object) )
 		{
 			return true;
 		}
 
 		return false;
+	}
+
+	/**
+	 * Return a VOD entry (sourceType = RECORDED_LIVE) based on the flavorAsset that is
+	 * associated with the given mediaInfo object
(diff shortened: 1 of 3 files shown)


plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -30,6 +32,19 @@ public function shouldConsumeDeletedEvent(BaseObject $object)
 	}
 	
 	/* (non-PHPdoc)
+	 * @see kObjectCreatedEventConsumer::shouldConsumeCreatedEvent()
+	*/
+	public function shouldConsumeCreatedEvent(BaseObject $object)
+	{
+		if($object instanceof mediaInfo)
```
tan-tan-kanarek: @AviBueno, I think you better verify here the anchor tag on the asset.

plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -448,10 +452,23 @@ public static function postProcessCuePoints( $liveEntry, $cuePointsIds = null )
...
 
 	/**
-	 * @param LiveEntry $liveEntry
+	 * @param mediaInfo $mediaInfo
 	 */
-	public static function copyCuePointsFromLiveToVodEntry( $vodEntry )
+	public static function copyCuePointsFromLiveToVodEntry( $mediaInfo )
 	{
+		$flavorAssetId = $mediaInfo->getFlavorAssetId();
+		$flavorAsset = assetPeer::retrieveById( $flavorAssetId );
```
tan-tan-kanarek: @AviBueno, you can use $flavorAsset = $mediaInfo->getasset();

plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -448,10 +452,23 @@ public static function postProcessCuePoints( $liveEntry, $cuePointsIds = null )
...
+		if ( ! $flavorAsset || ! $flavorAsset->hasTag(assetParams::TAG_RECORDING_ANCHOR) )
+		{
+			return;
+		}
+
+		$vodEntry = $flavorAsset->getentry();
+		if ( ! $vodEntry || $vodEntry->getSourceType() != EntrySourceType::RECORDED_LIVE )
```
tan-tan-kanarek: @AviBueno, no need to check the source type if you already checked the recording anchor tag.

plugins/cue_points/base/lib/kCuePointManager.php
AviBueno: @tan-tan-kanarek The check requires retrieving both the flavorAsset and it's entry from the DB.
Even though they'll be cached in memory, working this way will require duplicating the code in copyCuePointsFromLiveToVodEntry().
That is the reason the code is part copyCuePointsFromLiveToVodEntry() in the first place.

AviBueno: OK, since shouldConsume*() methods are there in order not to produce logs, I will shift the check to shouldConsumeCreatedEvent().

PR #2218, opened 2015-02-09
update kuser type column

assafb:


assafb:
@erankor  @etameran  @rkreich  @tan-tan-kanarek  - I updated the type column to be int and default value 0.

tan-tan-kanarek:
@assafb, please delete the merged branch.

PR #2225, opened 2015-02-10
return empty list if userIdIn or groupIdIn not exist

assafb:


assafb:
@rkreich  , @tan-tan-kanarek , @erankor  - return empty list if the users in userIdIn/groupIdIn not exist entirely.

PR #2228, opened 2015-02-11
Allowing a co-publisher to delete a category entry

AviBueno:
PLAT-2492 #time 2h


tan-tan-kanarek:
Avi,
The puser may change and therefore the pusers list on the entry may be old.
User the kuser id, it may never change.

T.

AviBueno:
@tan-tan-kanarek updated the pull request. Please review.

PR #2247, opened 2015-02-15
Fix null object ref error

hilak:


hilak:
@yossipapi  Please review

PR #2250, opened 2015-02-15
Jupiter 10.5.0 plat 2524

shpiNi:


plugins/metadata/lib/api/filters/advanced/MetadataSearchFilter.php
```diff
@@ -300,8 +300,21 @@ protected function createSphinxSearchPhrase($partnerScope = null, $text, $metada
 	}
 	
 	
-	public function getFreeTextConditions($partnerScope, $freeTexts , $metadataProfileFieldIds = array()) 
+	public function getFreeTextConditions($partnerScope, $freeTexts , $advancedSearch)
```
tan-tan-kanarek: Why do you need to pass $advancedSearch as argument, it's the same as using $this.

PR #2254, opened 2015-02-16
Jupiter 10.5.0 sup 3251

shpiNi:


alpha/config/syndication.core.xsd
```diff
@@ -398,6 +417,11 @@
 				<xs:documentation>Format of the thumbnail asset file</xs:documentation>
 			</xs:annotation>
 		</xs:attribute>
+		<xs:attribute name="extension" type="xs:string" use="required">
```
tan-tan-kanarek: I don't think the extension should be required.

alpha/config/syndication.core.xsd
```diff
@@ -413,6 +437,16 @@
 				<xs:documentation>Thumbnail width</xs:documentation>
 			</xs:annotation>
 		</xs:attribute>
+		<xs:attribute name="fileSize" type="xs:int" use="required">
```
tan-tan-kanarek: File size shouldn't be required, in case of remote asset.

PR #2260, opened 2015-02-17
Add support for audio-only live streams

hilak:


hilak:

PR #2273, opened 2015-02-23
Jupiter 10.5.0 sup 3864

shpiNi:


alpha/lib/model/asset.php
```diff
@@ -547,6 +547,22 @@ public function getDownloadUrl($useCdn = false, $forceProxy = false, $preview =
 
 		return $this->getDownloadUrlWithExpiry(86400, $useCdn, $forceProxy, $preview);
 	}
+
+	public function getServeFlavorUrl($entry)
```
tan-tan-kanarek: No need to pass the entry to the asset.
If you need the entry within the asset use $this->getentry().

alpha/lib/model/asset.php
```diff
@@ -547,6 +547,22 @@ public function getDownloadUrl($useCdn = false, $forceProxy = false, $preview =
 
 		return $this->getDownloadUrlWithExpiry(86400, $useCdn, $forceProxy, $preview);
 	}
+
+	public function getServeFlavorUrl($entry)
+	{
+		if ($entry->getType() != entryType::MEDIA_CLIP)
```
tan-tan-kanarek: Document entry may also have assets, and media entry assets might be thumbnails, captions and related files that shouldn't use serve-flavor.
Maybe this method should be placed in flavorAsset and not in asset.

PR #2279, opened 2015-03-02
modifying getUrl to return serveFlavor url

shpiNi:


alpha/apps/kaltura/lib/kMrssManager.php
```diff
@@ -365,11 +273,21 @@ protected static function appendThumbAssetMrss(thumbAsset $thumbAsset, SimpleXML
...
+		$servePlayManifest = false;
+		$playManifestClientTag = null;
+		$storageId = null;
+
 		if(!$mrss)
 			$mrss = new SimpleXMLElement('<item/>');
-		
+		else
+		{
+			$servePlayManifest = $mrss->getServePlayManifest();
```
erankor: this section seems wrong, all these properties need to come from $mrssParams, not $mrss

alpha/apps/kaltura/lib/myUrlUtils.class.php
```diff
@@ -0,0 +1,8 @@
+<?php
+class myUrlUtils
```
tan-tan-kanarek: We don't create my... files anymore, it's an old convention.
Also, you can place this method in asset utils and call it replaceSpecialChars.

alpha/lib/model/flavorAsset.php
```diff
@@ -267,6 +267,31 @@ public function getVideoBitrate()
 		return $this->getBitrate();
 	}
 	
+	public function getServeFlavorUrl($fileSize = null)
```
tan-tan-kanarek: Why do you use new method instead of overwriting getDownloadUrl?

PR #2298, opened 2015-03-04
Changing filesync status to ERROR if it is not alrady deleted

AviBueno:
PLAT-2542 #time 1d


plugins/multi_centers/lib/kMultiCentersFlowManager.php
```diff
@@ -100,8 +100,12 @@ protected function updatedFileSyncImportFailed(BatchJob $dbBatchJob, kFileSyncIm
 			return $dbBatchJob;
 		}
 		
-		$fileSync->setStatus(FileSync::FILE_SYNC_STATUS_ERROR);
-		$fileSync->save();
+		if ( $fileSync->getStatus() != FileSync::FILE_SYNC_STATUS_DELETED )
```
tan-tan-kanarek: Please also verify that it's not purged.

PR #2333, opened 2015-03-16
removing no_pc criteria when no entitlement for tagPeer

shpiNi:


plugins/tag_search/lib/model/TagPeer.php
```diff
@@ -31,10 +31,6 @@ public static function setDefaultCriteriaFilter()
 			$privacyContexts = kEntitlementUtils::getKsPrivacyContextArray();
 			$c->addAnd(self::PRIVACY_CONTEXT, $privacyContexts, Criteria::IN);
 		}
```
shpiNi: @hilak I see by commits here that you're familiar with this code segment , please take a look.

tan-tan-kanarek:
@hilak, please review.

PR #2343, opened 2015-03-19
[C# API] Memory-efficient implementation of PostMultiPartWithFiles

IrmatDen:
Hello,

Due to how MemoryStream capacity is handled, going over 1'073'741'823 bytes start stressing the GC like crazy. A potential OutOfMemoryException is also thrown by HttpWebRequest if the network stream is buffering the data (which is the default behaviour).
The proposed changes solve this problem by totally bypassing the "fit everything in memory" before streaming and disable request stream buffering.

Impl. note: I had to create an inner private struct in a new region. I'm not too sure if a separate file is preferred; let me know if you'd like me to change this to another approach.

### DIFF
--- a/generator/sources/csharp/KalturaClient/KalturaClientBase.cs
+++ b/generator/sources/csharp/KalturaClient/KalturaClientBase.cs
@@ -405,11 +405,46 @@ namespace Kaltura
             string boundary = "---------------------------" + DateTime.Now.Ticks.ToString("x");
             request.ContentType = "multipart/form-data; boundary=" + boundary;
 
-            // use a memory stream because we don't know the content length of the request when we have multiple files
-            MemoryStream memStream = new MemoryStream();
-            byte[] buffer;
-            int bytesRead = 0;
+            byte[] paramsBuffer = BuildMultiPartParamsBuffer(kparams, boundary);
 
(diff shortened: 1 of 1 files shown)


kaltura-hooks:
Hi @IrmatDen,

zoharbabin:
@IrmatDen Can you please provide a test for the change? In what context did you discover the issue?

Thanks!

IrmatDen:
@zoharbabin Sorry, not proper test case, except I had previously compiled & tested it on the autogenerated libs (also, mistakenly created a PR over there, see: https://github.com/kaltura/KalturaGeneratedAPIClientsCsharp/pull/1).

My claim about GC stress comes from [System.IO.MemoryStream.EnsureCapacity(...) implementation](http://referencesource.microsoft.com/#mscorlib/system/io/memorystream.cs,1416df83d2368912). Since int.MaxValue == 2,147,483,647 bytes, EnsureCapacity()'s exponential behavior falls apart at floor(int.MaxValue / 2), where the buffer is re-allocated/copied & gc'd on each write.

The context was just testing your C# API with some big video files, and the upload times were awful due to the double buffering combined to said capacity issue.

tan-tan-kanarek:
Will be merged as part of https://github.com/kaltura/server/pull/2907.

PR #2351, opened 2015-03-23
added composer.json to php5.3 client library sources

rkreich:
@jessp01, I'm adding the composer.json file to php5.3.

### DIFF
--- a/generator/sources/php53/composer.json
+++ b/generator/sources/php53/composer.json
@@ -1,0 +1,21 @@
+{
+    "name": "kaltura/api-client-library",
+    "description": "An auto generated Composer package for the Kaltura API client library",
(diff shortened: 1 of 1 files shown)


generator/sources/php53/composer.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "kaltura/api-client-library",
```
tan-tan-kanarek: I think the name should contain "5.3", to prevent other and future libraries confusion.

rkreich: We want to submit it to https://packagist.org/, the php5.3 library already follows the latest php/composer standards. Do you want to add that name for backward compatibility or for forward compatibility?

PR #2358, opened 2015-03-26
usefull scripts

shpiNi:


alpha/scripts/utils/encodeObjectField.php
```diff
@@ -0,0 +1,42 @@
+<?php
```
shpiNi: all scripts were code review and used.

tan-tan-kanarek:
@MosheMaorKaltura, please review and merge as needed.

PR #2378, opened 2015-03-31
added dynamic objects support in response profiles

rkreich:


tan-tan-kanarek:
Please delete merged branches.

PR #2386, opened 2015-04-05
Keep original image file extension

MosheMaorKaltura:
PLAT2489#Time 2h
@yossipapi  please review


tan-tan-kanarek:
What if the uploaded file contain invalid extension e.g. "image.invalid.extension" or no extension? I think in such cases we should keep using jpg.

MosheMaorKaltura:
Good question.
If no extension then filename.noext is dowloaded, in all other case the source extension will be downloaded.

???? ??-iPhone ???

??-5 ????? 2015, ???? 19:03, ??Tan-Tan ?<notifications@github.com<mailto:notifications@github.com>> ???/?:?


What if the uploaded file contain invalid extension e.g. "image.invalid.extension" or no extension? I think in such cases we should keep using jpg.

-

PR #2387, opened 2015-04-05
add optional fileName parameter to getUrl

erankor:


api_v3/services/FlavorAssetService.php
```diff
@@ -604,11 +604,12 @@ public function deleteAction($id)
 	 * @param string $id
 	 * @param int $storageId
 	 * @param bool $forceProxy
+	 * @param string $fileName
 	 * @return string
 	 * @throws KalturaErrors::FLAVOR_ASSET_ID_NOT_FOUND
 	 * @throws KalturaErrors::FLAVOR_ASSET_IS_NOT_READY
 	 */
-	public function getUrlAction($id, $storageId = null, $forceProxy = false)
+	public function getUrlAction($id, $storageId = null, $forceProxy = false, $fileName = "")
```
tan-tan-kanarek: I would add $options (KalturaFlavorAssetUrlOptions) instead of fileName.

PR #2397, opened 2015-04-06
Introduce Maven build process.

ratliff:
Added pom.xml file to support use of Maven for building Java source.  Moved various source and test files into directories expected by Maven.


kaltura-hooks:
Hi @ratliff,

zoharbabin:
@ratliff - thanks for the contrib! Since this is a rather braking change to a client library that has been in use by many for several years now, we will need to reapply these changes as a new type of client library rather than apply the changes to the existing Java client lib. As an example you can refer to the PHP-Zend vs. PHP5 vs. PHP 5.2 client libs. 
Can you please create a new folder for Java-Maven and apply these changes to that folder?

Thanks!

tan-tan-kanarek:
I fact, all we need to do is to change the JavaClientGenerator and PojoClientGenerator to generate java files according to maven directory structure standards.
If you could implement it in your branch we will be glad to merge it.

ratliff:
OK  I will have a look at those to Generator classes.

ratliff:
Can you tell me HOW I generate the Java source from the "server" repository?  After cloning that repository I try to (naively) run generator/generate.php, but I get a bunch of configuration errors.  And what command line arguments do I pass to have it generate ONLY the Java source?  I think I see how I can modify the two Generator classes to put things into the directories that Maven wants, but in order to test my modifications, I need to know how to run the generation process.

KalturaCommunity:
Hi Mark.

Please use:
# php /opt/kaltura/app/generator/generate.php java

The resulted libraries will be placed under:
/opt/kaltura/web/content/clientlibs/java_%d-%m-%Y.tar.gz

May the source be with you,

Jess Portnoy

jessp01:
Closing this one as a new one on the same topic was made here:
https://github.com/kaltura/server/issues/2971

PR #2405, opened 2015-04-12
The root cause for SUP-4103 is that there are no thumbnails available an...

noam-arad:
The root cause for SUP-4103 is that there are no thumbnails available andd we try to generate them on the fly, in order to generate them we try to get the flavor_asset with the highest bitrate to get the thumbnails from, in this case the customer has external storage configured and "delete after export" so that some of the flavor assets are connected to file syncs that are deleted. The fix is to see if to get the flavor asset with the highest bitrate and a file sync that is in status "ready". Such a flavor asset doesn't have to exist but this is still a better solution than the one we have in place.

@yuly-roberman could you please review.

### DIFF
--- a/alpha/apps/kaltura/lib/myEntryUtils.class.php
+++ b/alpha/apps/kaltura/lib/myEntryUtils.class.php
@@ -754,7 +754,7 @@ class myEntryUtils
 	    				if(is_null($flavorAsset) || !($flavorAsset->hasTag(flavorParams::TAG_MBR) || $flavorAsset->hasTag(flavorParams::TAG_WEB)))
 					    {
     						// try the best playable
-						    $flavorAsset = assetPeer::retrieveHighestBitrateByEntryId($entry->getId());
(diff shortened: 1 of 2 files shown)


tan-tan-kanarek:
Please change retrieveHighestBitrateByEntryId instead of creating a new method.
The existing method is used today only for finding the best source for thumbnailing, so it should be fixed for all thumbnailing flows.

noam-arad:
It is also called from "updateVideoDimensions". I'd rather not affect the code of flows I am not sure what is done in them.

tan-tan-kanarek:
updateVideoDimensions also requires a ready file-sync, that's exactly why I ask you to change the existing method, because until today it returned wrong data and I want you to fix it for all use cases.
As for the flows you're not sure about, that's why I'm here.

noam-arad:
@tan-tan-kanarek The change was done and pushed.

PR #2417, opened 2015-04-13
Partners that are part of local.ini:legacy_get_url_partners will get www...

MosheMaorKaltura:
....kaltura.com

PLAT-2811 #


MosheMaorKaltura:
Tested on PA report and as a hot fix on production server.
The code was written by Eran E. in origin.

PR #2420, opened 2015-04-14
in entry.php in line 1245 there was '=' instead of '==' that caused acti...

noam-arad:
...ons to run as partner '1'

Jupiter-10.8.0_SUP-4385


noam-arad:
This issue was fixed by @Galmoss. The HF was approved by @etameran .

PR #2424, opened 2015-04-14
Jupiter 10.8.0 plat 2042

hilak:


configurations/batch/batch.ini.template
```diff
@@ -544,6 +544,7 @@ type												= KAsyncDropFolderWatcher
 maximumExecutionTime								= 3000
 scriptPath											= ../plugins/drop_folder/batch/DropFolderWatcher/KAsyncDropFolderWatcherExe.php
 params.tags											= *
+params.mrss.xmlPath									= @WEB_DIR@/tmp/dropFolderFiles
```
tan-tan-kanarek: Why don't you use the common params.tempDirectoryPath that defined in [Worker] section?

plugins/bulk_upload/drop_folder_xml/lib/kDropFolderXmlEventsConsumer.php
```diff
@@ -190,20 +190,9 @@ private function setFileError(DropFolderFile $file, $status, $errorCode, $errorD
...
-			{
-				foreach ($parsedDropFolderFiles as $parsedFile) 
-				{
-					$parsedFile->setStatus(DropFolderFileStatus::PURGED);
-					$parsedFile->save();
-				}
-			}
-		}
 		
+		$xmlFileHandler = kDropFolderXmlFileHandler::getHandlerInstance($folder->getType());
```
tan-tan-kanarek: If the handling is so different, why won't use different event consumers instead?

hilak: @tan-tan-kanarek  I didn't think it was correct to add a new event consumer for just 1 new drop folder type. I can change this if required.

configurations/batch/batch.ini.template
hilak: Since the kMrssXmlHandler later needs to use this path to create a bulk upload job (moved this file from this location to create a filesync), the location has to be in the @WEB_DIR@, otherwise the code has no permission to use the file (as with the import job).

plugins/bulk_upload/drop_folder_xml/plugins/batch/Engine/KFeedDropFolderEngine.php
```diff
@@ -0,0 +1,269 @@
...
+		//get items
+		$feedItems = $feed->xpath ($this->dropFolder->feedItemInfo->itemXPath);
+		if ($this->dropFolder->itemHandlingLimit > 0 && count ($feedItems) > $this->dropFolder->itemHandlingLimit)
+		{
+			KalturaLog::err("Reached pulling limit for drop folder ID [" . $this->dropFolder->id . "].");
+			
+			array_splice ($feedItems, $this->dropFolder->itemHandlingLimit);
+			
+			$dropFolderUpdate = new KalturaFeedDropFolder();
+			$dropFolderUpdate->errorDescription = FeedDropFolderPlugin::ERROR_MESSAGE_INCOMPLETE_HANDLING . $this->dropFolder->id;
```
assafb: @hilak  - I'd add the limit size in the error message

PR #2430, opened 2015-04-16
Preview and embed broken when no playlists are available

jpluijmers:
The preview and embed was broken for some of my clients. It came down to the exit early statement in updatePlayers(). Here it is required to have at least one player and one playlist. After this patch you are no longer required to have both types. One player of either type will still work.


amirch1:
Good catch - thanks!
Since our current working branch is Jupiter-10.12.0, closing this pull request and opening https://github.com/kaltura/server/pull/2580

PR #2441, opened 2015-04-20
Fix flavorAsset->getUrl to support unified document entry serving.

yossipapi:


alpha/lib/model/asset.php
```diff
@@ -478,7 +478,7 @@ public function getFormat()
 			return null;
 	}
 	
-	public function getExternalUrl($storageId)
+	public function getExternalUrl($storageId, $fileName = null)
```
tan-tan-kanarek: Why won't you pass core representation of KalturaFlavorAssetUrlOptions?

yossipapi: @tan-tan-kanarek  I can pass the entire. Currently it has only one attribute so three wasn't much point to it.
I will update the code as to your request.

PR #2451, opened 2015-04-21
Support Boolean Time Based Advanced Search Condition

rkreich:


rkreich:
This is Jira PLAT-2817

PR #2467, opened 2015-04-26
Adding a permission per partner to be able to configure "copy entitelmen...

noam-arad:
...t" for kaltura live streams and if the flag is on I make sure the categories (and entitlements) are copied (synced).

PLAT-2744


alpha/lib/data/live/kLiveEntryRecordingOptions.php
```diff
@@ -23,10 +23,15 @@ public function setShouldCopyEntitlement($shouldCopyEntitlement)
 	}
 	
 	/**
+	 * @var int
 	 * @return boolean
 	 */
-	public function getShouldCopyEntitlement()
+	public function getShouldCopyEntitlement($partnerId)
 	{
+		if ( PermissionPeer::isValidForPartner(PermissionName::FEATURE_LIVE_STREAM_COPY_ENTITELMENTS, $partnerId) )
```
tan-tan-kanarek: This way you prevent disabling it using the API.
I suggest that you will ovveride KalturaEntryService::prepareEntryForInsert in KalturaLiveEntryService and check if this attribute is null, and only in such case, use the partner defaults.

noam-arad: @tan-tan-kanarek Please review changes to fit your comments

api_v3/lib/types/entry/KalturaLiveEntry.php
```diff
@@ -126,7 +126,21 @@ public function toInsertableObject($sourceObject = null, $propsToSkip = array())
 	{
 		if(is_null($this->recordStatus))
 			$this->recordStatus = (PermissionPeer::isValidForPartner(PermissionName::FEATURE_LIVE_STREAM_RECORD, kCurrentContext::getCurrentPartnerId()) ? KalturaRecordStatus::APPENDED : KalturaRecordStatus::DISABLED);
-			
+
+		if (is_null($this->recordingOptions))
```
tan-tan-kanarek: There is no reason to instantiate recordingOptions if FEATURE_LIVE_STREAM_COPY_ENTITELMENTS permission is off.

api_v3/lib/types/entry/KalturaLiveEntry.php
```diff
@@ -126,7 +126,21 @@ public function toInsertableObject($sourceObject = null, $propsToSkip = array())
...
+		{
+			$this->recordingOptions = new kLiveEntryRecordingOptions();
```
tan-tan-kanarek: You should instantiate KalturaLiveEntryRecordingOptions and not kLiveEntryRecordingOptions.

PR #2475, opened 2015-04-27
Fix permission of KalutraLiveEntry

noam-arad:


tan-tan-kanarek:
@noam-arad, please delete merged branches.

PR #2478, opened 2015-04-27
Remove tag length restriction

hilak:


tan-tan-kanarek:
It won't help removing it, the real limitation is in ktagword::MAXIMUM_TAG_LENGTH.

hilak:
@tan-tan-kanarek Sanity passed though, strangely- tag string of more than 100 chars was processed and ingested successfully. I will check again.

PR #2494, opened 2015-04-29
batch job file name should be taken from the actual physical file

hilak:


tan-tan-kanarek:
@hilak, please delete merged branches.

PR #2518, opened 2015-05-05
Jupiter 10.11.0 too many logs

noam-arad:
There are too many logs from the batch scheduler written to a file which may cause the system to be non responsive either because the scheduler is dealing with too much I/O or that the file is too big and there is no space left on the hard drive. The solution should be to change the logs to be written only once a minute.


batch/scheduler/KGenericScheduler.class.php
```diff
@@ -325,7 +333,11 @@ private function shouldExecute(KSchedularTaskConfig $taskConfig)
 				return false;
 		}
 
-		KalturaLog::debug("Worker [{$taskConfig->name}] id [{$taskConfig->id}] running batches [$runningBatches] max instances [{$taskConfig->maxInstances}]");
+		if ( !isset($this->lastWorkerLog[$taskConfig->id]) || (($this->logWorkerInterval >= 0) && (time() >= ($this->lastWorkerLog[$taskConfig->id] + $this->logWorkerInterval))) )
```
tan-tan-kanarek: If $this->logWorkerInterval == 0 it will log every time, make sure it's larger than 0.

tan-tan-kanarek:
Generally, try to avoid so many whitespace changes, it's annoying and makes the review much harder for no reason.

PR #2523, opened 2015-05-05
PLAT-2387

hilak:
1. Fix problems caused by incorrect translation of enum
2. Fix the "comment added" template so that the $scope->getEvent is not called in the
content parameter evaluations


hilak:
@assafb  Please review

PR #2525, opened 2015-05-06
SUP-4358 additional php fixes

EladCohenKaltura:
SUP-4358
add valid return value so that the ValidateObjectType will pass

Ran all tests of all clients through the testing scripts plus ran @jessp01 script as well 
Still waiting for the objc testing 
FYI @MosheMaorKaltura


tan-tan-kanarek:
Why do you return string in the Zend client and double in the php 4 and 5 clients, what's the difference?
What about php 5.3 client?

PR #2527, opened 2015-05-06
Jupiter 10.10.0 hide partner uiconfs

hilak:


hilak:
@assafb  Please review- requirement for Clarizen ticket
https://app2.clarizen.com/Clarizen/Project/299216357?orgId=130510

tan-tan-kanarek:
As I understand this action called rarely by account administrators only, so please use partner level permission instead of partner object custom data.

PR #2528, opened 2015-05-06
Added KalturaEntryCaptionAssetSearchItem

rkreich:
Added KalturaEntryCaptionAssetSearchItem to support caption search in entry list services


plugins/content/caption/search/lib/model/filters/EntryCaptionAssetSearchFilter.php
```diff
@@ -122,7 +122,49 @@ public function applyCondition(IKalturaDbQuery $query)
 			$this->addCondition($this->getContentMultiLikeOr(), $query);
 		}
 	}
-	
+
+	protected function createSphinxSearchPhrase($text)
```
tan-tan-kanarek: Please rename to createSphinxMatchPhrase to match the name in EntryCuePointSearchFilter.
And please use this method also in EntryCaptionAssetSearchFilter::addCondition, no reason to have the same logic twice.

PR #2543, opened 2015-05-10
Jupiter 10.11.0 plat 2857

noam-arad:
Creation of "UserEntry" abstract object and UserEntryService and all that is needed to support the service on the most basic needs.


noam-arad:
I'll close the pull request and wait until we update the main branch so I can pull again.

PR #2547, opened 2015-05-11
Save should be done in diffMetadataProfile function and not outside s…

noam-arad:
…ince there could be a race condition where the version isn't updated whent the transform metadata batch service runs.

### DIFF
--- a/plugins/metadata/services/MetadataProfileService.php
+++ b/plugins/metadata/services/MetadataProfileService.php
@@ -191,14 +191,13 @@ class MetadataProfileService extends KalturaBaseService
 		{		    
 			try
 			{
+				//Save is done in this function, no need to save again.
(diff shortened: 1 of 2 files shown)


plugins/metadata/services/MetadataProfileService.php
```diff
@@ -439,8 +439,6 @@ function updateDefinitionFromFileAction($id, $xsdFile)
 			throw new KalturaAPIException(MetadataErrors::METADATA_UNABLE_TO_TRANSFORM, $e->getMessage());
 		}
 		
-		$dbMetadataProfile->save();
```
tan-tan-kanarek: Keep the save, you should save in the same method that you call the setters.
Calling save without any change won't trigger a db transaction so I prefer to many saves than not enough.

PR #2548, opened 2015-05-11
SUP-3451 Swedish language support required

EladCohenKaltura:
SUP-3451
Accept-Language header sent from page is set by Zend framework as not installed
Adding files to support Swedish in specific to redirect to english (as hebrew)

@MosheMaorKaltura


tan-tan-kanarek:
No one need the VAR console, so unless the ticket asked for it specifically, then it's probably not needed.

PR #2550, opened 2015-05-12
#PLAT-2353 #comment index to sphinx deleted cue points

MichalRadwantzor:


plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -219,7 +219,11 @@ protected function deleteCuePoints(Criteria $c)
 		CuePointPeer::setUseCriteriaFilter(true);
 
 		foreach($cuePoints as $cuePoint)
+		{
+			$cuePoint->reload();
```
MichalRadwantzor: @tan-tan-kanarek do you think it's better to call $cuePoint->setStatus(CuePointStatus::DELETED) here instead?

tan-tan-kanarek: What for? it's already deleted in the doUpdate.
However, if you want to reload them all, do it using retrieveByPKs in order to do it in one command in the DB.

PR #2585, opened 2015-05-22
PLAT-2251

hilak:
New email notification dispatched when drop folder error description changes


hilak:
@assafb Please review

PR #2587, opened 2015-05-25
SUP4292 Validate int boundries

EladCohenKaltura:
SUP4292
Expedn the part int to include values that have E notation


MosheMaorKaltura:
Code change approved.

PR #2592, opened 2015-05-25
Change so that when someone wants to filter according to nameOrRefere…

noam-arad:
…nceIdStartsWith then we will definitely return also the exact match and also the completions for that expression.

I made the change so that only this request with this filter will be changed. _likex_name_or_reference_id is used only when the filter has nameOrReferenceIdStartsWith field this way I will affect only sphinx queries that have that

### DIFF
--- a/plugins/sphinx_search/lib/SphinxCategoryCriteria.php
+++ b/plugins/sphinx_search/lib/SphinxCategoryCriteria.php
@@ -143,8 +143,8 @@ class SphinxCategoryCriteria extends SphinxCriteria
 		{
 			$names = $filter->get('_likex_name_or_reference_id');
 			KalturaLog::debug("Attach free text [$names]");
-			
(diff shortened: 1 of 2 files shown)


plugins/sphinx_search/lib/SphinxCategoryCriteria.php
```diff
@@ -143,8 +143,8 @@ protected function applyFilterFields(baseObjectFilter $filter)
 		{
 			$names = $filter->get('_likex_name_or_reference_id');
 			KalturaLog::debug("Attach free text [$names]");
-			
-			$this->addFreeTextToMatchClauseByMatchFields($names, categoryFilter::NAME_REFERNCE_ID, null, true);
+			$additional = array("(@(name,reference_id) $names)");
```
tan-tan-kanarek: Please move the implementation into addFreeTextToMatchClauseByMatchFields and check also for KalturaUserFilter::idOrScreenNameStartsWith and firstNameOrLastNameStartsWith.

PR #2622, opened 2015-06-01
fix thumbAsset->list filter backward compatibility

Galmoss:
SUP-4924

### DIFF
--- a/api_v3/lib/types/KalturaObject.php
+++ b/api_v3/lib/types/KalturaObject.php
@@ -869,7 +869,7 @@ abstract class KalturaObject
 	}
 
 	public function cast($className) {
-		if(!is_subclass_of($className, get_class($this)))
+		if(!is_subclass_of($className, get_class($this)) && !is_subclass_of(get_class($this),$className))
 			throw new KalturaAPIException(KalturaErrors::INVALID_OBJECT_TYPE, get_class($this));
 			
 	    return unserialize(sprintf(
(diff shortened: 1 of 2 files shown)


Galmoss:
Added support for filter that is not KalturaThumbAssetFilterand not KalturaAssetFilter:

Cast it upward to KalturaAssetFilter and then cast it downward to KalturaThumbAssetFilterand

api_v3/services/ThumbAssetService.php
```diff
@@ -793,6 +793,9 @@ function listAction(KalturaAssetFilter $filter = null, KalturaFilterPager $pager
 		}
 		elseif(! $filter instanceof KalturaThumbAssetFilter)
 		{
+		    	if(!is_subclass_of('KalturaThumbAssetFilter', get_class($filter)))
+		        	$filter = $filter->cast('KalturaAssetFilter');
+		    
 			$filter = $filter->cast('KalturaThumbAssetFilter');
```
yossipapi: @Galmoss don't you need an else here otherwise you will never create a filter of type "KalturaAssetFilter"

api_v3/lib/types/KalturaObject.php
```diff
@@ -869,7 +869,7 @@ public function trimStringProperties(array $propertyNames)
 	}
 
 	public function cast($className) {
-		if(!is_subclass_of($className, get_class($this)))
+		if(!is_subclass_of($className, get_class($this)) && !is_subclass_of(get_class($this),$className))
```
tan-tan-kanarek: is_subclass_of accepts object as first argument, you don't need get_class.

tan-tan-kanarek:
Please pull against 10.15.0.

PR #2624, opened 2015-06-01
image ext preserved if legit

shpiNi:


MosheMaorKaltura:
@shpiNi  why do we need to verify that file ext. is one of the following - jpg,jpeg,bmp,png,gif,tif,tiff, we can just save the file with same extension that it was uploaded with. this is how the logic works now in case of upload \ download image.

shpiNi:
@MosheMaorKaltura these are valid image extensions we work with. other extensions , such as video,documents or executable ones for example, shouldn't be used in an image entry.

PR #2625, opened 2015-06-01
Whitelist filename chars for flavorAsset-getUrl

Galmoss:
SUP-4491
Up until now we used to go over the file name and strip chars from a blacklist, but there are too many chars that can brake the url.
From now on we keep only chars that are in the whitelist and that we know they work.

### DIFF
--- a/infra/general/kString.class.php
+++ b/infra/general/kString.class.php
@@ -440,4 +440,9 @@ class kString
 		return str_replace(array('?', '|', '*', '\\', '/' , '>' , '<', '&', '[', ']',' '), '_', $url);
 	}
 
+	public static function keepOnlyValidUrlChars($url)
+	{
+	    return preg_replace('/[^A-Za-z0-9\-\.\_\~\!\$\&\(\)\*\+\,\;\=\:\@]/', '_', $url);
+	}
+	
 }
--- a/infra/general/kString.class.php
+++ b/infra/general/kString.class.php
@@ -442,7 +442,7 @@ class kString
 
 	public static function keepOnlyValidUrlChars($url)
 	{
-	    return preg_replace('/[^A-Za-z0-9\-\.\_\~\!\$\&\(\)\*\+\,\;\=\:\@]/', '_', $url);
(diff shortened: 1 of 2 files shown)


infra/general/kString.class.php
```diff
@@ -440,4 +440,9 @@ public static function stripInvalidUrlChars($url)
 		return str_replace(array('?', '|', '*', '\\', '/' , '>' , '<', '&', '[', ']',' '), '_', $url);
 	}
 
+	public static function keepOnlyValidUrlChars($url)
+	{
+	    return preg_replace('/[^A-Za-z0-9\-\.\_\~\!\$\&\(\)\*\+\,\;\=\:\@]/', '_', $url);
```
tan-tan-kanarek: Please keep only the needed slash, before the dash -, all the rest are not needed.

Galmoss:
@tan-tan-kanarek I made the change, is it ok now?

infra/general/kString.class.php
```diff
@@ -440,4 +440,9 @@ public static function stripInvalidUrlChars($url)
 		return str_replace(array('?', '|', '*', '\\', '/' , '>' , '<', '&', '[', ']',' '), '_', $url);
 	}
 
+	public static function keepOnlyValidUrlChars($url)
+	{
+	    return preg_replace('/[^A-Za-z0-9\-._~!$&()*+,;=:@]/', '_', $url);
```
tan-tan-kanarek: What's the point in replacing underscore with underscore?

Galmoss: @tan-tan-kanarek  it's replacing what is not underscore or any other char in the list

MosheMaorKaltura: Gal,
Please state what tests have been done to validate this change?

Galmoss: @MosheMaorKaltura I tested the change of using keepOnlyValidUrlChars on my own environment and verified it replaced bad characters with underscore and that the download was successful.
About tantan's note, I removed the unnecessary slashes and tested that preg_replace still replacing characters not in that list.

MosheMaorKaltura:
Approved, please merge.

PR #2615, opened 2015-05-29
Jupiter 10.12.0 ps 2259

hilak:

PR #2626, opened 2015-06-01
Jupiter 10.12.0 ps 2259

hilak:


hilak:
Merging after code review and sanity. See:
https://github.com/kaltura/server/pull/2615

PR #2629, opened 2015-06-02
If the search parameter is less than 3 characters but contains spaces we will replace the spaces with " and search again.

noam-arad:
Reviewed by @MosheMaorKaltura .
This keeps the regular functionality that we already had but allows to search for complex expressions that have spaces in them.
Solution brought forth by @Galmoss .


noam-arad:
We decided to go in a different direction, instead of using " we will index all spaces with _ as well and when searching we will search _ as well (see branch Jupiter-10.13.0-SUP-4362_)

PR #2632, opened 2015-06-02
Jupiter 10.12.0 plat 3014

shpiNi:


api_v3/lib/KalturaEntryService.php
```diff
@@ -964,6 +964,15 @@ protected function add(KalturaBaseEntry $entry, $conversionProfileId = null)
 			$templateEntry = entryPeer::retrieveByPK($conversionProfile->getDefaultEntryId());
 			if($templateEntry)
 			{
+				$entryType = $entry->type ? $entry->type : "null";
+				$compareMsgString = "ENTRY_TEMPLATE_COPY - original entry:template entry. type - " . $entryType.':'.$templateEntry->getType();
```
tan-tan-kanarek: $entryType is API value and $templateEntry->getType() is core value, they won't be equal when plugins are used even if for the same type.

api_v3/lib/KalturaEntryService.php
```diff
@@ -964,6 +964,15 @@ protected function add(KalturaBaseEntry $entry, $conversionProfileId = null)
 			$templateEntry = entryPeer::retrieveByPK($conversionProfile->getDefaultEntryId());
 			if($templateEntry)
 			{
+				$entryType = $entry->type ? $entry->type : "null";
+				$compareMsgString = "ENTRY_TEMPLATE_COPY - original entry:template entry. type - " . $entryType.':'.$templateEntry->getType();
+				if (property_exists(get_class($entry),'mediaType'))
```
tan-tan-kanarek: Instead, check if $entry is_a KalturaMediaEntry.

PR #2633, opened 2015-06-02
SUP 4001 - sphinx category filter double partnerID

EladCohenKaltura:
SUP-4001
PartnerID was concatenated twice to the sphix search condition for the case of category _eq_display_in_search


EladCohenKaltura:
ran compat check for this fix on pa-reports

PR #2634, opened 2015-06-02
multi-language caption ingestion PHASE 1

shpiNi:


plugins/content/caption/base/lib/api/KalturaCaptionAsset.php
```diff
@@ -61,6 +61,24 @@ class KalturaCaptionAsset extends KalturaAsset
...
 	public $status;
 	
+	/**
+	 * The parent id of the asset
+	 *
+	 *@var string
+	 *@readonly
+	 *
+	 */
+	public $parentId;
```
assafb: @shpiNi  - I think this should not be readonly because the batch should populate it when creating the captions.

plugins/content/caption/base/services/CaptionAssetService.php
```diff
@@ -209,7 +215,29 @@ protected function attachFile(CaptionAsset $captionAsset, $fullPath, $copyOnly =
 		$captionAsset->setStatus(CaptionAsset::ASSET_STATUS_READY);
 		$captionAsset->save();
 	}
-    
+
+	/**
+	 * @param CaptionAsset $captionAsset
+	 * @param string $path
+	 */
+	protected function detectMultiLanguageFile(CaptionAsset $captionAsset, $path = null)
```
assafb: @shpiNi  - I'd change the name to something like isMultiLanguageFile. It implies on the expected returned value more clearly.

plugins/content/caption/base/lib/api/KalturaCaptionAsset.php
```diff
@@ -61,6 +61,24 @@ class KalturaCaptionAsset extends KalturaAsset
...
+	
+	/**
+	 * Is multi-language caption asset
+	 *
+	 *@var bool
+	 *@readonly
+	 *
+	 */
+	public $isMulti;
```
tan-tan-kanarek: 1. Why is it readonly?
2. Why won't you enable a new language enum value that called multilingual, or leave it null to mark it as multilingual.

plugins/content/caption/base/services/CaptionAssetService.php
```diff
@@ -201,7 +201,13 @@ protected function attachFile(CaptionAsset $captionAsset, $fullPath, $copyOnly =
 
 		$finalPath = kFileSyncUtils::getLocalFilePathForKey($syncKey);
 		list($width, $height, $type, $attr) = getimagesize($finalPath);
-		
+
+		if ($this->detectMultiLanguageFile($captionAsset , $finalPath))
```
tan-tan-kanarek: No reason to parse the content to detect it, just set it to multilingual when you ingest it.

plugins/content/caption/base/lib/model/CaptionAsset.php
```diff
@@ -23,11 +25,15 @@ public function applyDefaultValues()
...
 	public function getLabel()			{return $this->getFromCustomData(self::CUSTOM_DATA_FIELD_LABEL);}
+	public function getParentId()       {return $this->getFromCustomData(self::CUSTOM_DATA_PARENT_ID);}
+	public function getIsMulti()        {return $this->getFromCustomData(self::CUSTOM_DATA_IS_MULTI);}
 
 	public function setLanguage($v)		{$this->putInCustomData(self::CUSTOM_DATA_FIELD_LANGUAGE, $v);}
 	public function setDefault($v)		{$this->putInCustomData(self::CUSTOM_DATA_FIELD_DEFAULT, (bool)$v);}
 	public function setLabel($v)		{$this->putInCustomData(self::CUSTOM_DATA_FIELD_LABEL, $v);}
-	
+	public function setParentId($v)     {$this->putInCustomData(self::CUSTOM_DATA_PARENT_ID, $v);}
+	public function setIsMulti($v)      {$this->putInCustomData(self::IS_MULTI, (bool)$v);}
```
erankor: the const name is wrong here

plugins/content/caption/base/lib/api/KalturaCaptionAsset.php
yossipapi: @tan-tan-kanarek regarding you last comment for updating the parent id field.
Would we like to enable suers to update parent id via the API? If not maybe we should add requires permission on this field and add the permission only to the batch partner

tan-tan-kanarek: We can block it using @requiresPermission annotation.

shpiNi: @tan-tan-kanarek @yossipapi parent_id needs to be inserted at first only , hence it's insertonly.

plugins/content/caption/search/lib/kCaptionSearchFlowManager.php
```diff
@@ -11,8 +11,8 @@ class kCaptionSearchFlowManager implements kObjectDataChangedEventConsumer, kObj
 	public function shouldConsumeAddedEvent(BaseObject $object)
 	{
 		if(class_exists('CaptionAsset') && $object instanceof CaptionAsset 
-				&& CaptionSearchPlugin::isAllowedPartner($object->getPartnerId()
-					&& $object->getStatus() == CaptionAsset::ASSET_STATUS_READY)){
+				&& CaptionSearchPlugin::isAllowedPartner($object->getPartnerId())
+					&& $object->getStatus() == CaptionAsset::ASSET_STATUS_READY && $object->getLanguage() != CaptionPlugin::MULTI_LANGUAGE){
```
shpiNi: argument in isAllowedPartner function is currently invalid , needs to be considered in regression tests.

PR #2636, opened 2015-06-03
In case ExternalMediaEntry make sure that externalSourceType is copied.

MosheMaorKaltura:
PLAT-2983#time 4h


tan-tan-kanarek:
Approved.

PR #2639, opened 2015-06-03
Replace space with '=' before inserting/updating tags in the sphinx i…

noam-arad:
…n order to be able to search for tags that have space in them and the first word has less than 3 letters. When searching search for the expression with spaces replaced by '='


MosheMaorKaltura:
Rename gettagWithEqual to be getTagWithEqual based on coding convention.
No defects were found.

PR #2642, opened 2015-06-04
adding multi-language const

shpiNi:


api_v3/lib/types/enums/KalturaLanguage.php
```diff
@@ -148,4 +148,5 @@ class KalturaLanguage extends KalturaStringEnum
 	const JI = 'Yiddish';
 	const YO = 'Yoruba';
 	const ZU = 'Zulu';
+	const MU = 'Multi-language';
```
tan-tan-kanarek: I think that "Multilingual" will be better phrase.

PR #2682, opened 2015-06-15
Add current time clock and lasy sync point time display on big red button page.

noam-arad:
Reviewed by @MosheMaorKaltura


noam-arad:
Reviewed by @MosheMaorKaltura

PR #2686, opened 2015-06-16
Jupiter 10.13.0 cenc

noam-arad:


plugins/drm/lib/DrmLicenseUtils.php
```diff
@@ -64,10 +64,22 @@ public static function signDataWithKey($dataToSign, $signingKey)
 
 	public static function createCustomData($entryId, KalturaFlavorAssetArray $flavorAssets, $signingKey)
 	{
-		$flavorParamIds = "";
+		$flavorIds = "";
+        $first = true;
 		foreach ($flavorAssets as $flavor)
```
tan-tan-kanarek: @noam-arad, please use implode instead of this foreach.

PR #2687, opened 2015-06-16
impersonate when distributing cross-platform

shpiNi:


plugins/content_distribution/batch/Distribute/engines/DistributionEngine.php
```diff
@@ -77,11 +77,13 @@ protected function getEntry($partnerId, $entryId)
 	 * @param string $flavorAssetIds comma seperated
 	 * @return array<KalturaFlavorAsset>
 	 */
-	protected function getFlavorAssets($partnerId, $flavorAssetIds)
+	protected function getFlavorAssets($partnerId, $flavorAssetIds, $entryId = null)
```
tan-tan-kanarek: The entry id is not needed here.

shpiNi: @tan-tan-kanarek  i agree but it's for the engine's backward compatibility.

plugins/content_distribution/providers/cross_kaltura/lib/batch/CrossKalturaDistributionEngine.php
```diff
@@ -228,27 +228,43 @@ protected function getEntryObjects(KalturaClient $client, $entryId, KalturaDistr
...
-            }
-            catch (Exception $e) {
-                KalturaLog::err('Cannot get list of flavor assets - '.$e->getMessage());
-                throw $e;
-            }
+			try {
+				KalturaLog::debug('Getting entry\'s flavor assets');
+				if($this->distributionProfile->partnerId)
+					$flavorAssetsList = $this->getFlavorAssets($this->distributionProfile->partnerId, $data->entryDistribution->flavorAssetIds, $entryId);
+				else
```
tan-tan-kanarek: In what use case would the you get into the else?

plugins/content_distribution/batch/Distribute/engines/DistributionEngine.php
tan-tan-kanarek: Compatibility in API usage as client is meaningless.

plugins/content_distribution/providers/cross_kaltura/lib/batch/CrossKalturaDistributionEngine.php
shpiNi: @tan-tan-kanarek, using testMe ,dist-profiles are created without a partner id (also being fixed in this request).
there are currently some prod ones.

tan-tan-kanarek: Such distribution profiles (that are not associated with partner id) can't be used to distribute entries because they won't appear in the partner's KMC.
Let's take this discussion offline.

PR #2689, opened 2015-06-16
Jupiter 10.12.0 plat 2908

shpiNi:


MosheMaorKaltura:
Approved.

alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
```diff
@@ -447,7 +437,11 @@ protected function initSmilManifest($flavorAssets)
 
 	private function removeNotAllowedFlavors($flavorAssets)
 	{
-		$returnedFlavors = array();		
+		if(!$this->secureEntryHelper)
+			return $flavorAssets;
```
yuly-roberman: @shpiNi why do we allow a case where secureEntryHelper is not set?

shpiNi: @yuly-roberman we bypass it when a 'kt' token is passed.

yuly-roberman: @shpiNi so make sure to validate it also in initFlavorIds() function

PR #2696, opened 2015-06-18
Make sure order enums are not empty

noam-arad:


plugins/cue_points/quiz/lib/api/KalturaQuestionCuePoint.php
```diff
@@ -9,6 +9,7 @@ class KalturaQuestionCuePoint extends KalturaCuePoint
 	/**
 	 * Array of key value answerKey->optionAnswer objects
 	 * @var KalturaOptionalAnswersArray
+	 * @filter order
```
tan-tan-kanarek: It doesn't make sense to order by array.

PR #2701, opened 2015-06-18
Fix update of QuizService so that values that are not set will not be…

noam-arad:
PLAT-3291 #time 1 hours


plugins/cue_points/quiz/services/QuizService.php
```diff
@@ -58,7 +58,7 @@ public function updateAction( $entryId, KalturaQuiz $quiz )
 	{
 		$dbEntry = entryPeer::retrieveByPK($entryId);
 		$kQuiz = QuizPlugin::validateAndGetQuiz( $dbEntry );
```
tan-tan-kanarek: Instead of fetching the kQuiz here and passing it into validateAndUpdateQuizData, just fetch it in validateAndUpdateQuizData, it's not used here anyway, you also won't need to pass the version.

PR #2715, opened 2015-06-21
Add protocol at the beginning of cdn white list

noam-arad:


MosheMaorKaltura:
Approve.

jessp01:
What about port?

noam-arad:
We know that the API listens on port 80 for HTTP and 443 for HTTPS it is enough to return the protocol

jessp01:
This is only true to SaaS. We should always code to consider alt ports. 
Especially since we also support CE and OnPrem from the same code base.

May the source be with you,

Jess Portnoy

Join us for the Kaltura Connect Virtual Summits 2015 - Register now!
virtualconnect.kaltura.com

noam-arad:
So your suggestion is to add the port received in $_SERVER['SERVER_PORT']?

jessp01:
Yes.

PR #2726, opened 2015-06-23
Jupiter 10.14.0 plat 3088 1

MosheMaorKaltura:
The following unit tests were executed - 
Call kFile::moveFile with the following input - 
Success case - 
1. Normal case of move file
2. Source file is read only (Copy only - no source delete)
Fail case - 
1. Dest dir is read only
2. Source file is read only + request to delete
 In all of the case no PHP warning should be raised.


tan-tan-kanarek:
Looks good.

PR #2733, opened 2015-06-23
Jupiter 10.14.0 ps 2276+7

assafb:


plugins/content/caption/base/batch/ParseMultiLanguageCaptionAsset/KAsyncParseMultiLanguageCaptionAsset.class.php
```diff
@@ -110,9 +128,11 @@ protected function parseMultiLanguage(KalturaBatchJob $job, KalturaParseMultiLan
 			$contentResource->content = $completeXML;
 
 			if (!$onlyUpdate)
-				$this->addCaption($entryId,$captionAsset, $contentResource);
+				$currentCaptionCreated = $this->addCaption($entryId,$captionAsset, $contentResource);
 			else
-				$this->setCaptionContent($id, $contentResource);				
+				$currentCaptionCreated = $this->setCaptionContent($id, $contentResource);				
+
+			$captionsCreated = $captionsCreated | $currentCaptionCreated;
```
assafb: @shpiNi  - use  '||' and not '|'

tan-tan-kanarek:
Generally, using text message as a flag for failing the asset is wrong.
Either fail the job and as result of its failure fail also the asset, or, check on the entry how many child-assets exist and fail the parent asset in none found.

PR #2758, opened 2015-06-29
Jupiter 10.14.0 user id to puser

noam-arad:


alpha/lib/model/UserEntry.php
```diff
@@ -30,4 +30,13 @@ public function applyDefaultValues()
 		$this->setStatus(UserEntryStatus::ACTIVE);
 	}
 
+	public function setKuserId($userId)
+	{
+		$kuser= kuserPeer::getKuserByPartnerAndUid(kCurrentContext::$ks_partner_id, $userId);
+		if (!$kuser)
+		{
+			throw new KalturaAPIException(KalturaErrors::INVALID_USER_ID, $userId);
```
MosheMaorKaltura: Change to KalturaException

tan-tan-kanarek: There is no KalturaException, kCoreException could be used here, but this setter performing translation that should be done in the API level, I would translate the puser to kuser in KalturaUserEntry::toInsertableObject.

PR #2759, opened 2015-06-29
Jupiter 10.14.0 plat 3137 and PLAT-2793 and PLAT-2794

noam-arad:

PR #2760, opened 2015-06-29
Jupiter 10.14.0 plat 2794

noam-arad:


noam-arad:
Merged in https://github.com/kaltura/server/pull/2759

PR #2761, opened 2015-06-29
add port to the cdn host that comes from the white list, in case some…

noam-arad:
…one configured Kaltura server on non standard port.

PLAT-2971 #time 10 minutes


noam-arad:
Approved on 10.14 and wasn't merged in time so I am merging it into 10.15

PR #2780, opened 2015-07-01
Jupiter 10.14.0 ps 2276+7

shpiNi:


plugins/content/caption/base/batch/ParseMultiLanguageCaptionAsset/KAsyncParseMultiLanguageCaptionAsset.class.php
```diff
@@ -120,7 +139,22 @@ protected function parseMultiLanguage(KalturaBatchJob $job, KalturaParseMultiLan
...
+		{
+			$errType = null;
+			$message = CaptionPlugin::FINISHED_PARSING_MESSAGE;
+			$status = KalturaBatchJobStatus::FINISHED;
+			$data = null;
+		}
+		else
+		{
+			$errType = KalturaBatchJobErrorTypes::KALTURA_API;
+			$message = CaptionPlugin::NO_CAPTIONS_MESSAGE;
```
tan-tan-kanarek: No need to use constant here.

plugins/content/caption/base/batch/ParseMultiLanguageCaptionAsset/KAsyncParseMultiLanguageCaptionAsset.class.php
```diff
@@ -120,7 +139,22 @@ protected function parseMultiLanguage(KalturaBatchJob $job, KalturaParseMultiLan
...
+			$status = KalturaBatchJobStatus::FAILED;
+		}
+
+		$this->closeJob($job, $errType, $errNumber, $message, $status, $data);
```
tan-tan-kanarek: No need to pass any $data here.

plugins/content/caption/base/batch/ParseMultiLanguageCaptionAsset/KAsyncParseMultiLanguageCaptionAsset.class.php
```diff
@@ -120,7 +139,22 @@ protected function parseMultiLanguage(KalturaBatchJob $job, KalturaParseMultiLan
...
+		if ($captionsCreated)
+		{
+			$errType = null;
+			$message = CaptionPlugin::FINISHED_PARSING_MESSAGE;
+			$status = KalturaBatchJobStatus::FINISHED;
+			$data = null;
+		}
+		else
+		{
+			$errType = KalturaBatchJobErrorTypes::KALTURA_API;
```
tan-tan-kanarek: This is not API failure, it's applicative failure.
Also, use here $this->closeJobOnError instead of $this->closeJob.

shpiNi: I agree with the KalturaBatchJobErrorTypes::APP.
closeJobOnError requires an Exception as a third argument.
I'm closing the job explicitly without an error so i'm using the closeJob function.

tan-tan-kanarek: So the simplest thing to do would be to throw kApplicativeException instead of closing it yourself.

PR #2789, opened 2015-07-02
In BaseEntry->copyInto the first thing that should be copied is the c…

noam-arad:
…ustom_data since some of the other copy functions sometimes try to get it from the new object and if it doesn't exist it is set as the default and if it is set we will not actually set it in the future, so this has to be done first.

PLAT-3325 #time 0.5 days

### DIFF
--- a/alpha/lib/model/om/Baseentry.php
+++ b/alpha/lib/model/om/Baseentry.php
@@ -4377,6 +4377,8 @@ abstract class Baseentry extends BaseObject  implements Persistent {
 	public function copyInto($copyObj, $deepCopy = false)
 	{
 
+		$copyObj->setCustomData($this->custom_data);
+		
 		$copyObj->setId($this->id);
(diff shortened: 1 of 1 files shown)


tan-tan-kanarek:
I'm sure that it didn't fix the issue because all setters in Baseentry are of real columns and not of custom-data, tests would reveal it. Also, the Baseentry is auto-generated class and as such shouldn't be edited manually, please find a different solution.

noam-arad:
I would like to assure you firstly that this has been tested and it works. 
The issue is caused because the method "entry->setKuserId" is called (not baseentry) and it in turn calls "setCreatorKuserId" which tries to get a value from custom_data, since there is no custom_data yet it creates a default one, once there is a m_custom_data the set doesn't override it.
Anyhow if it is auto-generated I will figure out a different way to fix.

PR #2799, opened 2015-07-05
Fix copying of entry, first copy the custom_data then any field that …

noam-arad:
…needs to be copied can be copied with use of the custom_data and not affect it.

PLAT-3325 #time 1 day


noam-arad:
@tan-tan-kanarek is this a better solution than https://github.com/kaltura/server/pull/2789?

tan-tan-kanarek:
It is.

PR #2816, opened 2015-07-07
Jupiter 10.15.0 plat 3364

noam-arad:


plugins/content_distribution/providers/cross_kaltura/lib/batch/CrossKalturaDistributionEngine.php
```diff
@@ -228,8 +228,9 @@ protected function getEntryObjects(KalturaClient $client, $entryId, KalturaDistr
 	    
 	    // get entry
 	    KalturaLog::debug('Getting entry id ['.$entryId.']');
-	    $entry = $client->baseEntry->get($entryId);
-	    
+		KBatchBase::impersonate($this->distributionProfile->partnerId);
```
tan-tan-kanarek: Instead of all these internal impersonations, just impersonate before calling the getEntryObjects method and unimpersonate after it.

noam-arad: No problem. The reason I didn't want to do it this way is because there are a few calls I didn't touch like for metadata objects and thumb assets. 
I'll test to verify the change doesn't affect functionality

PR #2820, opened 2015-07-07
mark createPeriodicSyncPoints as deprecated and remove its logic sinc…

yuly-roberman:
…e sync points are sent automatically on live stream


tan-tan-kanarek:
Please also delete KalturaMediaServerCuePointsService and KalturaMediaServerCreateTimeCuePointsResponse.

PR #2843, opened 2015-07-12
Jupiter 10.15.0 user id from ks

noam-arad:
Needs to be pulled to the new branch Jupiter-10.16.0 when it will be created.

### DIFF
--- a/api_v3/lib/types/filters/base/KalturaUserEntryBaseFilter.php
+++ b/api_v3/lib/types/filters/base/KalturaUserEntryBaseFilter.php
@@ -74,7 +74,7 @@ abstract class KalturaUserEntryBaseFilter extends KalturaFilter
 	public $entryIdNotIn;
 
 	/**
-	 * @var int
+	 * @var string
 	 */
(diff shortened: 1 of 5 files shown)


api_v3/lib/types/filters/base/KalturaUserEntryBaseFilter.php
```diff
@@ -74,7 +74,7 @@ public function getOrderByMap()
 	public $entryIdNotIn;
 
 	/**
-	 * @var int
+	 * @var string
```
tan-tan-kanarek: Noam, this class is auto-generated, please revert all manual changes and apply them elsewhere.

noam-arad: Please note that although there is still a change it was auto-generated because there was a change in KalturaUserEntry. The rest of the code was moved elsewhere

noam-arad:
merged in https://github.com/kaltura/server/pull/2855

PR #2859, opened 2015-07-13
add copying of reference id so that when we clone a youtube entry it …

noam-arad:
…will still be able to play since this fieldholds the id of the youtube video.
PLAT-3314


alpha/lib/model/entry.php
```diff
@@ -3390,5 +3390,6 @@ public function copyInto($copyObj, $deepCopy = false)
 		parent::copyInto($copyObj,$deepCopy);
 		$copyObj->setEntitledPusersEdit($this->getEntitledPusersEdit());
 		$copyObj->setEntitledPusersPublish($this->getEntitledPusersPublish());
+		$copyObj->setReferenceID($this->getReferenceID());
```
tan-tan-kanarek: Although we don't enforce the uniqueness of reference id, we treat it as identifier of the entry and therefore shouldn't be copied.
I understand the use case in case of external media entry, so implement it only there.

PR #2870, opened 2015-07-14
Prevent webex drop-folder from being watched or proccessed on backup

tan-tan-kanarek:


tan-tan-kanarek:
Eran, please review for HF.

PR #2873, opened 2015-07-14
Jupiter 10.15.0 plat 2595

hilak:


hilak:
@tan-tan-kanarek  @assafb please review

api_v3/services/ResponseProfileService.php
```diff
@@ -188,4 +188,30 @@ function recalculateAction(KalturaResponseProfileCacheRecalculateOptions $option
 	{
 		return KalturaResponseProfileCacher::recalculateCacheBySessionType($options);
 	}
+	
+	/**
+	 * List response profiles by filter and pager
```
tan-tan-kanarek: Please fix the description.

api_v3/services/ResponseProfileService.php
```diff
@@ -188,4 +188,30 @@ function recalculateAction(KalturaResponseProfileCacheRecalculateOptions $option
...
+	 * @return KalturaResponseProfile
+	 */
+	function cloneAction ($id)
+	{
+		$origResponseProfileDbObject = ResponseProfilePeer::retrieveByPK($id);
+		if (!$origResponseProfileDbObject)
+			throw new KalturaAPIException(KalturaErrors::RESPONSE_PROFILE_ID_NOT_FOUND, $id);
+			
+		$newResponseProfileDbObject = $origResponseProfileDbObject->copy();
+		$newResponseProfileDbObject->setSystemName($origResponseProfileDbObject->getSystemName() . '_' . uniqid());
```
tan-tan-kanarek: Instead of generating system-name, add KalturaResponseProfile as second argument to the action (same as playlist.clone and eventNotificationTemplate.clone) and enforce system-name to be non-null.

api_v3/services/ResponseProfileService.php
```diff
@@ -188,4 +188,30 @@ function recalculateAction(KalturaResponseProfileCacheRecalculateOptions $option
...
+		
+		$newResponseProfile = new KalturaResponseProfile();
+		$newResponseProfileDbObject = $newResponseProfile->toInsertableObject($origResponseProfileDbObject);
```
tan-tan-kanarek: toInsertableObject doesn't make sense here, that method should validate for insert and then call toObject.
The validate doesn't make sense because it's validating the KalturaResponseProfile which is empty in this case and the toObject populate the $origResponseProfileDbObject which already filled in this case.

PR #2892, opened 2015-07-19
Handle case of more than one cookie in header.

MosheMaorKaltura:
Jupiter-10.16.0-SUP-5363#time 1d


MosheMaorKaltura:
@tan-tan-kanarek  please validate

PR #2907, opened 2015-07-21
Enable generating clients from external XML

tan-tan-kanarek:
Also PHP5 client supports:
 - Path instead of service and action.
 - Path parameters.
 - Hierarchical services
 - JSON request and response


tan-tan-kanarek:
To review, use https://github.com/kaltura/server/pull/2907/files?w=0

PR #2914, opened 2015-07-22
kwidget should return caching header for CDN

matsuu:
I found that CDN has no cache for /index.php/kwidget/wid/_101/uiconf_id/... because alpha/web/index.php returns 'Cache-Control: private'.

I propose to replace private to public for CDN caching.

### DIFF
--- a/alpha/web/index.php
+++ b/alpha/web/index.php
@@ -129,7 +129,7 @@ function checkCache()
 				$max_age = 60 * 10;
 				header("X-Kaltura:cached-dispatcher");
 				header("Content-Type: application/x-shockwave-flash");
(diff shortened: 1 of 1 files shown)


erankor:
Hi @matsuu, thank you for your contribution, we will not be merging this change at this point. The reason is that we had issues in the past with proxies that held the responses longer than the max-age we return. This caused all sorts of problems, like the use expired kaltura sessions or CDN tokens. For this reason, we return Cache-Control private for all responses that may change. Only static content like video URLs or thumbnails is returned with public (for these we generate a new URL on every change).
Btw, we are using Akamai for caching these calls, and they support caching private responses (it's configurable)

PR #2928, opened 2015-07-28
SaaS + edgeServer Auto failover support

yossipapi:


alpha/lib/model/MediaServer.php
```diff
@@ -190,4 +190,8 @@ public function setIsExternalMediaServer($v)
 		$this->isExternalMediaServer = $v;
 	}
 	
+	public function getIsExternalMediaServer()
```
tan-tan-kanarek: I don't see where you use this getter.

alpha/lib/model/DeliveryProfileLiveAppleHttp.php
```diff
@@ -147,7 +147,7 @@ protected function isDvrContent($content) {
 		return in_array(self::M3U8_PLAYLIST_END_LIST_IDENTIFIER, array_map('trim', explode("\n", $content)));
 	}
 	
-	public function finalizeUrls(&$baseUrl, &$flavorsUrls)
+	public function disableExtraAttributes(&$baseUrl)
```
tan-tan-kanarek: I would leave DeliveryProfileLiveAppleHttp::finalizeUrls as is instead of changing it to disableExtraAttributes and calling it separately.

alpha/lib/model/MediaServer.php
yossipapi: U R right ... I used to use it but I changed the code and it is no longer needed. I will remove this function.

alpha/lib/model/DeliveryProfileLiveAppleHttp.php
yossipapi: Finalize url build the edge server url so I cannot do it before the curl.

PR #2929, opened 2015-07-28
Validate that tags is part of the XML

MosheMaorKaltura:
PLAT-3088#4h


plugins/cue_points/base/batch/CuePointBulkUploadXmlHandler.php
```diff
@@ -207,16 +207,17 @@ protected function parseCuePoint(SimpleXMLElement $scene)
...
-			$value = "$tag";
-			if($value)
-				$tags[] = $value;
+			$tags = array();
+			foreach ($scene->tags->children() as $tag) {
+				$value = "$tag";
+				if ($value)
+					$tags[] = $value;
+			}
+			$cuePoint->setTags(implode(',', $tags));
```
tan-tan-kanarek: $cuePoint is of type KalturaCuePoint and doesn't have setters, you should set the public tags property, or n other words, just revert this change.

PR #2931, opened 2015-07-28
Jupiter 10.17.0 sup 5223

noam-arad:


tan-tan-kanarek:
@noam-arad, I think it doesn't make sense to change such important and widely used method (KalturaEntryService::add) in a poor way just to be compatible with a deprecated old action (BaseEntryService::addFromUploadedFileAction).
Also, I don't understand why you had to change KalturaEntryService::add, you could simply pass the KalturaBaseEntry $entry into it.

api_v3/services/BaseEntryService.php
```diff
@@ -851,4 +854,19 @@ function cloneAction( $entryId )
 
 		return $this->getEntry($clonedEntry->getId());
 	}
+
```
tan-tan-kanarek: Please move this method to KalturaEntryService and call it also from KalturaEntryService::add instead of having duplicated code.

PR #2953, opened 2015-08-02
Jupiter 10.17.0 plat 3477

noam-arad:


noam-arad:
Should be merged to next version (10.18.0)

PR #2963, opened 2015-08-03
Fix from code review

noam-arad:
SUP-5223 #time 10 minutes

### DIFF
--- a/api_v3/lib/KalturaEntryService.php
+++ b/api_v3/lib/KalturaEntryService.php
@@ -975,9 +975,9 @@ class KalturaEntryService extends KalturaBaseService
 			if ($templateEntry)
 			{
 				$dbEntry = $templateEntry->copyTemplate(true);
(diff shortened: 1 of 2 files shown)


api_v3/lib/KalturaEntryService.php
```diff
@@ -957,23 +957,27 @@ protected function prepareEntryForInsert(KalturaBaseEntry $entry, entry $dbEntry
...
-			if($templateEntry)
+			if ($templateEntry)
 			{
 				$dbEntry = $templateEntry->copyTemplate(true);
-				$dbEntry->save();
-			}
-			else
-			{
-				KalturaLog::err("Template entry id [" . $conversionProfile->getDefaultEntryId() . "] not found");
+				return $dbEntry;
```
tan-tan-kanarek: return $dbEntry at the end of the mothod in order to return null in case it was never set.

PR #2964, opened 2015-08-03
misc fixes

erankor:
- change the priority of request parameters - path, query string, body
- lock widget session user id to 0
- remove ks generation from kmc.reports ps2 action

### DIFF
--- a/alpha/apps/kaltura/lib/request/infraRequestUtils.class.php
+++ b/alpha/apps/kaltura/lib/request/infraRequestUtils.class.php
@@ -455,7 +455,7 @@ class infraRequestUtils
 			$params[$key['value']] = $value['value'];
 		}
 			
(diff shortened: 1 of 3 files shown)


alpha/apps/kaltura/lib/request/infraRequestUtils.class.php
```diff
@@ -455,7 +455,7 @@ public static function getRequestParams()
 			$params[$key['value']] = $value['value'];
 		}
 			
-		self::$requestParams = array_merge($params, $_GET, $_POST, $_FILES);
+		self::$requestParams = array_merge($_FILES, $_POST, $_GET, $params);
```
tan-tan-kanarek: How would you prioritize them against JSON?
See: https://github.com/kaltura/server/pull/2907/files#diff-a40d9041ddca11fe5c490619bdbecf7cR475

PR #2975, opened 2015-08-05
SUP-4363

EladCohenKaltura:
SUP-4363
Chnage serveByEntryIdAction to accept zero as argument
 which will lead to the same results as null
 since the 0 asset id is always the source


EladCohenKaltura:
- if ($captionAsset->getLanguage() == KalturaLanguage::MU)
-  {
-	kCaptionsContentManager::addParseMultiLanguageCaptionAssetJob($captionAsset, $finalPath);

This marked as change is not part of the change

PR #2980, opened 2015-08-05
When an exception is thrown from the API we have a PHP Fatal, then if…

noam-arad:
… there is a problem with a specific liveStream then all the live streams after that will not get handled


batch/batches/ValidateLiveMediaServers/KAsyncValidateLiveMediaServers.class.php
```diff
@@ -47,12 +47,19 @@ public function run($jobs = null)
...
+					/* @var $entry KalturaLiveEntry */
+					self::impersonate($entry->partnerId);
+					self::$kClient->liveStream->validateRegisteredMediaServers($entry->id);
+					self::unimpersonate();
+					$filter->createdAtGreaterThanOrEqual = $entry->createdAt;
+				}
+				catch (KalturaException $e)
+				{
+					self::unimpersonate();
+					KalturaLog::err("Caught exception with message [" . $e->getMessage()."]");
```
tan-tan-kanarek: Please replace with KalturaLog::err($e);

PR #2998, opened 2015-08-11
calculate quiz score correctly

noam-arad:


plugins/cue_points/quiz/lib/model/QuizUserEntry.php
```diff
@@ -23,39 +23,53 @@ public function setScore($v){ $this->putInCustomData("score", $v);}
...
+			$answersCrit->add(CuePointPeer::ENTRY_ID,$this->getEntryId());
+			$answersCrit->add(CuePointPeer::PARENT_ID, $question->getId());
+			$answers = CuePointPeer::doSelect($answersCrit);
+			$currAnswer = null;
+			foreach ($answers as $answer)
+			{
+				/**
+				 * @var AnswerCuePoint $answer
+				 */
+				if ($answer->getQuizUserEntryId() == $this->id)
```
tan-tan-kanarek: Can't the resulted $answers array be huge? Seems like the wrong way to filter it.

PR #3017, opened 2015-08-13
Jupiter 10.18.0 ivq cuepoint add

noam-arad:


noam-arad:
instead of creating a new service I will allow adding+updating cue points for player and widget KS and then enforce the restrictions on the object level.

PR #3035, opened 2015-08-17
Update FileSyncSubTypeTranslate.php

DBezemer:
Update FileSync Client definition to reflect changes in Enum structure.


tan-tan-kanarek:
What server change would cause that?

DBezemer:
@tan-tan-kanarek I think this is when the Client Lib structure was updated for 3.1.2 or so, has been broken for forever, you can see this in the Admin Console on every Batch Job

tan-tan-kanarek:
@MosheMaorKaltura, please approve merge.

DBezemer:
@MosheMaorKaltura rebased as this pull is still valid after almost 3 years

PR #3039, opened 2015-08-17
Jupiter 10.18.0 ivq cuepoint add

noam-arad:


tan-tan-kanarek:
Please implement using requiresPermission on the cue-point object and enable only answer cue-point, we don't want to implement different actions for different cue-point types.

PR #3040, opened 2015-08-17
Don't re-index the entry in case of cuePoint's indexOnEntry is false

yossipapi:


plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -512,11 +514,14 @@ public static function isPostProcessCuePointsEvent(BaseObject $object, array $mo
 		return false;
 	}
 	
-	public static function shouldReIndexEntry(BaseObject $object, array $modifiedColumns)
+	public static function shouldReIndexEntry(BaseObject $object, array $modifiedColumns = array())
 	{
 		if(!($object instanceof CuePoint))
 			return false;
 		
+		if($object instanceof Annotation && $object->getSearchableOnEntry() == false)
```
tan-tan-kanarek: You shouldn't reference Annotation outside annotation plugin.

PR #3047, opened 2015-08-18
Remove update of notificationResult on the batch job itself

hilak:
Reserve update write to worker log


hilak:
@erankor this was requested some time ago. Can you please approve? I will merge it to Jupiter-10.19.0.

PR #3053, opened 2015-08-19
Add the ability to decide weather specific users' entries should go t…

noam-arad:
…hrough moderation. The users list can be defined by adding a filter to the partner configuration to the field autoModerateEntryFilter of type KalturaBaseEntryFilter

with the field userIdIn set to the user names list comma seperated.
SUP-4934 #time 3 hours


MosheMaorKaltura:
Approved, please merge to 10.19.0

PR #3082, opened 2015-08-25
Remove pretty useless ListBuckets operation

kubrickfr:
Doing a ListBuckets() to check the credentials is pretty useless: it costs money and time and  only checks that the user has the s3:ListAllMyBuckets permission which we don't use anywhere else in the code anyway (and we don't want if for security reasons, principle of "least privilege"). The code will fail soon enough in the code elsewhere if the permissions are not sufficient.
See: 
* https://aws.amazon.com/articles/1904 "Avoiding Unnecessary Requests"
* http://docs.aws.amazon.com/IAM/latest/UserGuide/IAMBestPractices.html#grant-least-privilege
* http://stackoverflow.com/a/29367812

### DIFF
--- a/infra/storage/file_transfer_managers/s3Mgr.class.php
+++ b/infra/storage/file_transfer_managers/s3Mgr.class.php
@@ -72,18 +72,13 @@ class s3Mgr extends kFileTransferMgr
 				)
 		);
 		
-		$connectionSuccess = false;
-		try
-		{
-			$buckets = $this->s3->listBuckets(); // just to check whether the connection is good, throws an exception in case of error
-			$connectionSuccess = true;
-		}
-		catch ( Exception $e )
-		{
-			KalturaLog::err("Can't connect to S3: {$e->getMessage()}");
-		}
-
-		return $connectionSuccess;
+		/** 
+		 * There is no way of "checking the credentials" on s3.
+		 * Doing a ListBuckets would only check that the user has the s3:ListAllMyBuckets permission
+		 * which we don't use anywhere else in the code anyway. The code will fail soon enough in the
+		 * code elsewhere if the permissions are not sufficient.
+		 **/
+		return true;
 	}
 
 


kaltura-hooks:
Hi @kubrickfr,

kubrickfr:
Done

jessp01:
Hi @kubrickfr,

First, thank you for the pull.
Yes, indeed, no point in calling listBuckets() but I also don't feel comfortable with just returning true.
From a brief look at the API documentation, I don't see a call along the lines of a simple ping to verify connectivity. We could verify connectivity using a simple curl to the endpoint but ideally, we want to verify auth as well, especially since the function IS called doLogin().
I would have said lets drop doLogin() altogether but of course we cannot since the class extends kFileTransferMgr which declares:
abstract protected function doLogin($user, $pass);

Have you got any suggestions?

kubrickfr:
Hi, 

Sorry for the long time it took to reply, I missed your message.

I understand what your saying, however the class you're extending is obviously for a stateful protocol, like FTP, where a connection is established ans kept, this just doesn't make sense for a stateless protocol. There is indeed no way to "check connectivity" with S3 as even doing so would bring no guarantee that you have sufficient rights for the rest of the operations you want to perform.
What's more, this function is going to be even more useless and cumbersome if/when you try to support IAM roles (which I hope you do soon!). At which point there will be no credentials to pass to the function.
So I'm of the opinion that this method should be as much of a stub as possible for S3 and just return true. If the credentials are invalid, then when you try to PutObject (for example) then the error message will be "Invalid credentials" anyway, which is what you want I guess?
And you need to handle those errors correctly anyway, as if we take the extreme corner case of someone changing the credentials between the doLogin() and an operation, on FTP it would be fine as you have acquired a valid connection already, but it would fail with S3.

So I don't see any other option than returning true I'm afraid...

tan-tan-kanarek:
@jessp01, I tend to agree with @kubrickfr.
We will merge it, depends on QA availability.

jessp01:
I would suggest at least checking if the S3 endpoint is reachable then. While it is still a bit problematic due to the function name, which as I said in my first reply cannot be modified for obvious reasons, at least it is some sort of a check and IMHO, still better than just 'return true;'...

kubrickfr:
@jessp01 What are the benefits of doing any of these checks? Aren't we going to know soon enough when any of the real operations fail?

jessp01:
Only benefit I see is catching the failure in a function where you'd 
expect a failure like that to happen.

I agree it seems forced, If I were to write an interface for this from 
scratch, I'd design it differently considering the S3 API is different 
than others we handle using the same base class, but being what it is, I 
feel handling the failure at this point rather than later on makes more 
sense.
Plus, a simple HTTP request to check connectivity is really not that 
demanding, if it were, I'd speak differently:)

BTW, as a side note, just because an API is stateless, does not mean it 
necessarily should provide a connectivity check:) our API is also 
stateless and one needs to pass along a KS [Kaltura Session] string with 
each request but we do have the system->ping() call for testing 
connectivity and also, you can call session->start() to get back a KS, 
which is a very low resource operation..


May the source be with you,

Jess Portnoy

Kaltura Connect Virtual Summits 2015 - now on VOD! [connect.mediaspace.kaltura.com]

kubrickfr:
@jessp01 As you wish... But for me this expectation is flawed and will be even more so when there will be no credentials at all (IAM roles), which is the future.

Regarding what you say about connectivity check and stateless, I'd agree with you if Kaltura was written in a persistent language (like Java/Tomcat) but with php you'd have to check the connectivity every time, which is wasteful. Plus, in your example, KS makes it stateful :) that's the point of a session.

jessp01:
Regarding the issue at hand, perhaps it would make more sense to write a 
base class that does not have the doLogin() method at all, for connecting 
with S3 and similar services.

As for your comment about stateless, you do not need to check connectivity 
every time, only once, and when you get the KS back from session->start() 
the same KS can be used over and over in all subsequent calls so, IMHO, it 
is certainly stateless.

May the source be with you,

Jess Portnoy

Kaltura Connect Virtual Summits 2015 - now on VOD! [connect.mediaspace.kaltura.com]

tan-tan-kanarek:
Merged: https://github.com/kaltura/server/pull/3226

PR #3049, opened 2015-08-19
Support using only live cloud-transcode flavors if timeAlignedRenditions was set

danoss:
(PLAT-3588)

PR #3083, opened 2015-08-25
Jupiter 10.18.0 PLAT-3588

danoss:


tan-tan-kanarek:
See my comments in https://github.com/kaltura/server/pull/3049.

PR #3085, opened 2015-08-25
$player_uiconf_list[0] is null after installation

jorgevbo:
$player_uiconf_list[0] is null after installation. Checking if value exists


zoharbabin:
@jorgevbo - Thanks for the contrib!

I agree we should replace the short-tags with proper <?php tags. 
However, the suggested fix to check if the fields are empty and then set the width/height to empty strings is not recommended imo. We should understand the source of the issue and fix why you got to a situation where $player_uiconf_list[0] is null .

Can you please describe the installation flow you had, and whether you experienced any issues during the installation or deleted any files manually?

Thanks!

jessp01:
All short tags were replaced in these commits:
4767d46129a83771180102fd114108d677b65e94
360562c8afa61f7b9286a32a41f36f3201695ec4
Thanks @jorgevbo.

PR #3088, opened 2015-08-25
PLAT-3736

hilak:
1. Remove retry on ERROR of drop folder file
2. change the naming logic for feed item files


hilak:
@assafb Please review

PR #3069, opened 2015-08-23
Disregard answers for unsubmitted quizzes.

noam-arad:
PLAT-3705 #time 2 hours

PR #3090, opened 2015-08-26
Jupiter 10.18.0 plat 3705

noam-arad:


noam-arad:
see comments here: https://github.com/kaltura/server/pull/3069

PR #3092, opened 2015-08-26
Jupiter 10.17.0.media prep1.1

EladCohenKaltura:


plugins/content_distribution/providers/tvinci/TvinciDistributionPlugin.php
```diff
@@ -191,8 +198,15 @@ public static function getKalturaProvider()
 	 */
 	public static function contributeMRSS(EntryDistribution $entryDistribution, SimpleXMLElement $mrss)
```
MosheMaorKaltura: Check if this function is being called, As it looks, need to remove it.

plugins/content_distribution/providers/tvinci/TvinciDistributionPlugin.php
```diff
@@ -191,8 +198,15 @@ public static function getKalturaProvider()
...
 	{
+
+	}
+
+	public function getObjectFeatureType ()
```
MosheMaorKaltura: Lets discuss both functions  getObjectFeatureType  & contribute if we need them at all, since the parent function will be called if they are gone.

plugins/content_distribution/providers/tvinci/lib/feed/TvinciDistributionFeedHelper.php
```diff
@@ -144,47 +47,47 @@ public function buildDeleteFeed()
...
+    		"iphonenewPpvModule" => $this->distributionProfile->iphonenewPpvModule,
+    		"iphonenewTypeName" => $this->distributionProfile->iphonenewFileName,
+    		"mbrPpvModule" => $this->distributionProfile->mbrPpvModule,
+    		"mbrTypeName" => $this->distributionProfile->mbrFileName,
+		    "dashPpvModule" => $this->distributionProfile->dashPpvModule,
+    		"dashTypeName" => $this->distributionProfile->dashFileName
+		);
+	}
+
 	private function createXml($action)
```
MosheMaorKaltura: Remove parameter $action hence it is not needed.

plugins/content_distribution/providers/tvinci/TvinciDistributionPlugin.php
```diff
@@ -2,7 +2,8 @@
 /**
  * @package plugins.tvinciDistribution
  */
-class TvinciDistributionPlugin extends KalturaPlugin implements IKalturaPermissions, IKalturaEnumerator, IKalturaPending, IKalturaObjectLoader, IKalturaContentDistributionProvider, IKalturaConfigurator
+class TvinciDistributionPlugin extends KalturaParentContributedPlugin implements IKalturaPermissions, IKalturaEnumerator, IKalturaPending, IKalturaObjectLoader,
```
tan-tan-kanarek: Please add plugin dependency on KalturaParentContributedPlugin.

EladCohenKaltura: But this class is abstract and cannot be part of the plugin list - if I understand it correctly the dependency can work only on plugins that can appear in the plugins.ini and this one cannot (same as KalturaApplicationPlugin)

tan-tan-kanarek: You're right, and in fact it doesn't make sense to implement it as parent class.
If it's a standalone feature, implement it as a standalone plugin, or as a core feature, this way different plugins will be able to use it even if they already extend a different parent.

EladCohenKaltura: Since the influence here is about how the MRSS is built it should be structured as an hierarchy implementation - since you want to influence "yourself"

EladCohenKaltura: i think I got your point now  - some sort of a helper/util code that gives the children structure and can be called from the contribute of plugin itself 
if this is the case you are referring to - where would you place this code ? in the same location ?

tan-tan-kanarek: Consider to add it as core feature and not as plugin.

PR #3093, opened 2015-08-26
HF-3078

EladCohenKaltura:
HF-3078


EladCohenKaltura:
This HotFix will not occur - instead will be pushed in the next version Jupiter-10.19.0

PR #3097, opened 2015-08-27
Jupiter 10.19.0 sup5523

Galmoss:
Add partner id to google auth url and using it in the loginByLoginId action

### DIFF
--- a/alpha/web/lib/js/googleoauth2.js
+++ b/alpha/web/lib/js/googleoauth2.js
@@ -34,8 +34,8 @@
 				service: 'user',
 				action: 'loginByLoginId',
 				loginId: $email.val(),
-                password: $password.val(),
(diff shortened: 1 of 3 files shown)


alpha/web/lib/js/googleoauth2.js
```diff
@@ -34,11 +34,24 @@
...
 		}
+		
+		function getQueryVariable(variable) {
+	               var query = window.location.search.substring(1);
+	               var vars = query.split("&");
+	               for (var i=0;i<vars.length;i++) {
+	                       var pair = vars[i].split("=");
+	                       if(pair[0] == variable){return pair[1];}
+               		}
+               		return(false);
```
MosheMaorKaltura: If var is not found then need to raise exception.

Galmoss:
@MosheMaorKaltura  done.

PR #3103, opened 2015-08-27
Use $scope->getObject instead of $scope->getEvent

Galmoss:


Galmoss:
As you requested @tan-tan-kanarek

PR #3104, opened 2015-08-28
PLAT-3758

hilak:
PARSED drop folder files not created if drop folder conversion profile XSLT contains calls to authorized php function


hilak:
@assafb  Please review

PR #3107, opened 2015-08-30
Jupiter 10.19.0 plat 3718

MosheMaorKaltura:


alpha/apps/kaltura/lib/request/infraRequestUtils.class.php
```diff
@@ -173,12 +173,15 @@ public static function sendCdnHeaders($ext, $content_length, $max_age = 8640000
 				case "ts":
 					$content_type ="video/MP2T";
 					break;
-						case "3gp":
-								$content_type ="video/3gpp";
-								break;
+				case "3gp":
```
tan-tan-kanarek: I think we better externalize this list to base.ini.

PR #2982, opened 2015-08-09
Jupiter 10.17.0 plat 2599

shpiNi:

PR #3112, opened 2015-08-30
Cielo24 plugin

shpiNi:


deployment/updates/scripts/add_permissions/2015_09_20_update_integration_notify_permission_name.php
```diff
@@ -0,0 +1,20 @@
...
+$config = realpath(dirname(__FILE__)) . '/../../../permissions/partner.0.ini';
+passthru("php $script $config");
+
+$config = realpath(dirname(__FILE__)) . '/../../../permissions/service.integration.integration.ini';
+passthru("php $script $config");
+
+require_once(dirname(__FILE__) . "/../../../../" . "alpha/scripts/bootstrap.php");
+
+$userRole = UserRolePeer::getByNameAndPartnerId(kIntegrationFlowManager::EXTERNAL_INTEGRATION_SERVICES_ROLE_NAME, PartnerPeer::GLOBAL_PARTNER);
+$userRole->setPermissionNames("INTEGRATION_ACTIONS");
```
rkreich: @shpiNi, I would change it to remove `VOICEBASE_ACTIONS` and add `INTEGRATION_ACTIONS` instead of overwriting it.
If this role will be update with a new permission in the future, and this script will be executed, it will override the new permission.
Just explode it by `,`, `trim` the values, unset `VOICEBASE_ACTIONS` and add `INTEGRATION_ACTIONS`

plugins/integration/providers/cielo24/Cielo24ClientHelper.php
```diff
@@ -0,0 +1,164 @@
...
+		$baseUrl = $cielo24ParamsMap['base_url'];
+		$version = $cielo24ParamsMap['version'];
+		$baseUrl = $baseUrl . "account/login?v=$version";
+		$loginParams = array("username" => $username, "password" => $password);	
+		$loginUrl = $this->addUrlParams($baseUrl, $loginParams);
+	
+		$loginResult = self::sendAPICall($loginUrl);
+		$apiToken = $loginResult->ApiToken;
+			
+		$this->baseEndpointUrl = str_replace("account/login", self::API_CALL_NAME_PATTERN, $baseUrl);
```
rkreich: @shpiNi why do you need `API_CALL_NAME_PATTERN`? it seems like an unneeded string transformation that you are doing, you can write the code better to avoid it.
Always keep the `$this->baseEndpointUrl` untouched.

plugins/integration/providers/cielo24/Cielo24ClientHelper.php
```diff
@@ -0,0 +1,164 @@
...
+		$requestTranscriptionResult = $this->sendAPICall($requestTransctiptAPIUrl);
+		if(!$requestTranscriptionResult || !isset($requestTranscriptionResult->TaskId))
+			return false;
+			
+		return true;
+	}
+	
+	private function sendAPICall($url, $noDecoding = false)
+	{
+		KalturaLog::debug("sending API call - $url");
```
rkreich: @shpiNi you should pass the API name and the parameters to this function and build the full URL internally.
This will solve the `API_CALL_NAME_PATTERN` constant.

plugins/integration/providers/cielo24/Cielo24Plugin.php
```diff
@@ -0,0 +1,121 @@
...
+	 */	
+	public static function getPartnerCielo24Options($partnerId)
+	{
+		$partner = PartnerPeer::retrieveByPK($partnerId);
+		if(!$partner)
+			return null;
+		return $partner->getFromCustomData(Cielo24IntegrationProviderType::CIELO24);
+	}
+	
+	public static function setPartnerCielo24Options($partnerId, Cielo24Options $options)
```
rkreich: @tan-tan-kanarek what do you think about serializing a custom object into custom data? I think that we usually don't do it like this.

plugins/integration/services/IntegrationService.php
```diff
@@ -50,7 +50,13 @@ public function notifyAction($id)
 		$batchJob = BatchJobPeer::retrieveByPK($id);
 		$invalidJobId = false;
 		$invalidKs = false;
-		if(!$batchJob)
+		
+		if(!self::validateKs($batchJob))
```
assafb: @shpiNi  - first need to check if(!$batchJob) and then if(!self::validateKs($batchJob))

plugins/integration/providers/cielo24/lib/api/KalturaCielo24JobProviderData.php
```diff
@@ -0,0 +1,159 @@
...
+		$cielo24ParamsMap = kConf::get('cielo24','integration');
+		if($this->captionAssetFormats)
+		{
+			$formatsString = $this->captionAssetFormats;
+			$formatsArray = explode(',', $formatsString);
+			
+			$excludedFormats = $cielo24ParamsMap['exclude_formats'];
+			foreach($formatsArray as $format)
+			{
+				if(!constant("KalturaCaptionType::" . $format) || in_array($format, $excludedFormats))
```
rkreich: @shpiNi you need to sanitize `$format`.
Please do the same fix in VoiceBase.

plugins/integration/services/IntegrationService.php
shpiNi: @assafb it's intended to increase security.

plugins/integration/providers/cielo24/Cielo24Plugin.php
shpiNi: @rkreich I was asked explicitly to use an object in the same context: https://github.com/kaltura/server/pull/2982#discussion_r36587101

tan-tan-kanarek: The entire custom-data object is already serialized.

plugins/integration/providers/cielo24/lib/kCielo24FlowManager.php
```diff
@@ -0,0 +1,158 @@
...
+			$captionsContentArray = $clientHelper->getRemoteCaptions($remoteJobId, $formatsArray);
+			KalturaLog::debug("captions content - " . print_r($captionsContentArray, true));
+	
+			$captions = $this->getAssetsByLanguage($entryId, array(CaptionPlugin::getAssetTypeCoreValue(CaptionAssetType::CAPTION)), $spokenLanguage);
+		
+			$this->setObjectContent($transcript, $transcriptContent, null, true);
+	
+			foreach ($captionsContentArray as $format => $content)
+			{        
+				$captionFormatConst = constant("KalturaCaptionType::" . $format);
```
rkreich: @shpiNi can we trust `$format` here?

plugins/integration/providers/cielo24/lib/api/KalturaCielo24JobProviderData.php
```diff
@@ -0,0 +1,159 @@
...
+	
+		$formatsString = $object->getCaptionAssetFormats();
+		if($formatsString)
+		{
+			$formatsArray = explode(',', $formatsString);
+			$excludedFormats = $cielo24ParamsMap['exclude_formats'];
+			$sanitizedFormatsArray = array();
+			foreach($formatsArray as $format)
+			{
+				$format = preg_replace("/[^A-Z]/", "", $format);
```
rkreich: @shpiNi for forward compatibility, I would add `_` as whitelisted characater as well

plugins/integration/providers/cielo24/lib/kCielo24FlowManager.php
shpiNi: @rkreich yes, it's a reflection of the providerData's 'CaptionAssetFormats' field which is being validated in the object's constructor.

PR #3116, opened 2015-08-31
Add option to manually set hls/hds url for akamai universal live

Galmoss:
SUP-5673


Galmoss:
@MosheMaorKaltura  can you please merge, so we will test it with QA ?

PR #3119, opened 2015-08-31
PLAT-3769

EladCohenKaltura:
PLAT-3769


plugins/content_distribution/providers/tvinci/lib/xml/tvinci_default.xslt
```diff
@@ -411,42 +411,43 @@
                     <xsl:with-param name="matchingContents" select="$tagMatchingContents"/>
                 </xsl:call-template>
             </xsl:variable>
+            <xsl:if test="not(count(exsl:node-set($filteredItems)/contentItem) = 0)">
```
EladCohenKaltura: this is the line that has changed - the rest is tab shift

PR #3121, opened 2015-08-31
New ACL action "redirect player calls to Kaltura via internal server"

yossipapi:


alpha/apps/kaltura/lib/access_control/model/actions/kAccessControlAddRequestHostPrefixAction.php
```diff
@@ -0,0 +1,33 @@
+<?php
+/**
+ * @package Core
+ * @subpackage model.data
+ */
+class kAccessControlAddRequestHostPrefixAction extends kRuleAction
```
tan-tan-kanarek: What if we will want in the future a suffix or infix? I would name it differently and support any URL modification.

PR #3125, opened 2015-09-02
voicebase improvements

shpiNi:


plugins/integration/providers/voicebase/lib/kVoicebaseFlowManager.php
```diff
@@ -65,10 +74,10 @@ public function updatedJob(BatchJob $dbBatchJob)
 			$formatsArray[] = "TXT";
 			$contentsArray = $clientHelper->getRemoteTranscripts($entryId, $formatsArray);
 			KalturaLog::debug('contents are - ' . print_r($contentsArray, true));
-			$transcript = $this->getAssetsByLanguage($entryId, array(TranscriptPlugin::getAssetTypeCoreValue(TranscriptAssetType::TRANSCRIPT)), $spokenLanguage, true);
 			$captions = $this->getAssetsByLanguage($entryId, array(CaptionPlugin::getAssetTypeCoreValue(CaptionAssetType::CAPTION)), $spokenLanguage);
-		
-			$this->setObjectContent($transcript, $contentsArray["TXT"], null, true);
+			$accuracy = $clientHelper->calculateAccuracy($entryId);
```
shpiNi: @rkreich should we round up/down the result here?

rkreich: @zoharbabin, what do you think?

zoharbabin: @rkreich @shpiNi - why do we need to round at all? I would just cut after two decimals (0.862328 will become 0.86)

jessp01:
I think this condition:
if(!$entry || $entry->getType() != entryType::MEDIA_CLIP || !in_array($entry->getMediaType(), array(entry::ENTRY_MEDIA_TYPE_VIDEO,entry::ENTRY_MEDIA_TYPE_AUDIO)))

Can be simplified with:
if(!$entry || !in_array($entry->getMediaType(), array(entry::ENTRY_MEDIA_TYPE_VIDEO,entry::ENTRY_MEDIA_TYPE_AUDIO)))

Also:
die("Usage: php addVoicebaseParamsToPartner [partner id] [apiKey] [apiPassword]" . PHP_EOL);

Should use $argv[0] instead of hardcoded filename.

PR #3131, opened 2015-09-06
Jupiter 10.19.0 alter scheduler status bigint

kobimichaeli:
@tan-tan-kanarek please check (sorry for the multiple commits, only the last one needs to be checked)


tan-tan-kanarek:
Please apply the DB change also in deployment/base/sql/01.kaltura_ce_tables.sql.

kobimichaeli:
Will be re-merged into 10.20.0

PR #3148, opened 2015-09-07
PLAT-3753

EladCohenKaltura:
PLAT-3753
removed spaces to allow multi object ids even when SQL script does not exist for IN video QUIZ


plugins/cue_points/quiz/QuizPlugin.php
```diff
@@ -819,3 +819,7 @@ function innerCompare($a, $b)
 	}
 	return ($prctgA < $prctgB) ? -1 : 1;	
 }
+
+function getObjectIdsAsArray($objectIds){
```
tan-tan-kanarek: Please move it into the class as static method, we try to avoid functions.

PR #3153, opened 2015-09-08
Jupiter 10.20.0 plat 3838

sabanas:


alpha/lib/model/LiveEntry.php
```diff
@@ -539,6 +541,22 @@ public function getMediaServer($currentDcOnly = false)
 		KalturaLog::debug("No Valid Media Servers Were Found For Current Live Entry [" . $this->getEntryId() . "]" );
 		return null;
 	}
+
+	public function getHostnames()
```
tan-tan-kanarek: Please rename to getMediaServersHostnames and make it protected.

alpha/apps/kaltura/lib/model/objectfilters/entryFilter.class.php
```diff
@@ -722,6 +723,11 @@ public function setIsLive($v)
 	{
 		$this->set('_is_live', intval($v));
 	}
+
+	public function setMediaServerHostname($v)
```
tan-tan-kanarek: This method is redundant.

alpha/lib/model/LiveEntry.php
```diff
@@ -694,13 +712,24 @@ public function getMediaServers()
...
 				LiveEntry::FIRST_BROADCAST => $this->getFirstBroadcast(),
 				LiveEntry::RECORDED_ENTRY_ID => $this->getRecordedEntryId(),
+
 		);
-		
-		return array_merge( $dynamicAttributes, parent::getDynamicAttributes() ); 
+
+		$mediaServers = $this->getMediaServersHostnames();
+		$hostnamesCount = count($mediaServers);
+		if ($hostnamesCount > 0) {
```
tan-tan-kanarek: What if the count is 1 and the only index is 1? I think you better check isset($mediaServers[MediaServerIndex::PRIMARY]) instead of $hostnamesCount > 0, same for secondary.

PR #3164, opened 2015-09-09
Override setKuserId function to take the correct value (from userEntry) object

danoss:


plugins/cue_points/quiz/lib/model/AnswerCuePoint.php
```diff
@@ -52,6 +52,11 @@ public function copyToClipEntry( entry $clipEntry, $clipStartTime, $clipDuration
 	{
 		return false;
 	}
+	
+    public function setKuserId($v) {
+		$userEntry = UserEntryPeer::retrieveByPK($this->getQuizUserEntryId());
+		parent::setKuserId($userEntry->getKuserId());
```
tan-tan-kanarek: Please verify that $userEntry is not null.

PR #3166, opened 2015-09-09
SUP-5904-HOTFIX

EladCohenKaltura:
HF-SUP-5904
Changed self:: to this-> in an abstract class


yuly-roberman:
@EladCohenKaltura looks good

PR #3170, opened 2015-09-10
Allow access only to user's userEntries and return empty list when

danoss:
userId was not given

### DIFF
--- a/alpha/lib/model/UserEntryPeer.php
+++ b/alpha/lib/model/UserEntryPeer.php
@@ -19,6 +19,23 @@ class UserEntryPeer extends BaseUserEntryPeer {
 	protected static $class_types_cache = array(
 	);
 
+	/* (non-PHPdoc)
+	 * @see BaseUserEntryPeer::setDefaultCriteriaFilter()
+	 */
+	public static function setDefaultCriteriaFilter()
+	{
+	    // when session is not admin, allow access to user's userEntries only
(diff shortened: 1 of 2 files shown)


alpha/lib/model/UserEntryPeer.php
```diff
@@ -19,6 +19,23 @@ class UserEntryPeer extends BaseUserEntryPeer {
 	protected static $class_types_cache = array(
 	);
 
+	/* (non-PHPdoc)
+	 * @see BaseUserEntryPeer::setDefaultCriteriaFilter()
+	 */
+	public static function setDefaultCriteriaFilter()
+	{
+	    // when session is not admin, allow access to user's userEntries only
```
tan-tan-kanarek: This method should always create a default criteria, so place the creation of the criteriaFilter and the setFilter outside the if statement.

api_v3/services/UserEntryService.php
```diff
@@ -83,6 +83,10 @@ public function listAction(KalturaUserEntryFilter $filter, KalturaFilterPager $p
 		{
 			$pager = new KalturaFilterPager();
 		}
+		// return empty list when userId was not given
+		if ( $this->getKs() && !$this->getKs()->isAdmin() && (!kCurrentContext::$ks_uid || kCurrentContext::$ks_uid == "") ) {
```
tan-tan-kanarek: (kCurrentContext::$ks_uid == "") is equivalent to (!kCurrentContext::$ks_uid).

PR #3178, opened 2015-09-10
Add script and workflow to set specific thumbnail's cache age for par…

Galmoss:
…tner

SUP-5457


alpha/apps/kaltura/modules/extwidget/actions/thumbnailAction.class.php
```diff
@@ -458,6 +461,10 @@ public function execute()
 		{
```
MosheMaorKaltura: Check of partnerCacheAge is configured and handle it before the case of _NOCACHE_

Galmoss:
@MosheMaorKaltura done.

tan-tan-kanarek:
Reviewed.

PR #3181, opened 2015-09-10
Update kaltura_batch.sh

DBezemer:
Refactor batch init script to correctly cover start/restart/stop commands with correct status detection and pid/lockfile handling.

### DIFF
--- a/batch/kaltura_batch.sh
+++ b/batch/kaltura_batch.sh
@@ -1,10 +1,7 @@
 #!/bin/bash
-. /etc/kaltura.d/system.ini
-
-echo `date`
 
 #
-# batchMgr		This shell script takes care of starting and stopping a Kaltura Batch Service
+# batchMgr      This shell script takes care of starting and stopping a Kaltura Batch Service
 #
 # chkconfig: 2345 13 87
 # description: Kaltura Batch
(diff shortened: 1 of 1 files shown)


tan-tan-kanarek:
@jessp01, please review.

jessp01:
Hi David,

- usage should not exit with 0, please exit with 1
- if [ $# != 1 ]; then 
should be:
- if [ $# -ne 1 ]; then
since the value is numeric

- please use variables to hold the color codes, like in here: https://github.com/kaltura/platform-install-packages/blob/Jupiter-10.20.0/RPM/scripts/postinst/colors.sh, makes things more readable and can also be used for coloring other strings if needed in the future

- in echo_status(), echo -n "$1" can be done before the if

- where ever if [ -r $FILENAME ]; then, is used, better use -r instead

-     55         kill -0 $KP_PARENT
    56         if [ $? -eq 0 ]; then
    57             echo_status "Service Batch Manager already running [$KP_PARENT]" 0
    58         return 0
better redirect output of:
kill -0 $KP_PARENT 
to:
/dev/null
or you'll get an ugly output if it does not exist.
also, seem to be missing a 'fi' after the return 0 in line 58

- KP=`ps axf | awk '!/\\_ / {b=0} /php [K]GenericBatchMgr.class.php/ {b=1} b{print $1}'|xargs` 
repeats in start(), stop() and show_status(), better make it a get_pid() func, IMHO.

- in stop(), $1 might be empty as it is only passed to stop() in the event the force-stop arg.

- in stop() kill output should be redirected to /dev/null since $KP_PARENT might prove to be an invalid PID.

Thanks,

DBezemer:
Thanks @jessp01 for the finishing advice on this, changed and updated the file, will re-pull in a second

PR #3183, opened 2015-09-10
Update Batch Manager init script

DBezemer:
Refactor batch init script to correctly cover start/restart/stop commands with correct status detection and pid/lockfile handling.
Updated with advice of @jessp01


jessp01:
Looks good, David, thanks:)
@tan-tan-kanarek - please merge.

PR #3207, opened 2015-09-17
Jupiter 10.20.0 sup5976

Galmoss:


api_v3/lib/types/KalturaCategoryUser.php
```diff
@@ -158,6 +158,8 @@ public function validateForInsert($propertiesToSkip = array())
 		if ($category->getInheritanceType () == InheritanceType::INHERIT)
 			throw new KalturaAPIException ( KalturaErrors::CATEGORY_INHERIT_MEMBERS, $this->categoryId );
 		
+		$this->validatePropertyMinLength('userId',1);
+
```
MosheMaorKaltura: Need to validate that user ID is not Zero.

PR #3233, opened 2015-09-24
PLAT-3828 - Kaltura Transcript Asset Filter

rkreich:


plugins/content/transcript/lib/api/filters/KalturaTranscriptAssetFilter.php
```diff
@@ -0,0 +1,30 @@
...
+ * @subpackage api.filters
+ */
+class KalturaTranscriptAssetFilter extends KalturaTranscriptAssetBaseFilter
+{	
+	/* (non-PHPdoc)
+	 * @see KalturaAssetFilter::getTypeListResponse()
+	 */
+	public function getTypeListResponse(KalturaFilterPager $pager, KalturaDetachedResponseProfile $responseProfile = null, array $types = null)
+	{
+		$types = KalturaPluginManager::getExtendedTypes(assetPeer::OM_CLASS, TranscriptPlugin::getAssetTypeCoreValue(TranscriptAssetType::TRANSCRIPT));
```
rkreich: @tan-tan-kanarek I removed the `$typeIn` property as it seems to not being used, the type is forced from `$types` and we must override it here, otherwise we would get both TRANSCRIPT & ATTACHMENT in the query.

plugins/content/transcript/lib/api/filters/KalturaTranscriptAssetFilter.php
```diff
@@ -0,0 +1,30 @@
...
+		$response->totalCount = $totalCount;
+		return $response;
+	}
+
+	/* (non-PHPdoc)
+	 * @see KalturaAssetFilter::getListResponse()
+	 */
+	public function getListResponse(KalturaFilterPager $pager, KalturaDetachedResponseProfile $responseProfile = null)
+	{
+		$types = KalturaPluginManager::getExtendedTypes(assetPeer::OM_CLASS, TranscriptPlugin::getAssetTypeCoreValue(TranscriptAssetType::TRANSCRIPT));
```
rkreich: @tan-tan-kanarek Can `getListResponse` be called somehow?

tan-tan-kanarek: It could be called if this filter used in a related (nested) response-profile.

rkreich: OK, verified it.

PR #3241, opened 2015-09-29
New template file for percona DB

kobimichaeli:
PLAT-3898 #time 5m

### DIFF
--- a/configurations/monit/monit.d/percona.template.rc
+++ b/configurations/monit/monit.d/percona.template.rc
@@ -1,0 +1,16 @@
+        check process mysql
+           with pidfile "/var/lib/mysql/`hostname`.pid"
+           start program = "/sbin/service mysql start" with timeout 60 seconds
+           stop program = "/sbin/service mysql stop"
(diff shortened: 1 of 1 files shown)


kobimichaeli:
@jessp01 , please review

tan-tan-kanarek:
Why won't you name the file db.template.rc, it should be monitored regardless Percona is installed or used.

kobimichaeli:
If I understand you correctly, you are suggesting a unified / generic file for the DB monitoring.
At the moment, there’s a file for Mysql, Mariadb and now Percona.
The reason for the separate files are the small differences between them, for instance, mysql daemon is called ‘mysqld’, percona uses ‘mysql’.
PID file is also different….

Unless we want to create a logic which will be able to render a generic template file according to what’s installed, this would be the way to gap the differences.
AFAIK there aren’t more DB types which are planned to be introduced as default options in the near future, so I don’t know if such logic is worth the effort.
I leave it for your consideration, but regardless, if such a change is wished, it will have to be in the next development cycle due to the short time table at hand…

Thanks,
Kobi M.

PR #3269, opened 2015-10-01
fix typo in variable name

olampert:


alpha/lib/model/MediaServer.php
```diff
@@ -117,9 +117,14 @@ public function getManifestUrl($protocol = 'http', $partnerMediaServerConfigurat
 		    
 		    	if(isset($mediaServer[$portField]))
 		     		$port = $mediaServer[$portField];
+				if(isset($mediaServer[$portFieldByFormat]))
+		     		$port = $mediaServer[$portFieldByFormat];
+				
 		    
 		    	if(isset($mediaServer['domain']))
```
eladBenedict: As discussed, this is problematic since a "port" configuration of a lower level will override the "port-format" definition of a higher level, which isn't desirable. I think the logic we want here is - 
1. If "port-format" is defined at any level, we should use that value (with possible overrides of that parameter (explicitly) at lower levels), disregarding "port" altogether
2. If port-format is not defined at any level, its value should be the same as port's (and its overrides)

alpha/lib/model/MediaServer.php
```diff
@@ -79,6 +79,8 @@ public function getManifestUrl($protocol = 'http', $partnerMediaServerConfigurat
 		$domainFieldByFormat = 'domain';
 		$portFieldByFormat = 'port';
 		$appPrefix = '';
+		$isPortByFormat = false;
```
tan-tan-kanarek: I don't understand why you need these flags.

PR #3273, opened 2015-10-05
Jupiter 10.19.0 plat 3418

hilak:


alpha/apps/kaltura/lib/kMrssManager.php
```diff
@@ -460,6 +462,17 @@ public static function getEntryMrssXml(entry $entry, SimpleXMLElement $mrss = nu
 		}
 			
 		$categories = explode(',', $entry->getCategories());
+		if (count($features) && in_array (ObjectFeatureType::CATEGORY_ENTRIES, $features))
+		{
+			$categoryEntries = categoryEntryPeer::retrieveActiveByEntryId($entry->getId());
+			$categories = array ();
+			foreach ($categoryEntries as $categoryEntry)
+			{
+				$category = categoryPeer::retrieveByPK($categoryEntry->getCategoryId());
```
tan-tan-kanarek: This is very inefficient, please collect all ids and query them all together using retrieveByPKs.
Also, please disable all entitlement filters to make sure you actually get all the ids.

plugins/content_distribution/admin/forms/ConfigurableProfileConfiguration.php
```diff
@@ -5,6 +5,23 @@
  */
 abstract class Form_ConfigurableProfileConfiguration extends Form_ProviderProfileConfiguration
 {
+	public function init ()
+	{
+		parent::init();
+		
+		$this->addElement('checkbox', 'exclude_legacy_categories', array(
+			'label' => 'Exclude legacy categories',
```
tan-tan-kanarek: I'm sure you could find a text that actually explains the feature, e.g. "Enable categories with privacy-context".

api_v3/services/MediaService.php
```diff
@@ -702,10 +702,11 @@ function getAction($entryId, $version = -1)
      * @action getMrss
      * @param string $entryId Entry id
      * @param KalturaExtendingItemMrssParameterArray $extendingItemsArray
+     * @param string $features
      * @return string
      * @throws KalturaErrors::ENTRY_ID_NOT_FOUND
      */
-    function getMrssAction($entryId, KalturaExtendingItemMrssParameterArray $extendingItemsArray = null)
+    function getMrssAction($entryId, KalturaExtendingItemMrssParameterArray $extendingItemsArray = null, $features = null)
```
tan-tan-kanarek: I think the features are internal and used only when the MRSS requested internally.
In what use case do we need to expose it in the API? If it's used only by the admin-console XSLT tester, please add the feature in the action implementation when admin-console master partner is used.

plugins/content_distribution/admin/forms/ConfigurableProfileConfiguration.php
hilak: It's not really that either, sine entries can be associated with categories without PC via the categoryEntry objects as well.

PR #3283, opened 2015-10-06
Remove the explicit call the the kPdfRenderer::output() method

MosheMaorKaltura:
Jupiter-10.21.0-update-krenderer-output#time 0.5d

### DIFF
--- a/alpha/apps/kaltura/lib/renderers/kRendererPdfFile.php
+++ b/alpha/apps/kaltura/lib/renderers/kRendererPdfFile.php
@@ -14,7 +14,6 @@ class kRendererPdfFile implements kRendererBase
     private $pdfObj;
 
     public function __construct(PdfGenerator $obj)
-    //public function __construct($obj)
     {
         $this->pdfObj = $obj;
     }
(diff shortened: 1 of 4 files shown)


alpha/apps/kaltura/lib/renderers/kRendererPdfFile.php
```diff
@@ -26,9 +25,7 @@ public function validate()
 
     public function output()
     {
-        KalturaLog::debug("PDF::: kQuizPdf output");
         echo $this->pdfObj->Submit();
```
tan-tan-kanarek: Please use submit options to send headers either for stdout or for download.

infra/general/PdfGenerator.php
```diff
@@ -233,14 +226,11 @@ private function addText($text, PdfStyle $style)
 
     public function Submit()
     {
-        KalturaLog::debug("PDF::: PdfGenerator Submit()");
         if (strlen($this->signature) > 0)
         {
             $this->addSignature();
         }
-
         return ($this->Output('outQuizToBrowser.pdf','D'));
```
tan-tan-kanarek: Please add the quiz id to the file name.

alpha/apps/kaltura/lib/renderers/kRendererPdfFile.php
tan-tan-kanarek: Please ignore.

tan-tan-kanarek:
I think that the entire kQuizPdf class is redundant and could be implemented as a single method in the service.

PR #3304, opened 2015-10-12
remove redundant messages. Handle servePdf for non entry-id and for n…

TaliHa:
…on quiz entries


MosheMaorKaltura:
Approved.

TaliHa:
Prevent the servePdf action when the entry id is not a legal entry, or it is not a quiz.
Redundant debug messages were removed

PR #3312, opened 2015-10-13
changes after propel-gen

noam-arad:


tan-tan-kanarek:
@noam-arad,
Please revert all propel objects except the ones you need for your feature, the risk is too big.

PR #3329, opened 2015-10-16
added unique key to unsubscribe email feature

doubleshot:


tan-tan-kanarek:
I think that kaltura_email_hash should be moved from base.ini to local.template.ini.

doubleshot:
@tan-tan-kanarek  You're right, forgot that part. Let me know what else you think. Thanks

tan-tan-kanarek:
Everything else looks fine.

PR #3339, opened 2015-10-20
SUP-6102 Default Metadata on web upload

EladCohenKaltura:
SUP-6102 Default Metadata on web upload
Default Metadata on web upload

### DIFF
--- a/api_v3/services/MediaService.php
+++ b/api_v3/services/MediaService.php
@@ -332,8 +332,8 @@ class MediaService extends KalturaEntryService
 	 */
 	function addFromSearchResultAction(KalturaMediaEntry $mediaEntry = null, KalturaSearchResult $searchResult = null)
 	{
-    	if($mediaEntry->conversionQuality && !$mediaEntry->conversionProfileId)
-    		$mediaEntry->conversionProfileId = $mediaEntry->conversionQuality;
+		if($mediaEntry->conversionQuality && !$mediaEntry->conversionProfileId)
+			$mediaEntry->conversionProfileId = $mediaEntry->conversionQuality;
 
 		if ($mediaEntry === null)
 			$mediaEntry = new KalturaMediaEntry();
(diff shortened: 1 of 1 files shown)


MosheMaorKaltura:
Approved

api_v3/services/MediaService.php
```diff
@@ -1090,10 +1090,19 @@ protected function prepareEntryForInsert(KalturaBaseEntry $entry, entry $dbEntry
...
-		if (!is_null($conversionQuality))
+		if (!is_null($conversionQuality)) {
 			$entry->conversionQuality = $conversionQuality;
+			if (!$entry->conversionProfileId) {
+				$entry->conversionProfileId = $entry->conversionQuality;
+			}
+		}
+
+		if ($dbEntry == null){
+			$dbEntry = $this->duplicateTemplateEntry($entry->conversionProfileId);
```
tan-tan-kanarek: I think it makes more sense to call duplicateTemplateEntry in KalturaEntryService::prepareEntryForInsert.

tan-tan-kanarek: It means that you'll have to remove the call to duplicateTemplateEntry  from KalturaEntryService::add.

EladCohenKaltura: Yet I am not so sure about it - since the conversion profile is something that is relevant to Media and not all services - seems a bit unnatural to use the conversion profile ID (for example the document service)

tan-tan-kanarek: The name conversion-profile is confusing, it should be named ingestion-profile, it contains template-entry-id, storage-profile-id, and specifically to document-entries it's very relevant as it's used just like in media-entries to define the transcoded flavors.

EladCohenKaltura: Created https://kaltura.atlassian.net/browse/PLAT-4073 to give this issue the attention needed

PR #3361, opened 2015-10-22
Kajam 11.0.0 plat 4028

hilak:
@assafb  Please review. If code review is ok, this pull request will be closed and a new one opened for Kajam-12.0.0

### DIFF
--- a/api_v3/lib/syndication/KalturaFeedRenderer.php
+++ b/api_v3/lib/syndication/KalturaFeedRenderer.php
@@ -161,8 +161,14 @@ class KalturaFeedRenderer extends SyndicationFeedRenderer{
 		$mrssParams->setStorageId($syndicationFeed->getStorageId());
 		$mrssParams->setServePlayManifest($syndicationFeed->getServePlayManifest());
 		$mrssParams->setPlayManifestClientTag('feed:' . $syndicationFeed->getId());
-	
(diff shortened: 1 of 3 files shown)


api_v3/lib/syndication/KalturaFeedRenderer.php
```diff
@@ -161,8 +161,15 @@ private function getMrssEntryXml(entry $entry, syndicationFeed $syndicationFeed
...
 		$mrssParams->setServePlayManifest($syndicationFeed->getServePlayManifest());
 		$mrssParams->setPlayManifestClientTag('feed:' . $syndicationFeed->getId());
-	
-		$mrss = kMrssManager::getEntryMrssXml($entry, null, $mrssParams);
+
+		$features = null;
+		if ($syndicationFeed->getUseCategoryEntries())
+		{
+			KalturaLog::info("Getting entry's associated categories from the category_entry table");
+			$features = array ("3");
```
tan-tan-kanarek: Please use constant.

tan-tan-kanarek:
I think that from now on all feeds should use Category-Entries and the old deprecated categories ids should be used only in old feeds.
That's why I also wouldn't expose that as an API functionality but as a core attribute only, that will be changed only upon request using a propel script.
The question is how would that affect the performance.
+@erankor

PR #3364, opened 2015-10-25
PLAT-3902

EladCohenKaltura:
Allow "sview" privileged ks to play files


EladCohenKaltura:
Will not be pushed to Kajam.11.0.0 but to Kajam.11.1.0

PR #3375, opened 2015-10-26
Jupiter 10.16.0 plat 3270

EladCohenKaltura:


plugins/content_distribution/lib/api/KalturaCaptionDistributionInfo.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
+ * @package plugins.contentDistribution
+ * @subpackage api.objects
+ *
+ */
+class KalturaCaptionDistributionInfo extends KalturaObject{
```
tan-tan-kanarek: Instead of repeating fields from caption asset, use the asset id or extend the KalturaCaptionAsset.

tan-tan-kanarek: In second thought, the content-distribution plugin today is not pending on caption-assets and we don't want it to.
You can pend facebook plugin on captions if you think it's needed, but adding captions already solved by other distribution providers such as you-tube without creating dependency.

plugins/content_distribution/providers/facebook/lib/api/KalturaFacebookDistributionField.php
```diff
@@ -0,0 +1,9 @@
+<?php
+/**
+ * @package plugins.facebookDistribution
+ * @subpackage api.enum
+ */
+class KalturaFacebookDistributionField extends KalturaStringEnum implements FacebookDistributionField
```
tan-tan-kanarek: Where is that class used?

plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
```diff
@@ -0,0 +1,150 @@
...
+	
+	/* (non-PHPdoc)
+	 * @see DistributionEngine::configure()
+	 */
+	public function configure()
+	{
+		parent::configure();
+		if (isset(KBatchBase::$taskConfig->params->facebook))
+		{
+			if (isset(KBatchBase::$taskConfig->params->facebook->appId))
```
tan-tan-kanarek: I suggest that you'll add the app id and secret to the job provider data and load it from kConf instead of having two places that should be configured.

plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
```diff
@@ -0,0 +1,150 @@
...
+		$this->validate($data);	
+		if($data->entryDistribution->remoteId)
+		{
+			$data->remoteId = $data->entryDistribution->remoteId;
+		}
+		else 
+		{
+			$this->doSubmit($data, $data->distributionProfile);
+		}
+		return true;
```
tan-tan-kanarek: If submit always returns true then there is no need to implement IDistributionEngineCloseSubmit.

plugins/content_distribution/providers/facebook/lib/FacebookDistributionProvider.php
```diff
@@ -0,0 +1,127 @@
...
+	{
+		return false;
+	}
+	
+	/* (non-PHPdoc)
+	 * @see IDistributionProvider::isAvailabilityUpdateEnabled()
+	 */
+	public function isAvailabilityUpdateEnabled()
+	{
+		return true;
```
tan-tan-kanarek: Availability update is not supported by the engine, in order to support availability update you must implement IDistributionEngineEnable and IDistributionEngineDisable.

PR #3377, opened 2015-10-26
Add new action for Quiz service 'getUrl'. Update action 'servePdf' in…

TaliHa:
… the following way: 1. Action name is replaced 'servePdf' ==> 'serve' 2. The API receives a parameter which indicates the output file type. Currently only PDF files are supported

Kajam-11.1.0-PLAT-3975 #time 3 days


plugins/cue_points/quiz/services/QuizService.php
```diff
@@ -147,8 +151,49 @@ public function servePdfAction($entryId)
...
-		$kp->createQuestionPdf();
-		return $kp->submitDocument();
+		if ($quizFileType == KalturaQuizFileType::PDF) {
+			$kp = new kQuizPdf($entryId);
+			$kp->createQuestionPdf();
+			return $kp->submitDocument();
+		}
+		else
+		{
+			throw new KalturaAPIException(KalturaQuizErrors::NO_SUCH_FILE_TYPE);
```
tan-tan-kanarek: No need to verify that, it's already verified by the API.

PR #3385, opened 2015-10-27
Add 'getUrl' action and update the 'servePdf' action in the following…

TaliHa:
… way: 1. Rename action 'serverPdf'==> 'serve' 2. Add a parameter to the actionthat defines the file type that will be generated. Currently only PDF files are supported. In this commit the release notes are updated

Kajam-11.1.0-PLAT-3975-fix # 3 days


MosheMaorKaltura:
Please specify that need to run clear cache script and resolve the defect

PR #3395, opened 2015-10-29
Kajam 11.1.0 plat 3783

MosheMaorKaltura:


alpha/apps/kaltura/lib/kEntitlementUtils.php
```diff
@@ -521,4 +521,22 @@ protected static function initCategoryModeration (ks $ks = null)
...
+	/**
+	 * @param entry $dbEntry
+	 * @return bool if current user is admin / entry's owner / co-editor
+	 */
+	public static function isEntitledForEditEntry( entry $dbEntry )
+	{
+		if ( kCurrentContext::$is_admin_session || kCurrentContext::getCurrentKsKuserId() == $dbEntry->getKuserId())
+			return true;
+
+		$entitledKusers = baseObjectUtils::getObjectIdsAsArray($dbEntry->getEntitledKusersEdit());
```
tan-tan-kanarek: Simple explode should be good enough, but entry::getEntitledKusersEdit do implode before returning, so it doesn't make send to implode and explode, so I would create a new method called entry::getEntitledKusersEditArray.

tan-tan-kanarek: Actually there is entry::isEntitledKuserEdit that you can use, but that method also implode and explode and requires fixing.

plugins/cue_points/quiz/QuizPlugin.php
```diff
@@ -495,7 +476,7 @@ protected function getQuestionPercentageTableReport($objectIds, $orderBy)
 	
 	protected function getQuizQuestionPercentageTableReport($objectIds, $orderBy)
 	{
-		$questionIds = QuizPlugin::getObjectIdsAsArray($objectIds);
+		$questionIds = baseObjectUtils::getObjectIdsAsArray($objectIds);
```
tan-tan-kanarek: I think that explode is good enough, we don't expect spaces from the API.

tan-tan-kanarek:
@MichalRadwantzor, please revert this pull-request.
Please return the validateUser and change IOwnable to support co-owner.

+@erankor

tan-tan-kanarek:
Eran mentioned that maybe delete actions shouldn't be enabled to co-editors, which means that we will have to support a change in the annotation, but product team should approve this logic.

yossipapi:
Reminder that cue point's are not fully implementing the IOwnable so we will probably need a different handling mechanism specific for cue points.

PR #3405, opened 2015-10-29
PLAT 3902 allow isShceduledNow when sview privilege

EladCohenKaltura:
PLAT 3902 allow isShceduledNow when sview privilege


EladCohenKaltura:
Validated on KMS - video played even when out of scheduling

PR #3415, opened 2015-11-02
PLAT-4137 fix bad logic

assafb:


tan-tan-kanarek:
Assaf, I'm sure we can find a more elegant way to prevent this issue in the first place instead of fixing it later using regex.

PR #3418, opened 2015-11-02
Add property showGradeAfterSubmission. and the submitQuizAction remov…

TaliHa:
…es the score in case the user is not an admin user and submitQuizAction

is set to be false.
PLAT-4050 #time 1.5 days

### DIFF
--- a/plugins/cue_points/quiz/services/QuizUserEntryService.php
+++ b/plugins/cue_points/quiz/services/QuizUserEntryService.php
@@ -19,7 +19,6 @@ class QuizUserEntryService extends KalturaBaseService{
 	 */
 	public function submitQuizAction($id)
 	{
-		KalturaLog::debug("Submit Quiz for QuizUserEntryId [" .$id."]");
 		$dbUserEntry = UserEntryPeer::retrieveByPK($id);
 		if (!$dbUserEntry)
 		{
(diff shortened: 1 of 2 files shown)


plugins/cue_points/quiz/services/QuizUserEntryService.php
```diff
@@ -19,25 +19,67 @@ class QuizUserEntryService extends KalturaBaseService{
 	 */
 	public function submitQuizAction($id)
 	{
+		KalturaLog::debug("Submit Quiz for QuizUserEntryId [" .$id."]");
```
tan-tan-kanarek: Please remove unneeded logs.

plugins/cue_points/quiz/services/QuizUserEntryService.php
```diff
@@ -19,25 +19,67 @@ class QuizUserEntryService extends KalturaBaseService{
...
 		$dbUserEntry->save();
 
 		$userEntry = new KalturaQuizUserEntry();
 		$userEntry->fromObject($dbUserEntry, $this->getResponseProfile());
+
+		$entryId = $dbUserEntry->getEntryId();
+		$entry = entryPeer::retrieveByPK($entryId);
+		if(!$entry)
+		{
+			throw new KalturaAPIException(KalturaErrors::INVALID_OBJECT_ID, $entryId);
```
tan-tan-kanarek: I wouldn't throw exception after the actual save already done, the use may think that the quiz was not submitted.

PR #3440, opened 2015-11-04
Kajam 11.1.0 plat 4117

MichalRadwantzor:


MichalRadwantzor:
@tan-tan-kanarek @yossipapi please have a look. Thanks!

tan-tan-kanarek:
Now entry could be deleted by co-editor, we can't deploy it to production before product team approval and later break it.

MosheMaorKaltura:
@tan-tan-kanarek The delete issue is known, we will discuss it with Renan & Uri prior to merging this code.

MichalRadwantzor:
@MosheMaorKaltura @tan-tan-kanarek please review again. Thanks

api_v3/lib/KalturaDispatcher.php
```diff
@@ -209,7 +210,11 @@ protected function validateUser($objectClass, $objectId, $privilege = null)
 			
 		}
 
-		if (strtolower($dbObject->getPuserId()) != strtolower(kCurrentContext::$ks_uid)) 
+		if (strtolower($dbObject->getPuserId()) != strtolower(kCurrentContext::$ks_uid)
+			&&
+			(!$dbObject->isEntitledKuserEdit(kCurrentContext::getCurrentKsKuserId())
+				||
+				($condition && $condition == KalturaDocCommentParser::OWNER_ONLY_CONDITION)))
```
tan-tan-kanarek: I think you better place the constant in this class (KalturaDispatcher) and not in the KalturaDocCommentParser.
I also would call the last argument $options, first because we might have options that are not conditions in the future, secondly, because we may have more than one option comma separated.

PR #3462, opened 2015-11-09
Fixing password with apostrophe when using sshpass

Galmoss:
SUP-6344


yossipapi:
@Galmoss why won't you use php's addslashes function this will do the escaping for you.

Galmoss:
@yossipapi cause it doesn't handle strong quoting as we need in this case.

PR #3468, opened 2015-11-10
use ID3 tags in VOD to calculate the time of the cue points

asafrob:
with this change need to make sure ffprobe is configured on the machine


api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -205,21 +209,7 @@ function registerMediaServerAction($entryId, $hostname, $mediaServerIndex, $appl
 		if (!$dbEntry || !($dbEntry instanceof LiveEntry))
 			throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $entryId);
 		
-		try
```
asafrob: this code should not be deleted....

PR #3495, opened 2015-11-16
Correctly calculate last duration - from recorded entry

danoss:


api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -72,9 +72,12 @@ function appendRecordingAction($entryId, $assetId, $mediaServerIndex, KalturaDat
...
 			throw new KalturaAPIException(KalturaErrors::ASSET_ID_NOT_FOUND, $assetId);
 			
-		$lastDuration = $dbEntry->getLengthInMsecs();
-		if(!$lastDuration)
-			$lastDuration = 0;
+		$lastDuration = 0;
+		if ($dbEntry->getRecordedEntryId())
+		{
+			$recordedEntry = entryPeer::retrieveByPK($dbEntry->getRecordedEntryId());
+			$lastDuration = $recordedEntry->getLengthInMsecs();
```
tan-tan-kanarek: Please verify that $recordedEntry is not null.
Also, please check if the recorded entry is in replacement, in such case you'll have to use the replacing entry duration.
Also, please remove entirely the update of the duration on the live entry as it's not needed anymore.

PR #3499, opened 2015-11-16
Changed the admin console from 3 to unlimited

kobimichaeli:
PLAT-1129 # time 10m

### DIFF
--- a/alpha/lib/model/Partner.php
+++ b/alpha/lib/model/Partner.php
@@ -982,7 +982,7 @@ class Partner extends BasePartner
 	public function setLiveStreamPlaybackUrlConfigurations($v)		{$this->putInCustomData('live_stream_playback_url_configurations', $v);}
     
 	public function getLoginUsersQuota()				{return $this->getFromCustomData('login_users_quota', null, 0);}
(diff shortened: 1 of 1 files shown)


tan-tan-kanarek:
This change on SaaS may remove the limitation to most partners.
Please make it configurable instead.

kobimichaeli:
This change as I've written it, will never reach SaaS, since it's intended only for OnPrem.
If you want, an additional bug can be opened for SaaS, but I doubt if I'll be the one who handles it (at least in the near future)

tan-tan-kanarek:
You wrote it in an old SaaS branch, and that is why it won't reach SaaS.
But for the same reason it also won't reach next on-prem and CE versions.



נשלח ממכשיר הסמסונג שלי


-------- הודעה מקורית --------
מאת: kobimichaeli <notifications@github.com>
תאריך: 16/11/2015 16:36 (GMT+02:00)
אל: kaltura/server <server@noreply.github.com>
עותק: Jonathan Kanarek <Jonathan.Kanarek@kaltura.com>
נושא: Re: [server] Changed the admin console from 3 to unlimited (#3499)


This change as I've written it, will never reach SaaS, since it's intended only for OnPrem.
If you want, an additional bug can be opened for SaaS, but I doubt if I'll be the one who handles it (at least in the near future)

-

PR #3511, opened 2015-11-18
Return latest evnen if version is already the latest

tan-tan-kanarek:


plugins/capture_space/lib/kCaptureSpaceVersionManager.php
```diff
@@ -18,26 +18,23 @@ class kCaptureSpaceVersionManager{
...
 			$filename = __DIR__ . '/../config/' . self::INI_FILE_NAME;
 			self::$config = parse_ini_file($filename, true);
 		}
-		return self::$config;
-	}
-	
-	private static function getConfig($os, $osTypes, $version = null){
-		$config = self::initConfig();
+		
+		self::$config = self::initConfig();
```
gonenradai: there's no such function initConfig on this class.

plugins/capture_space/lib/kCaptureSpaceVersionManager.php
```diff
@@ -18,26 +18,23 @@ class kCaptureSpaceVersionManager{
...
 				return null;
 			}
-			$sections = array($version => $config[$version]);
+			$sections = array($version => self::$config[$version]);
 		}
 		else {
-			uksort($config, 'version_compare');
-			$sections = array_reverse($config, true);
+			uksort(self::$config, 'version_compare');
+			$sections = array_reverse(self::$config, true);
```
gonenradai: This basically means INI must be written in ascending order of versions (lowest version - first in INI).
I'm OK with that, although it is error prone.
wouldn't be easier to maintain a [latest] section?

tan-tan-kanarek: The order doesn't matter, I call uksort.

gonenradai: right... missed that one.

PR #3530, opened 2015-11-23
Wrapped remote_addr_header_salt val in ' '

kobimichaeli:
@jessp01, I've had to fix this on the old version since it fails the ssl installation.


jessp01:
Personally, I think a better fix is to escape it in the kaltura-base-config.sh script instead which is what I do.

PR #3534, opened 2015-11-24
Kajam 11.3.0 plat 3963

hilak:


plugins/content_distribution/providers/comcast_mrss/lib/ComcastMrssFeed.php
```diff
@@ -300,6 +313,22 @@ public function setThumbAssets(DOMElement $item, array $thumbAssets)
...
 	
+	public function setCaptionAssets (DOMElement $item, array $captionAssets)
+	{
+		foreach ($captionAssets as $captionAsset)
+		{
+			/* @var $captionAsset captionAsset */
+			$content = $this->caption->cloneNode(true);
+			$mediaGroup = $this->xpath->query('media:group', $item)->item(0);
+			$mediaGroup->appendChild($content);
+			$url = $captionAsset->getDownloadUrl(true);
```
hilak: @tan-tan-kanarek  Is there a way to get the captions download URL without the KS? Is there a possibility in your opinion of the KS being cached on the feed beyond the KS expiry date?

tan-tan-kanarek: We can use tokenized URLs just like our serveFlavor actions.

hilak: After checking, I see that the KS expiry is exactly the same as the feed cache expiry - 24 h.

tan-tan-kanarek:
Please add dependency on captions plugin in the comcast plugin.

PR #3537, opened 2015-11-24
Ensuring that a per session recorded entry is created exactly once pe…

eladBenedict:
…r session

### DIFF
--- a/api_v3/lib/KalturaLiveEntryService.php
+++ b/api_v3/lib/KalturaLiveEntryService.php
@@ -255,26 +255,29 @@ class KalturaLiveEntryService extends KalturaEntryService
 				if(!$dbEntry->getRecordedEntryId())
 				{
 					$createRecordedEntry = true;
-					KalturaLog::info("Creating new recoreded entry for $entryId ");
+					KalturaLog::info("Creating a new recorded entry for $entryId ");
 				}
 				else {
 					$dbRecordedEntry = entryPeer::retrieveByPK($dbEntry->getRecordedEntryId());
 					if (!$dbRecordedEntry) {
-						throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $dbRecordedEntry);
+						$createRecordedEntry = true;
 					}
-					$recordedEntryCreationTime = strtotime($dbRecordedEntry->getCreatedAt());
+					else{
+						$recordedEntryCreationTime = $dbRecordedEntry->getCreatedAt(null);
 
-					$isNewSession = $dbEntry->getLastBroadcastEndTime() + kConf::get('live_session_reconnect_timeout', 'local', 180) < $dbEntry->getCurrentBroadcastStartTime();
-					$recordedEntryNotYetCreatedForCurrentSession = $recordedEntryCreationTime < $dbEntry->getCurrentBroadcastStartTime();
+						$isNewSession = $dbEntry->getLastBroadcastEndTime() + kConf::get('live_session_reconnect_timeout', 'local', 180) < $dbEntry->getCurrentBroadcastStartTime();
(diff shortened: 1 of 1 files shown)


api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -246,26 +246,40 @@ function registerMediaServerAction($entryId, $hostname, $mediaServerIndex, $appl
...
 					$createRecordedEntry = true;
+					KalturaLog::info("Creating new recoreded entry for $entryId ");
 				}
-				elseif($dbEntry->getRecordStatus() == RecordStatus::PER_SESSION && ($dbEntry->getLastBroadcastEndTime() + kConf::get('live_session_reconnect_timeout', 'local', 180) < time()))
-				{
-					$createRecordedEntry = true;
+				else {
+					$dbRecordedEntry = entryPeer::retrieveByPK($dbEntry->getRecordedEntryId());
+					if (!$dbRecordedEntry) {
+						throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $dbRecordedEntry);
```
tan-tan-kanarek: It doesn't make sense to throw exception here, if the recorded entry doesn't exist simply create a new one.

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -246,26 +246,40 @@ function registerMediaServerAction($entryId, $hostname, $mediaServerIndex, $appl
...
+					}
+					$recordedEntryCreationTime = strtotime($dbRecordedEntry->getCreatedAt());
```
tan-tan-kanarek: Instead, use:
$recordedEntryCreationTime = $dbRecordedEntry->getCreatedAt(null);

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -246,26 +246,40 @@ function registerMediaServerAction($entryId, $hostname, $mediaServerIndex, $appl
...
+
+					$isNewSession = $dbEntry->getLastBroadcastEndTime() + kConf::get('live_session_reconnect_timeout', 'local', 180) < $dbEntry->getCurrentBroadcastStartTime();
+					$recordedEntryNotYetCreatedForCurrentSession = $recordedEntryCreationTime < $dbEntry->getCurrentBroadcastStartTime();
+
+					if ($dbEntry->getRecordStatus() == RecordStatus::PER_SESSION) {
+						if ($isNewSession && $recordedEntryNotYetCreatedForCurrentSession)
```
tan-tan-kanarek: Shouldn't it be OR instead of AND?

api_v3/lib/KalturaLiveEntryService.php
ronyad: Since "$dbEntry->getRecordedEntryId()" exists (i.e. not null), the assumption is that the entry exists.
However, since the API server is a distributed and complicated system, we want to guarantee  that  the API call would not return null.
The idea is similar to the code written in line  230 ( i.e. "if (!$dbEntry || !($dbEntry instanceof LiveEntry)" ), and in consultation with Elad Cohen.

api_v3/lib/KalturaLiveEntryService.php
ronyad: AND.
In order to create new recored entry, one require that
1. The session is new, namely, more than 3 minutes were passed from the last broadcast time.
2. Since, the current function (registerMediaServerAction) might called every one minutes, we also need to ensure that it the first time that we create new recorded entry.

api_v3/lib/KalturaLiveEntryService.php
ronyad: Got it. Thanks!

PR #3540, opened 2015-11-24
eCDN VOD support

yossipapi:


plugins/media/wowza/lib/model/WowzaMediaServerNode.php
```diff
@@ -51,7 +51,7 @@ public function getLiveWebServiceName()
 		return WowzaMediaServerNode::WEB_SERVICE_LIVE;
 	}
 	
-	public function getPlaybackHost($protocol = 'http', $format = null)
+	public function getPlaybackHost($protocol = 'http', $format = null, $deliveryType = null)
```
tan-tan-kanarek: I don't think you need both the format and the deliveryType, you can get one from the other.

alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
```diff
@@ -872,8 +872,9 @@ private function serveLiveEntry()
 			return null;
 		}
 		
-		$this->deliveryProfile->setDynamicAttributes($this->deliveryAttributes);	
-		return $this->deliveryProfile->serve($liveStreamConfig);
+		$this->deliveryProfile->setDynamicAttributes($this->deliveryAttributes);
+		$this->deliveryProfile->setLiveStreamConfig($liveStreamConfig);	
+		return $this->deliveryProfile->serve();
```
tan-tan-kanarek: Why won't we pass the $liveStreamConfig to the serve action?

yossipapi: There are only 3 places where the liveStreamConfig is used out of the 20+ delivery types we use and passing it to the serve will require me to pass it to the buildServeFlavors function as well so I think it is better to leave it as is.

plugins/media/wowza/lib/model/WowzaMediaServerNode.php
yossipapi: This is right for VOD but not right for live since for live we build the urls in the baseEntry get request. There we do not initialize a delivery profile before building the urls.

PR #3546, opened 2015-11-25
Add parameter "CaptureSpace - Full client" to client config

MosheMaorKaltura:
PLAT-4190#time 2h


gonenradai:
looking good to me.
I will be able to call, with admin KS:
permission.get(CAPTURESPACE_PLUGIN_PERMISSION) 
and then:
if status = 1 => partner has feature
if status != 1 => partner does not have feature.

PR #3552, opened 2015-11-26
Do not strict file_sync search for uiconf Data file

MosheMaorKaltura:
SUP-6297


MosheMaorKaltura:
Supporting the case of new players that don't need 'Conf File:' section.

PR #3553, opened 2015-11-26
Setting customData twice causes error saving when customData is null

yossipapi:


tan-tan-kanarek:
@yossipapi, if you remove the set from the builder, it will affect entry next time it will be generated and will cause damages as there are use cases that the custom-data changes during the pre-update hook.
We should remove the set from the pre-save and return it to run after the pre hooks, and we have to check is custom-data fields modified in order to decide if updated-at should be updated.

PR #3573, opened 2015-12-03
Csharp client chunk upload enhancements

gonenradai:
Adding support for MemoryStream + example for threaded chunk upload.

Threaded chunk upload requires a small change in the backend service to allow writing a chunk after the current-end-of-file, thus this PR should not be merged before this BE change is implemented and merged.

### DIFF
--- a/generator/sources/csharp/KalturaClientTester/KalturaUploadThread.cs
+++ b/generator/sources/csharp/KalturaClientTester/KalturaUploadThread.cs
@@ -1,0 +1,70 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+using System.Threading;
+
+namespace Kaltura
+{
+    class KalturaUploadThread
+    {
+        public string ks;
+        public FileStream file;
+        public string uploadTokenId;
+        public int chunkSize;
+
+
+        /*
+         * position => uploaded
+         */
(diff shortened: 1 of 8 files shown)


generator/sources/csharp/KalturaClientTester/KalturaUploadThread.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+using System.Threading;
+
+namespace Kaltura
+{
+    class KalturaUploadThread
```
tan-tan-kanarek: Please place this class as a facility under KalturaClient instead of under KalturaClientTester.

gonenradai: I don't think it fits as a facility class for a general purpose implementation.
specifically - because it uses Console.WriteLine and because it interacts with KalturaClientTester directly via
```
KalturaClientTester.workingThreads--;
```

This is more a sample implementation than a generic useful addition to the client lib itself.

gonenradai:
Apparently, there is a problem with chunk upload in this library that I haven't figured out yet (just found out today).
I don't know if it is in the library itself, or in the sample implementation - I will investigate and handle accordingly.

PR #3591, opened 2015-12-08
Update review comments

TaliHa:
Kajam-11.4.0-PLAT-4632 # 2 days


plugins/cue_points/quiz/lib/api/KalturaAnswerCuePoint.php
```diff
@@ -99,17 +100,17 @@ public function doFromObject($dbObject, KalturaDetachedResponseProfile $response
 			{
 				if (!$kQuiz->getShowCorrectAfterSubmission())
 				{
-					$this->isCorrect = KalturaNullableBoolean::NULL_VALUE;
+					$this->isCorrect = null;
```
MosheMaorKaltura: The type of this filed is KalturaNullableBoolean ,it would be a better practice to set a KalturaNullableBoolean::NULL_VALUE value OR to change its type to bool since it is a read only.
(This depends if the client is already using the field and the change will not break its behaviour).

PR #3593, opened 2015-12-09
Kajam 11.4.0 sup6332

MosheMaorKaltura:


alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -289,13 +289,10 @@ public static function handleConvertLiveSegmentFinished(BatchJob $dbBatchJob, kC
...
-			$replacingEntry = kFlowHelper::getReplacingEntry($recordedEntry, $asset);
-			if(is_null($replacingEntry)) {
-				KalturaLog::err('Failed to retrieve replacing entry');
-				return $dbBatchJob;
-			}
-			
+			$replacingEntry = self::getReplacingEntry($recordedEntry, $asset);
+			if(is_null($replacingEntry))
+				KalturaLog::err("Failed to get replacing entry");
+
```
noam-arad: If we continue we might have a fatal on line #304 since $replacingAsset will be NULL

PR #3610, opened 2015-12-15
#kms-10231 #comment keep timedThumbAssets on replace media

MichalRadwantzor:


alpha/lib/model/thumbAsset.php
```diff
@@ -42,4 +42,13 @@ public function getFinalDownloadUrlPathWithoutKs()
 						
 		return $finalPath;
 	}
+
+	public function keepOnEntryReplacement(entry $entry)
```
tan-tan-kanarek: Instead of passing the entry to this method you can simply use $entry = $this->getentry();

PR #3615, opened 2015-12-16
#SUP-6620 #comment save temp oldCustomDataValues so change event cons…

MichalRadwantzor:
…umers can read the values

### DIFF
--- a/alpha/lib/model/entry.php
+++ b/alpha/lib/model/entry.php
@@ -3372,7 +3372,8 @@ class entry extends Baseentry implements ISyncableFile, IIndexable, IOwnable, IR
 	 */
 	public function isCustomDataModified($name = null, $namespace = '')
 	{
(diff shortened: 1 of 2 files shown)


yossipapi:
@MichalRadwantzor you cannot edit Baseentry.php directly since it is an auto generated class which will result in you changes being overwritten next time some one will re-generate the base classes.

What you need to do is to change KalturaObjectBuilder  to include your changes and run propel-gen to re-generate the base classes.

After the re-generation commit only to the class you want to apply the changes on and if everything works properly after tests add them to all other base classes as well.

Also, isn't the tempOldCustomDataValues an exact copy of oldCustomDataValues which was already assigned to the modifiedColumns before throwing the event?

MichalRadwantzor:
@yossipapi Thanks for the explanation! I will change it. 
the idea of tempOldCustomDataValues is to temporarily save the old custom data values while saving, because "postSave" function resets the oldCustomDataValues before the changeEvent consumers checked for the changed values (it is similar to tempModifiedColumns).

PR #3618, opened 2015-12-20
When serving vod contnet filter for vod delivery profiles only

yossipapi:


yossipapi:
Already handled as part of another PR

PR #3622, opened 2015-12-21
Kajam 11.1.0 facebook connector

EladCohenKaltura:
Facebook distributor main functionality


plugins/content_distribution/providers/facebook/config/generator.ini
```diff
@@ -0,0 +1,3 @@
+
+[batchClient]
+additional = KalturaFacebookDistributionField
```
tan-tan-kanarek: I don't see where this class is used.

PR #3630, opened 2015-12-23
Kajam 11.5.0 plat 4795

MichalRadwantzor:
Add restoreDeletedEntry action

### DIFF
--- a/plugins/admin_console/services/EntryAdminService.php
+++ b/plugins/admin_console/services/EntryAdminService.php
@@ -132,7 +132,7 @@ class EntryAdminService extends KalturaBaseService
 		FileSyncPeer::setUseCriteriaFilter(true);
 
 		if (!$this->validateEntryForRestoreDelete($deletedEntry, $fileSyncs, $deletedAssets))
-			throw new KalturaAPIException(KalturaErrors::ENTRY_ASSETS_WRONG_STATUS_FOR_RESTORE, $entryId);
+			throw new KalturaAPIException(KalturaAdminConsoleErrors::ENTRY_ASSETS_WRONG_STATUS_FOR_RESTORE, $entryId);
 
 		$deletedEntry->setStatusReady();
 		$deletedEntry->setThumbnail($deletedEntry->getFromCustomData("deleted_original_thumb"), true);
--- a/plugins/admin_console/services/EntryAdminService.php
+++ b/plugins/admin_console/services/EntryAdminService.php
@@ -8,6 +8,18 @@
  */
 class EntryAdminService extends KalturaBaseService
 {
+	private static $fileSyncKeysToRestore = array (
+		entry::FILE_SYNC_ENTRY_SUB_TYPE_DATA,
+		entry::FILE_SYNC_ENTRY_SUB_TYPE_DATA_EDIT,
+		entry::FILE_SYNC_ENTRY_SUB_TYPE_THUMB,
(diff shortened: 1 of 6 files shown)


plugins/admin_console/services/EntryAdminService.php
```diff
@@ -81,4 +93,113 @@ function getTracksAction($entryId)
...
+
+		//restore assets
+		foreach($deletedAssets as $deletedAsset)
+		{
+			$deletedAsset->setStatus(asset::ASSET_STATUS_READY);
+			$deletedAsset->save();
+		}
+
+		//restore entry
+		$deletedEntry->setStatusReady();
```
tan-tan-kanarek: This method will reset the moderation status, please use setStatus instead.

plugins/admin_console/services/EntryAdminService.php
```diff
@@ -81,4 +93,113 @@ function getTracksAction($entryId)
...
+
+		$fileSyncs = array();
+		FileSyncPeer::setUseCriteriaFilter(false);
+		foreach ($fileSyncKeys as $fileSyncKey)
+		{
+			$fileSyncs = array_merge($fileSyncs, FileSyncPeer::retrieveAllByFileSyncKey($fileSyncKey));
+		}
+		FileSyncPeer::setUseCriteriaFilter(true);
+
+		if (!$this->validateEntryForRestoreDelete($deletedEntry, $fileSyncs, $deletedAssets))
```
tan-tan-kanarek: I think you better throw the exception inside validateEntryForRestoreDelete in case you would like to validate more conditions in the future.

plugins/admin_console/services/EntryAdminService.php
```diff
@@ -81,4 +93,113 @@ function getTracksAction($entryId)
...
+		$entry->fromObject($deletedEntry, $this->getResponseProfile());
+		return $entry;
+	}
+
+	protected function restoreFileSyncs( array $fileSyncs )
+	{
+		foreach ($fileSyncs as $fileSync)
+		{
+			$shouldUnDelete = false;
+			if (($fileSync->getFileType() == FileSync::FILE_SYNC_FILE_TYPE_FILE && file_exists($fileSync->getFullPath()))
```
tan-tan-kanarek: File-sync objects for other DCs will never return true for file_exists.
If you want to restore them too, you can call another API internally to the other DC.

plugins/admin_console/services/EntryAdminService.php
```diff
@@ -81,4 +93,113 @@ function getTracksAction($entryId)
...
+				|| ($fileSync->getFileType() == FileSync::FILE_SYNC_FILE_TYPE_URL)){
+				$shouldUnDelete = true;
+			}
+			else if ($fileSync->getFileType() == FileSync::FILE_SYNC_FILE_TYPE_LINK){
+				$linkedId = $fileSync->getLinkedId();
```
tan-tan-kanarek: If you restore a link you should increment the link count.

plugins/admin_console/services/EntryAdminService.php
MichalRadwantzor: The moderation status should be updated as well, otherwise it will remain in MODERATION_STATUS_DELETE

MichalRadwantzor:
will open to 11.6.0

PR #3658, opened 2016-01-04
PLAT-3962 - Support usage dashboard

rkreich:


alpha/lib/enums/PermissionName.php
```diff
@@ -236,6 +236,7 @@
 	
 	const FEATURE_SHOW_FLASH_STUDIO = 'FEATURE_SHOW_FLASH_STUDIO';
 	const FEATURE_SHOW_HTML_STUDIO = 'FEATURE_SHOW_HTML_STUDIO';
+	const FEATURE_ENABLE_USAGE_DASHBOARD = 'FEATURE_ENABLE_USAGE_DASHBOARD';
```
tan-tan-kanarek: If you don't refer to this permission in the server code, there is no reason to add it here.

rkreich: @tan-tan-kanarek, we are not trying to keep all available permissions organized in one place?
cc @erankor

tan-tan-kanarek: Yes, in the DB.
This enum contains only few permissions that used internally by the server code.

rkreich: Removed in 5eebcb79018b2c80fa31c5e84037d83274d834db

PR #3661, opened 2016-01-04
PLAT-4922 reduce memory usage at categoryUser->add

MosheMaorKaltura:
Kajam-11.6.0-PLAT-4922#time 1d


tan-tan-kanarek:
Reviewed.

PR #3665, opened 2016-01-04
Handle case where cached code file is not ready

MosheMaorKaltura:
PLAT-4842#time 3H

### DIFF
--- a/api_v3/lib/types/KalturaObject.php
+++ b/api_v3/lib/types/KalturaObject.php
@@ -371,7 +371,7 @@ abstract class KalturaObject implements IApiObject
 		$fromObjectClass = "Map_{$thisClass}_{$srcObjClass}";
 		if (!class_exists($fromObjectClass))
 		{
-			$cacheFileName = kConf::get("cache_root_path") . "api_v3/fromObject/{$fromObjectClass}.php";
+			$cacheFileName = kConf::get("cache_root_path") . "/api_v3/fromObject/{$fromObjectClass}.php";
 			$max_include_retries=10;
 			$fromObjectClassCode=null;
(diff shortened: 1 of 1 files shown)


api_v3/lib/types/KalturaObject.php
```diff
@@ -371,23 +371,31 @@ protected function doFromObject($srcObj, KalturaDetachedResponseProfile $respons
 		$fromObjectClass = "Map_{$thisClass}_{$srcObjClass}";
 		if (!class_exists($fromObjectClass))
 		{
-			$cacheFileName = kConf::get("cache_root_path") . "/api_v3/fromObject/{$fromObjectClass}.php";
-			if (!file_exists($cacheFileName))
+			$cacheFileName = kConf::get("cache_root_path") . "api_v3/fromObject/{$fromObjectClass}.php";
```
tan-tan-kanarek: Why did you remove the slash before api_v3? No damage will be cause by two slashes, but you can use realpath or trim to remove extra slashes just in case.

api_v3/lib/types/KalturaObject.php
```diff
@@ -371,23 +371,31 @@ protected function doFromObject($srcObj, KalturaDetachedResponseProfile $respons
...
-				$fromObjectClassCode = $this->generateFromObjectClass($srcObj, $fromObjectClass);
-				if (!$fromObjectClassCode)
-					return;
 				kFile::safeFilePutContents($cacheFileName, $fromObjectClassCode);
 			}
-	
-			require_once($cacheFileName);
+			if(!max_retries)
+			{
+				KalturaLog::err("Could not load File {$cacheFileName}");
```
tan-tan-kanarek: I think you better throw an exception here.

PR #3676, opened 2016-01-05
Limit amount of users per category to 5k

MosheMaorKaltura:
PLAT-4934#time 1h

### DIFF
--- a/api_v3/services/CategoryUserService.php
+++ b/api_v3/services/CategoryUserService.php
@@ -20,8 +20,15 @@ class CategoryUserService extends KalturaBaseService
 		/* @var $dbCategoryKuser categoryKuser */
 		$category = categoryPeer::retrieveByPK($categoryUser->categoryId);
 		if (!$category)
-			throw new KalturaAPIException(KalturaErrors::CATEGORY_NOT_FOUND, $categoryUser->categoryId);						
-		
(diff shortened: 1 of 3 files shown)


api_v3/services/CategoryUserService.php
```diff
@@ -20,8 +20,15 @@ function addAction(KalturaCategoryUser $categoryUser)
...
 		$category = categoryPeer::retrieveByPK($categoryUser->categoryId);
 		if (!$category)
-			throw new KalturaAPIException(KalturaErrors::CATEGORY_NOT_FOUND, $categoryUser->categoryId);						
-		
+			throw new KalturaAPIException(KalturaErrors::CATEGORY_NOT_FOUND, $categoryUser->categoryId);
+
+		$max_user_per_category=kConf::get('max_user_per_category');
+		$criteria = KalturaCriteria::create(categoryKuserPeer::OM_CLASS);
+		$criteria->addAnd(categoryKuserPeer::CATEGORY_ID, $categoryUser->categoryId, Criteria::EQUAL);
+		$users_per_category=categoryKuserPeer::doCount($criteria);
```
tan-tan-kanarek: Since KalturaCriteria on categoryKuserPeer will use sphinx, use $criteria->applyFilters() and $criteria->getRecordsCount() instead of categoryKuserPeer::doCount, it will skip the mySql and will return the count based on sphinx query only.

PR #3691, opened 2016-01-11
Validate entry is not deleted in DB when executing a playlist

MosheMaorKaltura:
PLAT-4966#time 2h


tan-tan-kanarek:
Reviewed

tan-tan-kanarek:
I expect it to fix the entry filter by ids as well.

PR #3707, opened 2016-01-13
fix dumping api request with array object in the params

yossipapi:


etameran:
Is this fix correct in all scenarios?

yossipapi:
As to PHP's documentation yes.
http://php.net/manual/en/function.curl-setopt.php

The other option is to identify cases in which one of the valus inside the postParams is an array and set Content-Type to multipart/form-data.

I prefer the first one it much more elegant.

PR #3717, opened 2016-01-18
Kajam 11.6.0 sup 6618

noam-arad:


api_v3/services/UserService.php
```diff
@@ -42,8 +44,29 @@ function addAction(KalturaUser $user)
 		$dbUser = $user->toObject($dbUser);
 		
 		try {
-			$dbUser = kuserPeer::addUser($dbUser, $user->password);
+			$lockKey = $this->getPartnerId() + $user->id;
+			$lock = kLock::create($lockKey);
+
+			try
+			{
+				if ($lock && $lock->lock(self::ADD_USER_LOCK_GRAB_TIMEOUT, self::ADD_USER_LOCK_HOLD_TIMEOUT))
```
shpiNi: @noam-arad , in case there's an issue with the caching mechanism inside kLock::create, no users could be added during that time.
might want to use if($lock && !$lock->lock....) -> throw exception, else add user.
also you can use kLock::runLocked with the addAction function as a callback.

PR #3728, opened 2016-01-18
Add support in cue points in baseEntry clone options

MosheMaorKaltura:


plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -249,6 +260,32 @@ public function objectReplaced(BaseObject $object, BaseObject $replacingObject,
...
+	/* (non-PHPdoc)
+	 * @see kObjectCopiedEventConsumer::objectCopied()
+	 */
+	public function objectCopied(BaseObject $fromObject, BaseObject $toObject)
+	{
+		if($fromObject instanceof entry)
+		{
+			$c = new KalturaCriteria();
+			$c->add( CuePointPeer::ENTRY_ID, $fromObject->getId() );
+			$c->addAscendingOrderByColumn(CuePointPeer::CREATED_AT);
```
tan-tan-kanarek: Why does the order matters? You could use CuePointPeer::retrieveByEntryId instead.

plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -249,6 +260,32 @@ public function objectReplaced(BaseObject $object, BaseObject $replacingObject,
...
+			$cpToCopy = CuePointPeer::doSelect($c);
+			if ( count( $cpToCopy ) <= self::MAX_CUE_POINTS_TO_COPY )
```
tan-tan-kanarek: You should count the cue-points that copied and not the cue-points that returned from the select.
For example, you may have 10,000 answer cue-point that will be skipped in shouldCloneByProperty, and it will cause that no other cue-point types will be copied.

tan-tan-kanarek: You better call all cue-point plugins to check if their copy is enabled and select only the cue-points of these types, instead of selecting only 1000 and checking them one by one.

alpha/lib/model/entry.php
```diff
@@ -3487,9 +3488,13 @@ public function shouldCloneByProperty($property)
 				$currentOption = $cloneOption->getItemType();
 				$currentType = $cloneOption->getRule();
 
-				if ($currentOption ==  $property && $currentType == CloneComponentSelectorType::EXCLUDE_COMPONENT)
+				if ($currentOption ==  $property)
 				{
-					$cloneCategories = false;
+					if ($currentType == CloneComponentSelectorType::EXCLUDE_COMPONENT)
+						$cloneCategories = false;
```
tan-tan-kanarek: instead of setting $cloneCategories, you could simply return false, then you won't need else, break, setting to default at the beginning of the method.

plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -167,6 +167,17 @@ public function shouldConsumeReplacedEvent(BaseObject $object)
 		return false;
 	}
 
+	/* (non-PHPdoc)
+	 * @see kObjectCopiedEventConsumer::shouldConsumeCopiedEvent()
+	 */
+	public function shouldConsumeCopiedEvent(BaseObject $fromObject, BaseObject $toObject)
```
tan-tan-kanarek: Where this method used?

alpha/lib/model/entry.php
MosheMaorKaltura: $cloneCategories is essential for the use of default value.

PR #3738, opened 2016-01-20
Use secure random function to generate password reset tokens

dozernz:
I've modified this function to use a secure random byte source. mcrypt_create_iv() is used instead of random_bytes() so the change is compatible with PHP < 7.


kaltura-hooks:
Hi @dozernz,

PR #3744, opened 2016-01-21
Kajam 11.7.0 plat 4839

noam-arad:


batch/batches/Import/KAsyncImport.class.php
```diff
@@ -197,10 +197,18 @@ private function fetchFile(KalturaBatchJob $job, KalturaImportJobData $data)
 				
 				KalturaLog::info("headers " . print_r($curlHeaderResponse, true));
 				$pluginInstances = KalturaPluginManager::getPluginInstances('IKalturaImportHandler');
+
```
tan-tan-kanarek: You can't relate to drop-folder plugin in a core class.

tan-tan-kanarek: Also, ARF files could be ingested also outside the scope of drop-folder.

PR #3747, opened 2016-01-21
add new filter to return quiz user entries without any anonymous user…

noam-arad:
…s data

PLAT-4970

### DIFF
--- a/plugins/cue_points/quiz/lib/api/filters/KalturaQuizUserEntryNoAnonymousUsersFilter.php
+++ b/plugins/cue_points/quiz/lib/api/filters/KalturaQuizUserEntryNoAnonymousUsersFilter.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * @package api
+ * @subpackage filters
+ */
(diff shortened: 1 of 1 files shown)


plugins/cue_points/quiz/lib/api/filters/KalturaQuizUserEntryNoAnonymousUsersFilter.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * @package api
+ * @subpackage filters
+ */
+class KalturaQuizUserEntryNoAnonymousUsersFilter extends KalturaQuizUserEntryFilter
+{
+	public function getListResponse(KalturaFilterPager $pager, KalturaDetachedResponseProfile $responseProfile = null)
```
tan-tan-kanarek: I don’t think it makes sense to create class extension for it, it just another filter attribute on KalturaQuizUserEntryFilter, you can call it isAnonymous and if it’s true or false apply userIdNotIn or the userIdIn accordingly.

PR #3754, opened 2016-01-24
Kajam 11.7.0 plat 4970 1

MosheMaorKaltura:


plugins/cue_points/quiz/lib/api/filters/KalturaQuizUserEntryFilter.php
```diff
@@ -5,4 +5,28 @@
...
+		if(!is_null($this->isAnonymous))
+		{
+			if(KalturaNullableBoolean::toBoolean($this->isAnonymous)===false)
+			{
+				if($this->userIdNotIn==null)
+				{
+					$anonKuserIds = "";
+					$anonKusers = kuserPeer::getKuserByPartnerAndUids(kCurrentContext::getCurrentPartnerId(), array('', 0));
+					foreach ($anonKusers as $anonKuser) {
+						$anonKuserIds .= $anonKuser->getKuserId().",";
```
shpiNi: @MosheMaorKaltura, since this is an API side filter, getPuserId() should be used here.
actually i think there's no need to look for these users and iterate each one - just put their names on userIdNotIn.

plugins/cue_points/quiz/lib/api/filters/KalturaQuizUserEntryFilter.php
```diff
@@ -5,4 +5,28 @@
...
+	public function getListResponse(KalturaFilterPager $pager, KalturaDetachedResponseProfile $responseProfile = null)
+	{
+		if(!is_null($this->isAnonymous))
+		{
+			if(KalturaNullableBoolean::toBoolean($this->isAnonymous)===false)
+			{
+				if($this->userIdNotIn==null)
+				{
+					$anonKuserIds = "";
+					$anonKusers = kuserPeer::getKuserByPartnerAndUids(kCurrentContext::getCurrentPartnerId(), array('', 0));
```
tan-tan-kanarek: What about null puser-id?

plugins/cue_points/quiz/lib/api/filters/KalturaQuizUserEntryFilter.php
tan-tan-kanarek: You better append the ids to array and implode it with commas after the array is complete, it will help you to avoid redundant comma at the end of the string.

plugins/cue_points/quiz/lib/api/filters/KalturaQuizUserEntryFilter.php
```diff
@@ -5,4 +5,28 @@
...
+	/**
+	 * @var KalturaNullableBoolean
+	 */
+	public $isAnonymous;
+
+	public function getListResponse(KalturaFilterPager $pager, KalturaDetachedResponseProfile $responseProfile = null)
+	{
+		if(!is_null($this->isAnonymous))
+		{
+			if(KalturaNullableBoolean::toBoolean($this->isAnonymous)===false)
```
tan-tan-kanarek: Please handle also case that (KalturaNullableBoolean::toBoolean($this->isAnonymous)===**true**).

PR #3761, opened 2016-01-25
Move anonymous handling upper layer to KalturaUserEntry

MosheMaorKaltura:
PLAT-4970#time 3h


api_v3/lib/types/filters/KalturaUserEntryFilter.php
```diff
@@ -103,6 +109,23 @@ protected function fixFilterUserId()
 			$this->userIdIn = $this->preparePusersToKusersFilter( $this->userIdIn );
```
tan-tan-kanarek: Please handle $this->userIdNotIn in the same way.

PR #3778, opened 2016-02-01
Kajam 11.8.0 plat 4951 enqryption at rest

anatolkaltura:


alpha/lib/model/asset.php
```diff
@@ -734,6 +735,9 @@ public function setStatusLocalReady()
 	
 	public function getActualSourceAssetParamsIds()		{return $this->getFromCustomData(self::CUSTOM_DATA_FIELD_ACTUAL_SOURCE_ASSET_PARAMS_IDS);}
 	public function setActualSourceAssetParamsIds($v)	{$this->putInCustomData(self::CUSTOM_DATA_FIELD_ACTUAL_SOURCE_ASSET_PARAMS_IDS, $v);}
+
```
tan-tan-kanarek: It's relevant only to flavor-asset and not to all asset types.

api_v3/lib/types/conversionProfile/KalturaAsset.php
```diff
@@ -115,6 +115,12 @@ class KalturaAsset extends KalturaObject implements IRelatedFilterable, IApiObje
 	 */
 	public $actualSourceAssetParamsIds;
 	
+	/**
```
tan-tan-kanarek: Also, should be part of flavor-asset and not all asset types.

PR #3796, opened 2016-02-03
Plat 4923

asafrob:


tan-tan-kanarek:
I think you should at least check if the asset already exists on the entry before creating a new one.

PR #3798, opened 2016-02-04
Case insensitive search for Transfer-Encoding HTTP header

wzur:
for example, AWS ELB returns that as 'transfer-encoding' and it causes problem when
requesting embedded player loader.


kaltura-hooks:
Hi @wzur,

PR #3805, opened 2016-02-07
Allow quiz plugin for all partners

MosheMaorKaltura:
KMS-10698#time 1h


tan-tan-kanarek:
If it's always true, you can remove this method, however, this plugin depends on cue-points plugin which is no always enabled.

MosheMaorKaltura:
The function must be implemented.

tan-tan-kanarek:
You can remove the interface.
Anyway, I think it's wrong to always return true, let's discuss the use case.

PR #3808, opened 2016-02-07
image ext preserved if legit

noam-arad:


noam-arad:
created by @shpiNi reviewed by me

PR #3812, opened 2016-02-08
Kajam 11.8.0 plat 4968

MichalRadwantzor:


alpha/apps/kaltura/lib/events/consumers/kObjectReadyForIndexInheritedTreeEventConsumer.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+interface kObjectReadyForIndexInheritedTreeEventConsumer extends KalturaEventConsumer
+{
+	/**
+	 * @param BaseObject $object
+	 * @param array $partnerCriteriaParams
+	 * @param BatchJob $raisedJob
+	 * @return bool true if should continue to the next consumer
+	 */
+	public function objectReadyForIndexInheritedTreeEvent(BaseObject $object, array $partnerCriteriaParams, BatchJob $raisedJob = null);
```
tan-tan-kanarek: Since the $partnerCriteriaParams is relevant to any multi-deferred event, I think it should be implemented as an interface with setPartnerCriteriaParams method.
If you follow this path, you won't have to add isMulti argument to the raiseDeferredEvent method because you'll be able to check if the event is instance of the multi-deferred interface.

api_v3/lib/KalturaFrontController.php
```diff
@@ -216,6 +216,15 @@ public function handleMultiRequest()
...
+			kEventsManager::enableMultiDeferredEvents(true);
+		}
+		else
+		{
+			kEventsManager::enableMultiDeferredEvents(false);
+		}
+
 		for($i = $requestStartIndex; $i <= $requestEndIndex; $i++)
 		{
 			$currentParams = $listOfRequests[$i];
```
tan-tan-kanarek: If $currentParams contains ks that is different than the one in $commonParams then the multi-deferred should be disabled.

PR #3842, opened 2016-02-16
Kajam 11.9.0 add thumbnail ext

hilak:


hilak:
@assafb @tan-tan-kanarek  please review.

tan-tan-kanarek:
Generally, I think it's too low level technical detail that usually no one cares about and therefore I wouldn't expose it in the API.
You can even add a default value true to the core object for new created objects.
Just create a server side script that change this value to true or false and accepts the profile id to change.

alpha/lib/model/DeliveryProfileVod.php
```diff
@@ -75,21 +75,21 @@ protected function doGetFlavorAssetUrl(flavorAsset $flavorAsset)
 	 * @param thumbAsset $thumbAsset
 	 * @return string
 	 */
-	protected function doGetThumbnailAssetUrl(thumbAsset $thumbAsset)
+	protected function doGetThumbnailAssetUrl(thumbAsset $thumbAsset, $addExtension = false)
 	{
 		$thumbAssetId = $thumbAsset->getId();
 		$partnerId = $thumbAsset->getPartnerId();
-		$url = "/api_v3/service/thumbAsset/action/serve/partnerId/$partnerId/thumbAssetId/$thumbAssetId";
+		$url = "/api_v3/service/thumbAsset/action/serve/partnerId/$partnerId/thumbAssetId/$thumbAssetId" . ($addExtension ?  "/$thumbAssetId." . $thumbAsset->getFileExt() : "");
```
assafb: @hilak - it seems that in case $addExtension==true, $thumbAssetId will be added twice . Please double check.

assafb:
@tan-tan-kanarek  - I agree that this is a low level configuration but i still think it should be exposed in the admin console so PMs/CSMs/support/DD will be able to configure it without the need for a custom PS work.

tan-tan-kanarek:
I think they don't care about it.
Also, this pull request won't enable admin-console functionality.



נשלח ממכשיר הסמסונג שלי


-------- הודעה מקורית --------
מאת: assafb <notifications@github.com>
תאריך: 17/02/2016 13:49 (GMT+02:00)
אל: kaltura/server <server@noreply.github.com>
עותק: Jonathan Kanarek <Jonathan.Kanarek@kaltura.com>
נושא: Re: [server] Kajam 11.9.0 add thumbnail ext (#3842)


@tan-tan-kanarek<https://github.com/tan-tan-kanarek> - I agree that this is a low level configuration but i still think it should be exposed in the admin console so PMs/CSMs/support/DD will be able to configure it without the need for a custom PS work.

-

hilak:
@tan-tan-kanarek  .content_distribution/providers/comcast_mrss/lib/admin/ComcastMrssProfileConfiguration.php

Enables the admin console thing.

alpha/lib/model/DeliveryProfileVod.php
hilak: @assafb This is intentional. The URL provided here is for the thumbAsset->serve API. The URL will look like this:

http://{hostname}/api_v3/service/thumbAsset/action/serve/partnerId/$partnerId/thumbAssetId/$thumbAssetId/$thumbAssetId.ext

hilak:
@tan-tan-kanarek  the reason for the low-level change is that I don't want to break the functionality for anyone already using the functionality as is (no extension).

PR #3844, opened 2016-02-17
add deployment script for entry server node

noam-arad:


tan-tan-kanarek:
If media-servers were unregistered during the script execution, the updated-at will be changed and the live-entry will be returned again by the query, if there LiveEntryServerNode objects to this entry, one or both of them might require deletion.

PR #3845, opened 2016-02-17
Kajam 11.9.0 plat 5097

shpiNi:


hilak:
@shpiNi  Why do we need this? If you check the SaaS, we already have this HTTP notification - Flavor Asset Status Equals.Also, I don't think we should commit these templates to the server repo, but to the SaaS tag.

shpiNi:
@hilak that notification requires a specific flavor params id and status. I can't find a reason for it not to be public.

hilak:
@shpiNi  Where is the condition related to the flavor params ID?

shpiNi:
@hilak in the notification you mentioned. mine simply has less conditions.

hilak:
@shpiNi ok, didn't see that. And the content parameters?

PR #3848, opened 2016-02-18
FEC-4691 #comment sort players according to library version (descending)

amirch1:


amirch1:
problem with changes tracking, will create a new pull

PR #3850, opened 2016-02-18
Kajam 11.8.0 fix monit

noam-arad:


jessp01:
Hi @noam-arad,

Can you please explain why you moved the PID to /var/run?
It would make perfect since if we stored all our PIDs and logs under /var to begin with, but seeing how we're not, I'm not sure what this in particular fixes..

Thanks,

noam-arad:
Hi @jessp01 ,
There was a specific production issue that sometimes more than one scheduler would run since the pid file was in $LOGDIR/batch and that file is log rotated and so was the pid file then it wasn't found by monit or by the batch service itself.
The pid file shouldn't be in the log dir that is not a good place for it, the correct place in Linux envs is in /var/run. In Windows envs (very rare occasion) this location can be customized via "batch.ini" file, in our env in order to be consistent I have places it under "C:\var\run"

jessp01:
Hi Noam,

I think you can trust that I have a rather good understand as to the UNIX 
FS tree [FSH - https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard] 
in general and Linux and specific Linux distros conventions 
in particular.

Like I said, in general, I agree PIDs should not be under the log dir. 
However, since the rest of our application places files only within the 
/opt/kaltura tree, it also does not make sense for just batch to suddenly 
place it under /var/run.

Applications that place everything in a dedicated tree structure, like 
ours, should not suddenly place a specific file outside that tree.

If you compile Apache, as for instance, by yourself, you will see it will 
by no means place its PID file under /var/run, it will place it under the 
tree you provided with --prefix, if you did not provide a default prefix, 
it will place it under /usr/local.

If you look at the link I provided above you'll see that /opt is 
the place for "Optional application software packages.[5]" so, in that 
respect we are doing the right thing. Applications that were compiled by 
hand, should go to /usr/local.

I'm not sure why we setup log rotatation that takes things with the '.pid' 
suffix, looking here:
https://github.com/kaltura/server/blob/Kajam-11.9.0/configurations/logrotate/kaltura_batch.template

We only take:
@LOG_DIR@/kaltura_batch.log {
And:
@LOG_DIR@/batch/*log {


So, not sure why "@LOG_DIR@/batch/batch.pid" should be rotated...

Considering the FS logic our app uses, the best solution would be to 
create an /opt/kaltura/var/run dir and place PIDs there.

As an example, the sphinx PID is placed under:
https://github.com/kaltura/server/blob/Kajam-11.9.0/configurations/sphinx/kaltura.conf.template#L503
pid_file = @BASE_DIR@/sphinx/searchd.pid

i.e, under the /opt/kaltura tree.

It is very important to remain consistent.


May the source be with you,

Jess Portnoy

Kaltura Connect Virtual Summits 2015 - now on VOD! [connect.mediaspace.kaltura.com]

PR #3854, opened 2016-02-18
Kajam 11.8.0 plat 4620

MichalRadwantzor:
Support delete and send co_guid entryId if referenceId is not set


EladCohenKaltura:
Looks good

PR #3859, opened 2016-02-20
Kajam 11.9.0 sup 7185

MosheMaorKaltura:


alpha/apps/kaltura/lib/batch2/bcdl/kBusinessConvertDL.php
```diff
@@ -2,6 +2,25 @@
 
 class kBusinessConvertDL
 {
+
+
+private static function shouldDeleteOldAsset($oldAsset,$entry)
```
tan-tan-kanarek: This method is very specific to replacement flow, I would place the code inside replaceEntry method or at least name it to shouldDeleteMissingAssetDuringReplacement or something like it.

alpha/apps/kaltura/lib/batch2/bcdl/kBusinessConvertDL.php
```diff
@@ -2,6 +2,25 @@
...
+{
+	if ($oldAsset instanceof flavorAsset) {
+		// In case of live entry Don't drop the old asset
+		$rootEntryId = $entry->getRootEntryId();
```
tan-tan-kanarek: Instead of root entry, you better check that the source-type is RECORDED_LIVE.

PR #3862, opened 2016-02-21
Kajam 11.9.0 sup 6997

noam-arad:


api_v3/lib/types/entry/KalturaLiveStreamEntry.php
```diff
@@ -151,7 +147,27 @@ public function doFromObject($dbObject, KalturaDetachedResponseProfile $response
...
 		if(!($dbObject instanceof LiveStreamEntry))
 			return;
-			
+
+		/**
+		 * @var LiveStreamEntry @dbObject
+		 */
+		if ( kCurrentContext::$is_admin_session || kCurrentContext::getCurrentKsKuserId() == $dbObject->getKuserId()
+				|| kCurrentContext::$ks_object->verifyPrivileges(ks::PRIVILEGE_EDIT, $this->id) )
+		{
```
MosheMaorKaltura: Reverse logic, test only the negative case and null all the fields if needed.

PR #3868, opened 2016-02-22
Exposing Exceptions creations to methods enabling clients overriding …

gotlieb:
…the exceptions

PLAT-5091 #time 2d


generator/sources/zend/Kaltura/Client/ClientBase.php
```diff
@@ -907,4 +907,29 @@ public static function getKalturaNullValue()
 	{
 		return KalturaNull::getInstance();
 	}
+
+
+	public function getKaltura_Client_Exception($errorMsg, $errorCode, $args)
```
tan-tan-kanarek: Please rename the methods to getClientException and getAPIException, same as php 5.3 lib.

PR #3882, opened 2016-02-25
Kajam 11.9.0 entry server node

EladCohenKaltura:


EladCohenKaltura:
FYI @MosheMaorKaltura  @kalturaguy @asafrob

alpha/apps/kaltura/lib/model/objectfilters/EntryServerNodeFilter.class.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+class EntryServerNodeFilter extends baseObjectFilter {
+
+	/* (non-PHPdoc)
+	 * @see myBaseObject::init()
+	 */
+	public function init()
+	{
+		$this->fields = kArray::makeAssociativeDefaultValue(array(
+			'_eq_id',
```
tan-tan-kanarek: Please add all used attributes from API filter: https://github.com/kaltura/server/pull/3882/files#diff-391f8a851560a456d0d514d4052ddd7eR11

alpha/lib/model/EntryServerNodePeer.php
```diff
@@ -0,0 +1,131 @@
...
+ */
+class EntryServerNodePeer extends BaseEntryServerNodePeer {
+
+	// cache classes by their type
+	protected static $class_types_cache = array();
+
+	/* (non-PHPdoc)
+	 * @see BaseUserEntryPeer::setDefaultCriteriaFilter()
+	 */
+	public static function setDefaultCriteriaFilter()
```
tan-tan-kanarek: setDefaultCriteriaFilter doesn't add any criterion to the criteria, you can remove the entire method.

alpha/lib/model/EntryServerNodePeer.php
```diff
@@ -0,0 +1,131 @@
...
+		self::$s_criteria_filter->setFilter($c);
+	}
+
+	/**
+	 * Function returns KalturaEntryServerNode sub-type according to protocol
+	 * @var string $type
+	 * @return KalturaEntryServerNode
+	 *
+	 */
+	public static function getInstanceByType ($sourceObject, KalturaDetachedResponseProfile $responseProfile = null)
```
tan-tan-kanarek: You can't relate to API objects here.

alpha/lib/model/EntryServerNodePeer.php
```diff
@@ -0,0 +1,131 @@
...
+ * application requirements.  This class will only be generated as
+ * long as it does not already exist in the output directory.
+ *
+ * @package Core
+ * @subpackage model
+ */
+class EntryServerNodePeer extends BaseEntryServerNodePeer {
+
+	// cache classes by their type
+	protected static $class_types_cache = array();
```
tan-tan-kanarek: Please add EntryServerNodeType::LIVE_BACKUP and LIVE_PRIMARY here and remove them from the getOMClass implementation.

alpha/lib/model/EntryServerNodePeer.php
```diff
@@ -0,0 +1,131 @@
...
+	 * @return 	   array Array of matching EntryServerNodes
+	 * @throws     kCoreException
+	 */
+	public static function retrieveByEntryIdAndServerType($entryId, $serverType, PropelPDO $con = null)
+	{
+		$criteria = new Criteria();
+		$criteria->add(EntryServerNodePeer::ENTRY_ID, $entryId);
+		$criteria->add(EntryServerNodePeer::SERVER_TYPE, $serverType);
+
+		$v = EntryServerNodePeer::doSelect($criteria, $con);
```
tan-tan-kanarek: If you want only the first one, do:
return EntryServerNodePeer::doSelectOne($criteria, $con);

alpha/lib/model/EntryServerNodePeer.php
```diff
@@ -0,0 +1,131 @@
...
+	 * @param      string $entryId.
+	 * @param      PropelPDO $con the connection to use
+	 * @return     array Array of matching EntryServerNodes
+	 */
+	public static function retrieveByEntryId($entryId, PropelPDO $con = null)
+	{
+		$criteria = new Criteria();
+		$criteria->add(EntryServerNodePeer::ENTRY_ID, $entryId);
+
+		$v = EntryServerNodePeer::doSelect($criteria, $con);
```
tan-tan-kanarek: Just return the results of doSelect, if no records found you'll get an empty array.

api_v3/lib/types/EntryServerNode/base/KalturaEntryServerNode.php
```diff
@@ -0,0 +1,113 @@
...
+ * @subpackage objects
+ * @abstract
+ */
+abstract class KalturaEntryServerNode extends KalturaObject implements IRelatedFilterable
+{
+	/**
+	 * unique auto-generated identifier
+	 * @var int
+	 * @readonly
+	 * @filter eq,in
```
tan-tan-kanarek: I see no reason to filter by this id, it's meaningless.

api_v3/lib/types/EntryServerNode/base/KalturaEntryServerNodeArray.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+class KalturaEntryServerNodeArray extends KalturaTypedArray
+{
+	public static function fromDbArray($arr)
+	{
+		$newArr = new KalturaEntryServerNodeArray();
+		foreach($arr as $obj)
+		{
+			/* @var $obj EntryServerNode */
+			$nObj = EntryServerNodePeer::getInstanceByType($obj);
```
tan-tan-kanarek: implement getInstanceByType on KalturaEntryServerNode, implement IApiObjectFactory and call the method accordingly.

api_v3/lib/types/EntryServerNode/base/KalturaEntryServerNode.php
```diff
@@ -0,0 +1,113 @@
...
+	 * @insertonly
+	 * @filter eq,in
+	 */
+	public $entryId;
+
+	/**
+	 * @var int
+	 * @filter eq,in
+	 */
+	public $serverNodeId;
```
tan-tan-kanarek: Also should be readonly.

api_v3/lib/types/EntryServerNode/live/KalturaLiveEntryServerNode.php
```diff
@@ -0,0 +1,36 @@
+<?php
+/**
+ * @package api
+ * @subpackage objects
+ */
+class KalturaLiveEntryServerNode extends KalturaEntryServerNode {
```
tan-tan-kanarek: I see no reason to have array of parameters.
Please the parameters directly on the object.

api_v3/services/EntryServerNodeService.php
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	/**
+	 * Adds a entry_user_node to the Kaltura DB.
+	 *
+	 * @action add
+	 * @param KalturaEntryServerNode $entryServerNode
+	 * @return KalturaEntryServerNode
+	 */
+	public function addAction(KalturaEntryServerNode $entryServerNode)
```
tan-tan-kanarek: In what use case we call add action?
The EntryServerNode should be created on the server side only (for now).

deployment/base/sql/01.kaltura_ce_tables.sql
```diff
@@ -2448,3 +2448,18 @@ CREATE TABLE app_token
 	PRIMARY KEY (id),
 	KEY int_id_index (int_id)
 )ENGINE=INNODB DEFAULT CHARSET=utf8;
+
+CREATE TABLE entry_server_node
+(
+	id INTEGER  NOT NULL AUTO_INCREMENT,
```
tan-tan-kanarek: Please use bigint(20) instead.

api_v3/lib/types/EntryServerNode/live/KalturaLiveEntryServerNode.php
EladCohenKaltura: This is since it can have multiple of each one and we don't want multiple DB lines for the same node

api_v3/services/EntryServerNodeService.php
EladCohenKaltura: If we want to move the logic from the backend to the actual media server we need this functionality available @kalturaguy

alpha/config/schema.xml
```diff
@@ -2130,4 +2130,19 @@
 		<column name="token" type="longvarchar"/>
 		<column name="custom_data" type="longvarchar"/>
 	</table>
+    <table skipSql="false" abstract="false" name="entry_server_node">
+        <column primaryKey="true" required="true" autoIncrement="true" lazyLoad="false" name="id" type="integer"/>
```
yossipapi: change id type to bigint(20)

alpha/lib/enums/EntryServerNodeStatus.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * @package Core
+ * @subpackage model.enum
+ */
+interface EntryServerNodeStatus extends BaseEnum{
```
yossipapi: This is the same as LiveEntryStatus so there is no reason to keep both ... choose one or the other and create the necessary client libs accordingly.

alpha/lib/enums/EntryServerNodeType.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * @package Core
+ * @subpackage model.enum
+ */
+interface EntryServerNodeType extends BaseEnum
```
yossipapi: Here as well this is the same as MediaServerIndex so there is no reason to keep both ... choose one or the other and create the necessary client libs accordingly.

alpha/apps/kaltura/lib/kObjectDeleteHandler.php
```diff
@@ -138,6 +138,15 @@ protected function entryDeleted(entry $entry)
 		}
 		
 		kJobsManager::addDeleteJob($entry->getPartnerId(), DeleteObjectType::CATEGORY_ENTRY, $filter);
+
```
yossipapi: Maybe move this code to the peer and create a function like deleteEntryServerNodes and provide the entry id $entry->getId()

alpha/apps/kaltura/lib/model/objectfilters/EntryServerNodeFilter.class.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+class EntryServerNodeFilter extends baseObjectFilter {
+
+	/* (non-PHPdoc)
+	 * @see myBaseObject::init()
+	 */
+	public function init()
+	{
+		$this->fields = kArray::makeAssociativeDefaultValue(array(
+			'_gte_created_at',
```
yossipapi: Don't think we will need to search by created_at.
Is this any scenario that will require this?

alpha/lib/model/LiveEntry.php
```diff
@@ -313,21 +313,23 @@ public function getLiveStreamConfigurations($protocol = 'http', $tag = null, $cu
...
 		$isExternalMediaServerStream = false;
-		
-		$kMediaServers = $this->getMediaServers();
-		if(count($kMediaServers))
+		/* @var $liveEntryServerNode LiveEntryServerNode */
+		$liveEntryServerNode = null;
+		/* @var $serverNode WowzaMediaServerNode */
+		$serverNode = null;
+
+		$liveEntryServerNodes = $this->getLiveEntryServerNodes();
```
yossipapi: This will return live entries which are in status "BROADCASTING" and "AUTHENTICATED" as well ... this changes the previous logic that would have returned kMediaservers only when the status was PLAYABLE.

filter in "getLiveEntryServerNodes" only entryServerNodes with status PLAYABLE

alpha/lib/model/LiveEntry.php
```diff
@@ -335,28 +337,26 @@ public function getLiveStreamConfigurations($protocol = 'http', $tag = null, $cu
...
-					if($primaryMediaServer)
-						$isExternalMediaServerStream = $primaryMediaServer->getIsExternalMediaServer();
-					else 
-						KalturaLog::debug("Cannot retrrive extra information for un-registered media server with host name [" . $kMediaServer->getHostname() . "]");
-				}
+
+				$liveEntryServerNode = array_shift($liveEntryServerNodes);
+				$serverNode = ServerNodePeer::retrieveByPK($liveEntryServerNode->getServerNodeId());
+				$primaryMediaServer = $serverNode;
+				if ($serverNode)
```
yossipapi: combine the two if's into one it would be much more simpler.

alpha/lib/model/LiveEntry.php
```diff
@@ -335,28 +337,26 @@ public function getLiveStreamConfigurations($protocol = 'http', $tag = null, $cu
...
 			{
-				$kMediaServer = reset($kMediaServers);
-				if($kMediaServer && $kMediaServer instanceof kLiveMediaServer)
-				{
-					$backupMediaServer = $kMediaServer->getMediaServer();
-					$backupApplicationName = $kMediaServer->getApplicationName();
-				}
+				$liveEntryServerNode = reset($liveEntryServerNodes);
+				$serverNode = ServerNodePeer::retrieveByPK($liveEntryServerNode->getServerNodeId());
+				$backupMediaServer = $serverNode;
```
yossipapi: check if $serverNode (same logic as for primaryMediaServer)

alpha/lib/model/LiveEntry.php
```diff
@@ -538,14 +537,14 @@ public function getMediaServer($currentDcOnly = false)
...
-		foreach($kMediaServers as $kMediaServer)
+		/* @var LiveEntryServerNode $liveEntryServerNode*/
+		foreach($liveEntryServerNodes as $liveEntryServerNode)
 		{
-			if($kMediaServer instanceof kLiveMediaServer)
-			{
-				$hostnames[$kMediaServer->getIndex()] = $kMediaServer->getHostname();
-			}
+			/* @var WowzaMediaServerNode $serverNode*/
+			$serverNode = ServerNodePeer::retrieveByPK($liveEntryServerNode->getServerNodeId());
```
yossipapi: Check if serverNode type is of type MediaServer

alpha/lib/model/LiveEntry.php
```diff
@@ -538,14 +537,14 @@ public function getMediaServer($currentDcOnly = false)
...
+			$hostnames[$serverNode->getType()] = $serverNode->getHostname();
```
yossipapi: This is wrong should be:
$hostnames[liveEntryServerNode->getServerType()] = $serverNode->getHostname();

alpha/lib/model/LiveEntry.php
```diff
@@ -577,23 +575,29 @@ private static function getCacheType()
...
 	}
-	
-	private function isCacheValid(kLiveMediaServer $kMediaServer)
+
+	/**
+	 * @param LiveEntryServerNode $liveEntryServerNode
+	 * @return bool|mixed
+	 * @throws Exception
+	 */
+	private function isCacheValid($liveEntryServerNode)
```
yossipapi: Params should LiveEntryServerNode $liveEntryServerNode

alpha/lib/model/LiveEntry.php
```diff
@@ -577,23 +575,29 @@ private static function getCacheType()
...
 	{
 		$cacheType = self::getCacheType();
 		$cacheStore = kCacheManager::getSingleLayerCache($cacheType);
 		if(! $cacheStore)
 		{
 			KalturaLog::warning("Cache store [$cacheType] not found");
-			$lastUpdate = time() - $kMediaServer->getTime();
+			$lastUpdate = time() - $liveEntryServerNode->getUpdatedAt(null);
```
yossipapi: Note that on the kMediaServer getTime always returned the same value ... the time on kMediaServer was set in the __construct therefore never changed.

I agree that this change fixes a bug here that we never noticed but we need to make sure this wont trigger any hidden issues that we never noticed.

Please verify.

alpha/lib/model/LiveEntry.php
```diff
@@ -609,19 +613,28 @@ private function storeInCache($key)
 			KalturaLog::debug("cacheStore is null. cacheType: $cacheType . returning false");
 			return false;
 		}
-		return $cacheStore->set($key, true, kConf::get('media_server_cache_expiry', 'local', self::DEFAULT_CACHE_EXPIRY));
+		KalturaLog::debug("Set cache key [$key] from store [$cacheType] ");
+		$returnValue =  $cacheStore->set($key, true, kConf::get('media_server_cache_expiry', 'local', self::DEFAULT_CACHE_EXPIRY));
+		$ans = $cacheStore->get($key);
```
yossipapi: You never return the $returnValue.

Remove the:
$ans = $cacheStore->get($key); and the debug it duplicate to the previous debug.

alpha/lib/model/LiveEntry.php
```diff
@@ -609,19 +613,28 @@ private function storeInCache($key)
...
+		KalturaLog::debug("Get cache key [$key] from store [$cacheType] returned [$ans] and reurn value was [$returnValue]");
 	}
 	
-	public function setMediaServer($index, $hostname, $applicationName = null)
+	public function setMediaServer($index, $hostname)
```
yossipapi: Calling functions still send applicationName so please remove.

yossipapi: Think about moving the creating of the entryServerNode object to this function and adding status param to this function

alpha/lib/model/LiveEntry.php
```diff
@@ -609,19 +613,28 @@ private function storeInCache($key)
...
 			throw new kCoreException("Media server with host name [$hostname] not found", kCoreException::MEDIA_SERVER_NOT_FOUND);
-		
-		$key = $this->getId() . "_{$hostname}_{$index}";
+
+		/* @var $dbLiveEntryServerNode LiveEntryServerNode*/
+		$dbLiveEntryServerNode = EntryServerNodePeer::retrieveByEntryIdAndServerType($this->getId(), $index);
+		if (!$dbLiveEntryServerNode)
+			throw new KalturaAPIException("Failed to find valid entry server node ", KalturaErrors::ENTRY_SERVER_NODE_NOT_FOUND);
+
+		$key = $this->getEntryServerNodeCacheKey($dbLiveEntryServerNode);
```
yossipapi: There is no need to change the cache key leave it as it was.

alpha/lib/model/LiveEntry.php
```diff
@@ -629,80 +642,80 @@ public function setMediaServer($index, $hostname, $applicationName = null)
 
 		KalturaLog::debug("about to setMediaServer. index: $index, hostname: $hostname");
 		$this->setLastBroadcast(kApiCache::getTime());
-		$server = new kLiveMediaServer($index, $hostname, $mediaServerNode->getDc(), $mediaServerNode->getId(), $applicationName ? $applicationName : $mediaServerNode->getApplicationName());
-		$this->putInCustomData("server-$index", $server, LiveEntry::CUSTOM_DATA_NAMESPACE_MEDIA_SERVERS);
-		$this->setLiveStatus(LiveEntryStatus::PLAYABLE, $index);
+		$dbLiveEntryServerNode->setStatus(EntryServerNodeStatus::PLAYABLE);
+		$dbLiveEntryServerNode->setServerType($index);
```
yossipapi: You already set setServerType & setServerNodeId in the service when creating the object no need to do the same here.

alpha/lib/model/LiveEntry.php
```diff
@@ -629,80 +642,80 @@ public function setMediaServer($index, $hostname, $applicationName = null)
...
+		$dbLiveEntryServerNode->setServerNodeId($mediaServerNode->getId());
+		$dbLiveEntryServerNode->save();
+		$this->setLiveStatus(EntryServerNodeStatus::PLAYABLE, $index);
```
yossipapi: I see that save on the entry is no longer called from the KalturaLiveEntryService so or add it after here after the setLiveStatus call or in KalturaLiveEntryService function setLiveStatusOrMediaServer inside the try block

alpha/lib/model/LiveEntry.php
```diff
@@ -629,80 +642,80 @@ public function setMediaServer($index, $hostname, $applicationName = null)
...
 	}
-	
+
+	public function unsetMediaServer($index)
+	{
+		// Set media server with index #index to be stopped
+		// the entry->liveStatus() is calculated from the state of both servers
+		$this->setLiveStatus(EntryServerNodeStatus::STOPPED, $index);
```
yossipapi: Add validation for the hostName as was previously done + consider moving some logic from unregisterMediaServerAction inside the unset function ... such as the doDelete, etc ...

alpha/lib/model/LiveEntry.php
```diff
@@ -629,80 +642,80 @@ public function setMediaServer($index, $hostname, $applicationName = null)
...
+	{
+		return $entryServerNode->getId()."_".$entryServerNode->getServerType()."_".$entryServerNode->getServerNodeId();
+	}
+
 	protected function isMediaServerRegistered($index, $hostname)
 	{
-		$server = $this->getFromCustomData("server-$index", LiveEntry::CUSTOM_DATA_NAMESPACE_MEDIA_SERVERS);
-		/* @var $server kLiveMediaServer */
-		if($server && $server->getHostname() == $hostname)
+		/* @var $dbLiveEntryServerNode LiveEntryServerNode*/
```
yossipapi: Please validate hostName matches and also validate entryServerNode is on status PLAYABLE as in my previous comments.

alpha/lib/model/LiveEntry.php
```diff
@@ -629,80 +642,80 @@ public function setMediaServer($index, $hostname, $applicationName = null)
...
-	
 	/**
 	 * @return bool true is list changed
 	 */
 	public function validateMediaServers()
 	{
 		$listChanged = false;
-		$kMediaServers = $this->getFromCustomData(null, LiveEntry::CUSTOM_DATA_NAMESPACE_MEDIA_SERVERS, array());
-		foreach($kMediaServers as $key => $kMediaServer)
+		$dbLiveEntryServerNodes = EntryServerNodePeer::retrieveByEntryId($this->id);
```
yossipapi: Again retrieve only liveEntryServerNodes in status PLAYABLE.

alpha/lib/model/LiveEntry.php
```diff
@@ -629,80 +642,80 @@ public function setMediaServer($index, $hostname, $applicationName = null)
...
+		foreach($dbLiveEntryServerNodes as $dbLiveEntryServerNode)
 		{
-			if(!$kMediaServer || ! $this->isCacheValid($kMediaServer))
+			if (! $this->isCacheValid($dbLiveEntryServerNode))
 			{
 				$listChanged = true;
-				KalturaLog::info("Removing media server [" . ($kMediaServer ? $kMediaServer->getHostname() : $key) . "]");
-				$this->removeFromCustomData($key, LiveEntry::CUSTOM_DATA_NAMESPACE_MEDIA_SERVERS);
+				// we need to set its live_status field if the supported line is deleted
+				$this->setLiveStatus(EntryServerNodeStatus::STOPPED, $dbLiveEntryServerNode->getServerType());
```
yossipapi: Maybe call unsetMediaServer here instead of duplicating the logic

alpha/lib/model/LiveEntry.php
```diff
@@ -629,80 +642,80 @@ public function setMediaServer($index, $hostname, $applicationName = null)
...
+
 	/**
-	 * @return array<kLiveMediaServer>
+	 * @return array<LiveEntryServerNode>
 	 */
-	public function getMediaServers()
+	public function getLiveEntryServerNodes()
 	{
-		return $this->getFromCustomData(null, LiveEntry::CUSTOM_DATA_NAMESPACE_MEDIA_SERVERS, array());
+		return EntryServerNodePeer::retrieveByEntryId($this->getId());
```
yossipapi: Here you should return only EntryServerNodePeer that hold entries in status PLAYABLE.

alpha/lib/model/LiveEntry.php
```diff
@@ -803,24 +816,6 @@ public function isConvertingSegments()
 		return false;
 	}
 	
-	protected function getTrackColumns ()
```
yossipapi: Why did you remove this functions ... you need to adjust them to your new code.

alpha/lib/model/LiveEntryServerNode.php
```diff
@@ -0,0 +1,31 @@
+<?php
+/**
+ * @package Core
+ * @subpackage model
+ */
+class LiveEntryServerNode extends EntryServerNode{
```
yossipapi: KalturaLiveEntryServerNode has only one param "parameters" there is a mismatch between the classes.

alpha/lib/model/LiveEntry.php
EladCohenKaltura: Seems OK

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -204,49 +204,52 @@ private function ingestAsset(entry $entry, $dbAsset, $filename)
...
 		{
-			$code = $ex->getCode();
-			switch($code)
-			{
-				case kCoreException::MEDIA_SERVER_NOT_FOUND :
-					throw new KalturaAPIException(KalturaErrors::MEDIA_SERVER_NOT_FOUND, $hostname);
-				default:
-					throw $ex;
-			}
+			$liveEntryServerNode = new KalturaLiveEntryServerNode();
```
yossipapi: Better move this code inside setMediaServer in LiveEntry while passing the liveEntryStatus

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -204,49 +204,52 @@ private function ingestAsset(entry $entry, $dbAsset, $filename)
...
+		$dbLiveEntryServerNode->setServerNodeId($dbServerNode->getId());
+		$dbLiveEntryServerNode->setStatus($liveEntryStatus);
+		$dbLiveEntryServerNode->setUpdatedAt(time());
+
 		// setRedirectEntryId to null in all cases, even for broadcasting...
 		$dbEntry->setRedirectEntryId(null);
 
-		if($dbEntry->save())
+		if($dbLiveEntryServerNode->save() && $dbEntry->save()
+			&& $this->setLiveStatusOrMediaServer($liveEntryStatus, $dbEntry ,$mediaServerIndex, $hostname, $applicationName))
```
yossipapi: Need to call save inside setLiveStatusOrMediaServer or after

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -204,49 +204,52 @@ private function ingestAsset(entry $entry, $dbAsset, $filename)
...
+			$dbLiveEntryServerNode->setServerType($mediaServerIndex);
+			$dbLiveEntryServerNode->setPartnerId($this->getPartnerId());
+		} else {
+			$dbLiveEntryServerNode = $dbEntryServerNode;
 		}
 
+		/* @var $dbLiveEntryServerNode LiveEntryServerNode */
+		$dbLiveEntryServerNode->setServerNodeId($dbServerNode->getId());
+		$dbLiveEntryServerNode->setStatus($liveEntryStatus);
+		$dbLiveEntryServerNode->setUpdatedAt(time());
```
yossipapi: Previously we never updated the kMediaServer time attribute ... why do we need to update for each register the updateAt value now?

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -357,40 +387,42 @@ private function createRecordedEntry(LiveEntry $dbEntry, $mediaServerIndex)
...
 		KalturaLog::debug("Entry [$entryId] from mediaServerIndex [$mediaServerIndex] with hostname [$hostname]");
-
-		$dbEntry = entryPeer::retrieveByPK($entryId);
-		if (!$dbEntry || !($dbEntry instanceof LiveEntry))
+		/* @var $dbLiveEntry LiveEntry */
+		$dbLiveEntry = entryPeer::retrieveByPK($entryId);
+		if (!$dbLiveEntry || !($dbLiveEntry instanceof LiveEntry))
 			throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $entryId);
 
-		$dbEntry->unsetMediaServer($mediaServerIndex, $hostname);
```
yossipapi: We have to pass the hostname when calling unregister to validate the entryServerNode being unregistered matches the hostName that called the unregister call

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -357,40 +387,42 @@ private function createRecordedEntry(LiveEntry $dbEntry, $mediaServerIndex)
...
-		}
+		/* @var $dbLiveEntryServerNode LiveEntryServerNode */
+		$dbLiveEntrySeverNode = EntryServerNodePeer::retrieveByEntryIdAndServerType($entryId, $mediaServerIndex);
 
-		if ( count( $dbEntry->getMediaServers() ) == 0 )
+		if ($dbLiveEntrySeverNode)
 		{
-			if ( $dbEntry->getCurrentBroadcastStartTime() )
+			EntryServerNodePeer::doDelete($dbLiveEntrySeverNode);
+			$dbLiveEntry->setLastBroadcastEndTime(kApiCache::getTime());
```
yossipapi: Where did this call came from?
previously we did $dbEntry->setCurrentBroadcastStartTime( 0 ); where did ti go to?

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -357,40 +387,42 @@ private function createRecordedEntry(LiveEntry $dbEntry, $mediaServerIndex)
...
-
-		$dbEntry = entryPeer::retrieveByPK($entryId);
-		if (!$dbEntry || !($dbEntry instanceof LiveEntry))
+		/* @var $dbLiveEntry LiveEntry */
+		$dbLiveEntry = entryPeer::retrieveByPK($entryId);
+		if (!$dbLiveEntry || !($dbLiveEntry instanceof LiveEntry))
 			throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $entryId);
 
-		$dbEntry->unsetMediaServer($mediaServerIndex, $hostname);
+		$dbLiveEntry->unsetMediaServer($mediaServerIndex);
```
yossipapi: The logic below is wrong ... will cause:
1. setRedirectEntryId will never be called.
2. setCurrentBroadcastStartTime will never be reset.

You need to check if count(entryServerNodes) by entry id is 0 than do the previous logic else same logic as before.

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -405,7 +437,6 @@ function unregisterMediaServerAction($entryId, $hostname, $mediaServerIndex)
 	function validateRegisteredMediaServersAction($entryId)
 	{
 		KalturaResponseCacher::disableCache();
-		$this->dumpApiRequest($entryId, false);
 
 		$dbEntry = entryPeer::retrieveByPK($entryId);
 		if (!$dbEntry || !($dbEntry instanceof LiveEntry))
```
yossipapi: with your new logic there is no need to call $dbEntry->save(); anymore

api_v3/lib/types/EntryServerNode/base/KalturaEntryServerNode.php
```diff
@@ -0,0 +1,139 @@
...
+	/**
+	 * @var int
+	 * @readonly
+	 */
+	public $partnerId;
+
+	/**
+	 * @var time
+	 * @readonly
+	 * @filter lte,gte,order
```
yossipapi: Do we really need the filters here? 
This will be an intensive db query since we have no index on the created at.

api_v3/lib/types/EntryServerNode/base/KalturaEntryServerNode.php
```diff
@@ -0,0 +1,139 @@
...
+	 */
+	public $createdAt;
+
+	/**
+	 * @var time
+	 * @readonly
+	 * @filter lte,gte,order
+	 */
+	public $updatedAt;
```
yossipapi: Same as the filters on created at.

api_v3/lib/types/EntryServerNode/base/KalturaEntryServerNode.php
```diff
@@ -0,0 +1,139 @@
...
+
+	/**
+	 * @var KalturaEntryServerNodeStatus
+	 * @readonly
+	 * @filter eq
```
yossipapi: Maybe think of adding in

api_v3/lib/types/EntryServerNode/base/KalturaEntryServerNode.php
```diff
@@ -0,0 +1,139 @@
+<?php
+/**
+ * @package api
+ * @subpackage objects
+ * @abstract
+ */
+abstract class KalturaEntryServerNode extends KalturaObject implements IRelatedFilterable, IApiObjectFactory
```
yossipapi: The is a bit of mismatch in the readOnly and insertOnly for the below attributes.
Either make all of them readOnly or make attributes such as: entryId & serverNodeId & status & serverType as inserOnly

api_v3/lib/types/EntryServerNode/base/KalturaEntryServerNodeStatus.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * @package api
+ * @subpackage enum
+ */
+class KalturaEntryServerNodeStatus extends KalturaDynamicEnum implements EntryServerNodeStatus{
```
yossipapi: Why does this class extends KalturaDynamicEnum do you have any plugins extending this class?
if not extend KalturaEnum.

api_v3/lib/types/EntryServerNode/base/filters/KalturaEntryServerNodeFilter.php
```diff
@@ -0,0 +1,66 @@
...
+	/**
+	 * @param KalturaFilterPager $pager
+	 * @param KalturaDetachedResponseProfile $responseProfile
+	 * @return KalturaListResponse
+	 * @throws KalturaAPIException
+	 */
+	public function getListResponse(KalturaFilterPager $pager, KalturaDetachedResponseProfile $responseProfile = null)
+	{
+		if ($this->entryIdEqual == null && $this->entryIdNotIn == null && $this->entryIdIn == null &&
+			$this->idEqual == null && $this->idNotIn == null && $this->idIn == null &&
```
yossipapi: Where did the notIn came from?

api_v3/lib/types/EntryServerNode/base/filters/KalturaEntryServerNodeFilter.php
```diff
@@ -0,0 +1,66 @@
...
+		if($this->entryIdEqual != null)
+		{
+			$entry = entryPeer::retrieveByPK($this->entryIdEqual);
+			if(!$entry)
+				throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $this->entryIdEqual);
+		}
+
+		if($this->entryIdIn != null)
+		{
+			$entryArray = entryPeer::retrieveByPKs($this->entryIdIn);
```
yossipapi: $this->entryIdIn you need to send explode(",", $this->entryIdIn)

api_v3/lib/types/EntryServerNode/base/filters/KalturaEntryServerNodeFilter.php
```diff
@@ -0,0 +1,66 @@
...
+			if(!$entryArray)
```
yossipapi: here you should check count on entryArray?
Also what about cases where some of the entries in the entryIdIn did not return in the entryArray don't you want to remove them from the filter? 
or simply do implode(",", $entryArray)

api_v3/lib/types/EntryServerNode/base/filters/KalturaEntryServerNodeFilter.php
```diff
@@ -0,0 +1,66 @@
...
+				throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $this->entryIdIn);
```
yossipapi: Error should be ENTRY_IDS_NOT_FOUND

api_v3/lib/types/EntryServerNode/base/filters/KalturaEntryServerNodeFilter.php
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		$entryServerNodeFilter = $this->toObject();
+
+		$c = KalturaCriteria::create(EntryServerNodePeer::OM_CLASS);
+		$entryServerNodeFilter->attachToCriteria($c);
+		$dbEntryServerNodes = EntryServerNodePeer::doSelect($c);
+
+		$resultCount = count($dbEntryServerNodes);
```
yossipapi: why not gust do $pager->attachToCriteria($c);

api_v3/lib/types/EntryServerNode/live/KalturaLiveStreamParams.php
```diff
@@ -0,0 +1,43 @@
+<?php
+/**
+ * @package api
+ * @subpackage objects
+ */
+class KalturaLiveStreamParams extends KalturaObject {
+
+	/**
+	 * Bit rate of the stream. (i.e. 900)
+	 * @var string
```
yossipapi: should be of type int same as KalturaFlavorParams

api_v3/lib/types/EntryServerNode/live/filters/KalturaLiveEntryServerNodeFilter.php
```diff
@@ -0,0 +1,8 @@
+<?php
+/**
+ * @package api
+ * @subpackage filters
+ */
+class KalturaLiveEntryServerNodeFilter extends KalturaLiveEntryServerNodeBaseFilter
```
yossipapi: LiveEntryServerNodeFilter s not relevant there is not filter defined on it

api_v3/lib/types/EntryServerNode/live/filters/base/KalturaLiveEntryServerNodeBaseFilter.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+ * @package api
+ * @subpackage filters.base
+ * @abstract
+ */
+abstract class KalturaLiveEntryServerNodeBaseFilter extends KalturaEntryServerNodeFilter
```
yossipapi: why is this class needed

api_v3/lib/types/EntryServerNode/live/filters/orderEnums/KalturaLiveEntryServerNodeOrderBy.php
```diff
@@ -0,0 +1,8 @@
+<?php
+/**
+ * @package api
+ * @subpackage filters.enum
+ */
+class KalturaLiveEntryServerNodeOrderBy extends KalturaEntryServerNodeOrderBy
```
yossipapi: why is this class needed

api_v3/services/EntryServerNodeService.php
yossipapi: Based on your decision about the readOnly/insert only please decide what to do with the add/update actions in this class.

deployment/permissions/service.entryservernode.ini
```diff
@@ -0,0 +1,26 @@
+[action_permission_items]
+permissionItem1.service = entryservernode
```
yossipapi: Based on your decision about keeping/removing the add/update/etc ... update this file accordingly.

deployment/updates/scripts/2016_02_17_move_live_entry_to_entry_server_node.php
```diff
@@ -0,0 +1,67 @@
...
+CONST LIMIT = 500;
+$initialCreatedAtValue = '2000-01-01 00:00:00';
+
+//Script doesn't have to start at the beginning of time, just call it with the initial time you want.
+if ($argc == 2)
+{
+	$initialCreatedAtValue = $argv[1];
+}
+
+$c = new Criteria();
```
yossipapi: You need only stream which are live.

deployment/updates/scripts/2016_02_17_move_live_entry_to_entry_server_node.php
```diff
@@ -0,0 +1,67 @@
...
+				$mediaServerIds[] = $mediaServer->getMediaServerId();
+				/**
+				 * @var kLiveMediaServer $mediaServer
+				 */
+				$liveEntryServerNode = new LiveEntryServerNode();
+				$liveEntryServerNode->setEntryId($liveEntry->getId());
+				$liveEntryServerNode->setServerNodeId($mediaServer->getMediaServerId());
+				$liveEntryServerNode->setPartnerId($liveEntry->getPartnerId());
+				$liveEntryServerNode->setStatus($liveEntry->getLiveStatus());
+				$entryServerNodeType = EntryServerNodeType::LIVE_PRIMARY;
```
yossipapi: you never assign entryServerNodeType value to the liveEntryServerNode

deployment/updates/scripts/2016_02_17_move_live_entry_to_entry_server_node.php
```diff
@@ -0,0 +1,67 @@
...
+				if ($mediaServer->getIndex() == 1)
+				{
+					$entryServerNodeType = EntryServerNodeType::LIVE_BACKUP;
+				}
+//				$liveEntryServerNode->save();
```
yossipapi: why is this commented out?

alpha/lib/model/LiveEntry.php
EladCohenKaltura: Generating the same key is a overhead I don't think that we need (additional call to the DB [not sphinx] to get the hostname ) even at the cost of losing the current elements (for first deployment) that are in the cache

PR #3885, opened 2016-02-25
Kajam 11.9.0 add thumbnail ext

hilak:


hilak:
@erankor  Please review.

PR #3904, opened 2016-03-01
input validation when add a user

gotlieb:
PLAT-4655
When adding a user we are not validating the input for insert and we then
insert the data to the object causing exceptions in the log
trying to insert a read only member.
Adding the input validation prior to the user object creation
will cause the add user to fail since we are
trying to set a read only member.

@MosheMaorKaltura please review it.

### DIFF
--- a/api_v3/services/UserService.php
+++ b/api_v3/services/UserService.php
@@ -40,7 +40,6 @@ class UserService extends KalturaBaseUserService
 		}
 		$user->partnerId = $this->getPartnerId();
 
-
 		$lockKey = "user_add_" . $this->getPartnerId() . $user->id;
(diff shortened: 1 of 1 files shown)


api_v3/services/UserService.php
```diff
@@ -45,9 +45,10 @@ function addAction(KalturaUser $user)
 		return kLock::runLocked($lockKey, array($this, 'adduserImpl'), array($user));
 	}
 	
-	function addUserImpl($user)
+	function addUserImpl(KalturaUser $user)
 	{
 		$dbUser = null;
+		$user->validateForInsert();
```
noam-arad: I think you should replace these two lines (validateForInsert and toObject) with 
$user->toInsertableObject

PR #3918, opened 2016-03-03
during bulk XML upload for creating new entries, change so we first c…

noam-arad:
…reate each flavor asset and only then we add the content. This is in order to overcome the race condition described in the SUP

SUP-7267

### DIFF
--- a/plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
+++ b/plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
@@ -124,7 +124,7 @@ class BulkUploadEngineXml extends KBulkUploadEngine
 		libxml_clear_errors();
 		if(!$xdoc->loadXML($this->xslTransformedContent)){
 			$errorMessage = kXml::getLibXmlErrorDescription($this->xslTransformedContent);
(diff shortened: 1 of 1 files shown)


plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
```diff
@@ -1097,7 +1103,7 @@ protected function sendItemUpdateData($entryId, $entry, array $flavorAssets, arr
 				KBatchBase::$kClient->thumbAsset->setContent($existingthumbAssets[$thumbParamsId], $thumbAssetsResource);
 			}
 			if (strpos($thumbAssetsResource->tags, self::DEFAULT_THUMB_TAG) !== false)
-				KBatchBase::$kClient->thumbAsset->setAsDefault($thumAsset->id);
+				KBatchBase::$kClient->thumbAsset->setAsDefault($thumbsAsset->id);
```
yossipapi: Won't this throw an exception in case we failed to add/update in that case thumbsAsset wont have the id attribute set

noam-arad: I only fixed a minor typo, if you see something more major to fix we can have a look.

tan-tan-kanarek: $thumbsAsset->id is not a real attribute with value, it's a magic attribute that place a multi-request place holder in the request.

PR #3924, opened 2016-03-03
set all flavors urgency to required when replacing an entry mode

gotlieb:
SUP-7143
@MosheMaorKaltura

### DIFF
--- a/alpha/apps/kaltura/lib/batch2/model/kConvertJobData.php
+++ b/alpha/apps/kaltura/lib/batch2/model/kConvertJobData.php
@@ -247,11 +247,11 @@ class kConvertJobData extends kConvartableJobData
 			$readiness = flavorParamsConversionProfile::READY_BEHAVIOR_REQUIRED;
 
 		// Case 4
-		$jobEntry = $batchJob->getEntry();
-		if (!empty($jobEntry))
+		$entry = $batchJob->getEntry();
+		if ($entry)
 		{
 			$replacedEntryId = $jobEntry->getReplacedEntryId();
-			if (!empty($replacedEntryId))
+			if ($replacedEntryId)
 				$readiness = flavorParamsConversionProfile::READY_BEHAVIOR_REQUIRED;
 		}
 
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/lib/batch2/model/kConvertJobData.php
```diff
@@ -245,6 +246,15 @@ function calculateUrgency(BatchJob $batchJob) {
 		if($allNoImpact)
 			$readiness = flavorParamsConversionProfile::READY_BEHAVIOR_REQUIRED;
 
+		// Case 4
```
MosheMaorKaltura: rename jobEntry  to entry., it is more commonly known and used.

alpha/apps/kaltura/lib/batch2/model/kConvertJobData.php
```diff
@@ -245,6 +246,15 @@ function calculateUrgency(BatchJob $batchJob) {
...
+		$jobEntry = $batchJob->getEntry();
+		if (!empty($jobEntry))
```
MosheMaorKaltura: Instead of if (!empty($jobEntry)) use if($entry), It does the same , and more commonly used, plus less code.

gotlieb:
edited SUP-7143
@MosheMaorKaltura

alpha/apps/kaltura/lib/batch2/model/kConvertJobData.php
```diff
@@ -245,6 +246,15 @@ function calculateUrgency(BatchJob $batchJob) {
 		if($allNoImpact)
 			$readiness = flavorParamsConversionProfile::READY_BEHAVIOR_REQUIRED;
 
+		// Case 4
+		$entry = $batchJob->getEntry();
+		if ($entry)
+		{
+			$replacedEntryId = $jobEntry->getReplacedEntryId();
```
yossipapi: Where do you define jobEntry? from what I see its not defined

gotlieb:
@MosheMaorKaltura

PR #3925, opened 2016-03-03
handle cases where live entry cue-point are not marked as handled and…

yossipapi:


DBezemer:
I don't think that 86400 is sufficient time to safely delete the cuePoints, from analysis of some long running Live streams @ Clarin I found that the interval can be as big as 12 days. So I'd suggest setting the interval to 30 days or so.

yossipapi:
I am not deleting the cue point I am simply marking it as handled.. you can still retrieve handled cue points via the testme console for analysis....

This is only the backend changes I also need to change the player side to not fetch cue points marked as handled (not in all cases ... there are some use cases we want to fetch them as well)

DBezemer:
Why not change the Wowza Component behavior to use an update call instead of creating a new eventCuePoint? The eventCuePoint has both a creation timestamp and an updated timestamp.

configurations/batch/batch.ini.template
```diff
@@ -747,4 +748,12 @@ friendlyName										= Recalculate Cache
 type												= KAsyncRecalculateCache
 scriptPath											= batches/Cache/KAsyncRecalculateCacheExe.php
 
+[KAsyncClearCuePoints : PeriodicWorker]
+id													= 620
+friendlyName                                        = Clear old cue points from live entry
+type                                                = KAsyncClearCuePoints
+scriptPath                                          = ../plugins/cue_points/base/batch/clearCuePonts/KAsyncClearCuePointsExe.php
+params.cuePointTypeIn                               = "thumbCuePoint.Thumb,adCuePoint.Ad,codeCuePoint.Code"
```
tan-tan-kanarek: I would implement it on filter and not on params, see KAsyncConvert as example.

plugins/cue_points/base/batch/clearCuePonts/KAsyncClearCuePoints.class.php
```diff
@@ -0,0 +1,89 @@
...
+		$pager->pageIndex = 1;
+		
+		$entries = self::$kClient->liveStream->listAction($entryFilter, $pager);
+		
+		while(count($entries->objects))
+		{
+			foreach($entries->objects as $entry)
+			{
+				/* @var $entry KalturaLiveEntry */
+				if($entry->recordStatus !== KalturaRecordStatus::DISABLED)
```
tan-tan-kanarek: Please add comment that explains why recorded entries are already handled.

plugins/cue_points/base/batch/clearCuePonts/KAsyncClearCuePoints.class.php
```diff
@@ -0,0 +1,89 @@
...
+	
+	private function clearEntryCuePoints($entry)
+	{
+		$cuePointPlugin = KalturaCuePointClientPlugin::get(self::$kClient);
+		
+		$cuePointFilter = new KalturaCuePointFilter();
+		$cuePointFilter->cuePointTypeIn = $this->getAdditionalParams("cuePointTypeIn");
+		$cuePointFilter->statusEqual = KalturaCuePointStatus::READY;
+		$cuePointFilter->orderBy = KalturaCuePointOrderBy::CREATED_AT_ASC;
+		$cuePointFilter->createdAtLessThanOrEqual = time() - $this->getAdditionalParams("cuePointCreatedAtLessThanOrEqual");
```
tan-tan-kanarek: You can use relative time here.

yossipapi:
updated to support dynamic filter generation for KPeriodicWorker

PR #3936, opened 2016-03-07
alpha/apps/kaltura/lib/cache/kApiCache.php

MosheMaorKaltura:


etameran:
Why not use an actual user id instead of just a flag?
At any rate don't use a space which may be trimmed and hard to notice , "anonymous" / "no cache" is better

alpha/lib/model/widget.php
```diff
@@ -280,4 +280,8 @@ public function getAddEmbedHtml5Support (  )	{	return (bool) $this->getFromCusto
 	// set a widget role to allow a specific widget to have extra privileges
 	public function setRoles ( $v )	{	$this->putInCustomData ( "roles" , $v );	}
 	public function getRoles (  )	{	return $this->getFromCustomData( "roles" );	}
+
+	public function setAvoidAnonymousCache ( $v )	{	$this->putInCustomData ( "avoidAnonymousCache" , $v );	}
+	public function getAvoidAnonymousCache (  )	{	return $this->getFromCustomData( "avoidAnonymousCache" );	}
```
tan-tan-kanarek: Please define default value.

PR #3938, opened 2016-03-08
Java API: Malformed UTF-8 decoding in response data

### LINKED ISSUE #3937: Java client API: Malformed UTF-8 decoding in response data in non-utf8 systems
Some UTF-8 special characters are not correctly decoded in response data, if the system default charset is not set to UTF-8.

### Example ###
`report->getTable(TOP_CONTENT_DROPOFF)` call returns video names with special characters in it which are not correctly decoded:
```
//response fragment from curl:
...Şubat...;

//response fragment from the API:
...??ubat...;
```

### Cause ###

**KalturaClientBase.java:**
```java
//String doReadRemoteInvocationResult(InputStream is) @ 307
    return new String(out.toByteArray());
```
This constructor decodes the byte array using the platform's default charset.
According to `String` javadoc:

> The behavior of this constructor when the given bytes are not valid in the default charset is unspecified. 

Furthermore,

**KalturaClientBase.java:**
```java
//String executeMethod(HttpClient client, PostMethod method) @ 355
    responseString = new String (responseBody.getBytes(), UTF8_CHARSET);
```

Again, according to the `String` javadoc, in the `getBytes()` method description:

> The behavior of this method when this string cannot be encoded in the default charset is unspecified.

### Solution ###
Providing `UTF-8` as the charset in `String` constructor should be enough to make sure the text is decoded properly:
```java
//String doReadRemoteInvocationResult(InputStream is) @ 307
    return new String(out.toByteArray(), UTF8_CHARSET);
```
Also, in the `executeMethod()` method, `responseString` should be correctly cast:
```java
//String executeMethod(HttpClient client, PostMethod method) @ 355
    responseString = new String (responseBody.getBytes(UTF8_CHARSET), UTF8_CHARSET);
```

canbulutb:
Fixes #3937: Malformed UTF-8 decoding in response data in non-utf8 platforms.

### DIFF
--- a/generator/sources/java/src/main/java/com/kaltura/client/KalturaClientBase.java
+++ b/generator/sources/java/src/main/java/com/kaltura/client/KalturaClientBase.java
@@ -304,7 +304,7 @@ abstract public class KalturaClientBase implements Serializable {
         {
         	out.write(buf,0,len);
         }
-        return new String(out.toByteArray());
+        return new String(out.toByteArray(), UTF8_CHARSET);
     }
 
 	protected String executeMethod(HttpClient client, PostMethod method) throws KalturaApiException {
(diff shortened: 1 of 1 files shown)


kaltura-hooks:
Hi @canbulutb,

PR #3949, opened 2016-03-10
Validate that entry is not live before updating crtian fields + post…

yossipapi:
… process update

### DIFF
--- a/alpha/lib/model/LiveEntry.php
+++ b/alpha/lib/model/LiveEntry.php
@@ -15,6 +15,7 @@ abstract class LiveEntry extends entry
 	
 	const CUSTOM_DATA_NAMESPACE_MEDIA_SERVERS = 'mediaServers';
 	const CUSTOM_DATA_RECORD_STATUS = 'record_status';
+	const CUSTOM_DATA_RECORD_OPTIONS = 'recording_options';
 	
 	static $kalturaLiveSourceTypes = array(EntrySourceType::LIVE_STREAM, EntrySourceType::LIVE_CHANNEL, EntrySourceType::LIVE_STREAM_ONTEXTDATA_CAPTIONS);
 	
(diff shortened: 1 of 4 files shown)


api_v3/lib/types/entry/KalturaLiveEntry.php
```diff
@@ -176,23 +175,57 @@ public function validateConversionProfile(entry $sourceObject = null)
 	 */
 	public function validateForUpdate($sourceObject, $propertiesToSkip = array())
 	{
-		/* @var $sourceObject LiveEntry */
-		if(isset($this->dvrStatus))
+		$stopedOnlyUpdateAttributes = array(
+				"dvrStatus" => array("validatePropertyChanged"), 
+				"dvrWindow" => array("validatePropertyChanged"), 
+				"recordStatus" => array("validatePropertyChanged","validateRecordedEntryId"), 
+				"conversionProfileId" => array("validatePropertyChanged","validateRecordedEntryId")
```
tan-tan-kanarek: What about recordingOptions?

api_v3/lib/types/entry/KalturaLiveEntry.php
```diff
@@ -176,23 +175,57 @@ public function validateConversionProfile(entry $sourceObject = null)
...
 		}
 		
-		if(isset($this->dvrWindow))
+		parent::validateForUpdate($sourceObject, $propertiesToSkip);
+	}
+	
+	protected function validatePropertyChanged($sourceObject, $attr)
+	{
+		/* @var $sourceObject LiveEntry */
+		$getter = "get" . ucfirst($attr);
```
tan-tan-kanarek: Please use map-between-objects to get the right getter.

alpha/lib/model/LiveEntry.php
```diff
@@ -121,7 +123,22 @@ public function generateFileName( $sub_type, $version = null)
...
 	
-	protected $decidingLiveProfile = false;
+	/**
+	 * Code to be run before updating the object in database
+	 * @param PropelPDO $con
+	 * @return bloolean
+	 */
+	public function preUpdate(PropelPDO $con = null)
+	{
+		if($this->isColumnModified(entryPeer::CONVERSION_PROFILE_ID))
```
tan-tan-kanarek: I think that if recording status or options changed, you also should clear the recorded and redirected ids.

api_v3/lib/types/entry/KalturaLiveEntry.php
yossipapi: recordingOptions is insertOnly so no need to handle this use case.

tan-tan-kanarek: It can't be insert-only anymore, if user can add recording he might need to set the options as well.

yossipapi: You are right ... will change to support this as well.

yossipapi:
@tan-tan-kanarek added support for clearing vod entry when recordingOptions has changed please view

PR #3956, opened 2016-03-14
fix all issues with anonymous answer cue points

noam-arad:


plugins/cue_points/base/lib/model/CuePointPeer.php
```diff
@@ -158,8 +148,11 @@ public static function filterSelectResults(&$selectResults, Criteria $criteria)
 	public static function retrieveByPK($pk, PropelPDO $con = null)
 	{
 		KalturaCriterion::disableTags(array(KalturaCriterion::TAG_USER_SESSION));
```
tan-tan-kanarek: You can disable and restore both TAG_USER_SESSION and TAG_WIDGET_SESSION in the same call to KalturaCriterion::disableTags.

PR #3974, opened 2016-03-16
validate recorded entry really exists before ending method execution

yossipapi:


api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -319,9 +319,12 @@ private function createRecordedEntry(LiveEntry $dbEntry, $mediaServerIndex)
...
 	    if(($dbEntry->getRecordStatus() != RecordStatus::PER_SESSION) && ($dbEntry->getRecordedEntryId())) {
-	    	$lock->unlock();
 	    	$recordedEntry = entryPeer::retrieveByPK($dbEntry->getRecordedEntryId());
-	    	return $recordedEntry;
+	    	if($recordedEntry)
+	    	{
+	    		$lock->unlock();
+	    		return $recordedEntry;
+	    	}	    	
 	    }
```
tan-tan-kanarek: Unlock should be done before the select in order to wait for entry creation by a different flavor.

yossipapi: The entire reason for this change is so we would validate if the recordedEntry really exists before freeing the lock and not continuing to the flow that re-creates the recordedEntry

PR #3977, opened 2016-03-17
utility to add partner additional parameters for mailing list

EladCohenKaltura:
NO-PLAT

### DIFF
--- a/alpha/scripts/utils/setPartnerAdditionalParameters.php
+++ b/alpha/scripts/utils/setPartnerAdditionalParameters.php
@@ -1,0 +1,35 @@
+<?php
+
+chdir(dirname(__FILE__));
+require_once(dirname(__FILE__) . '/../bootstrap.php');
+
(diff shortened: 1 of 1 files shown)


alpha/scripts/utils/setPartnerAdditionalParameters.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+chdir(dirname(__FILE__));
+require_once(dirname(__FILE__) . '/../bootstrap.php');
+
+if (count($argv) !== 3)
+{
+        die('please provide a partner id and a mailing list ' . PHP_EOL .
```
noam-arad: I think it would be clearer if you just wrote something like:
Usage: basename(__FILE__) <partner_id> <mailing-list>

tan-tan-kanarek:
Please rename the script name to something more meaningful.

PR #3995, opened 2016-03-21
When live entry server node status has changes we need to re-index li…

yossipapi:
…ve entry to update sphinx live status


alpha/lib/model/LiveEntryServerNode.php
```diff
@@ -3,13 +3,57 @@
...
 	const CUSTOM_DATA_DC = "dc";
+	
+	private $statusChanged = false;
+	
+	/* (non-PHPdoc)
+	 * @see BaseEntryServerNode::postSave()
+	 */
+	public function postSave(PropelPDO $con = null)
+	{
+		if($this->statusChanged)
```
tan-tan-kanarek: What's the purpose of the statusChanged flag? You can simply check the modified columns.
Also, if entry not found, log it as an error.
Also, use shorter texts in the logs.

PR #3988, opened 2016-03-20
fix category delete to be iterative and not recursive

gotlieb:
SUP-5851

@MosheMaorKaltura for review

PR #3997, opened 2016-03-21
doUpdate does not clear query cache so manually reload cuePoint befor…

yossipapi:
…e calling indexToSearchIndex


plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -697,6 +697,7 @@ public static function postProcessCuePoints( $liveEntry, $cuePointsIds = null )
 		foreach($cuePoints as $cuePoint)
 		{
 			/* @var $cuePoint CuePoint */
+			$cuePoint->reload();
```
tan-tan-kanarek: Instead of reloading them one by one from the DB, reload them all at once by PKs.
@gotlieb, you'll have the same issue in https://github.com/kaltura/server/pull/3988/files#diff-a9a336a911916afc23aa5058a30338bbR1815.

PR #4001, opened 2016-03-23
Wrap $event->consume with try and catch

Galmoss:
SUP-7683

### DIFF
--- a/alpha/apps/kaltura/lib/events/kEventsManager.php
+++ b/alpha/apps/kaltura/lib/events/kEventsManager.php
@@ -233,10 +233,15 @@ class kEventsManager
 		{
 			if (!class_exists($consumerClass))
 				continue;
-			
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/lib/events/kEventsManager.php
```diff
@@ -233,10 +233,15 @@ public static function raiseEvent(KalturaEvent $event)
...
-			
-			if($event->consume(new $consumerClass()) || !($event instanceof IKalturaCancelableEvent))
-				continue;
-				
+
+			try{
+				if($event->consume(new $consumerClass()) || !($event instanceof IKalturaCancelableEvent))
+					continue;
+			}
+			catch (CouchbaseException $e){
```
MosheMaorKaltura: CouchbaseException ?
looks like the wrong type.

tan-tan-kanarek: I agree, I think we should catch any Exception type here.

PR #4002, opened 2016-03-23
fix delete categories

gotlieb:
SUP-5851

@MosheMaorKaltura


alpha/lib/model/category.php
```diff
@@ -502,7 +502,7 @@ public function deleteChildCategories()
 
 		if ($this->getInheritanceType() == InheritanceType::MANUAL)
 			$this->addDeleteCategoryTreeKuserJob($this->getId(), false, $fullIds);
-		if ($this->move_entries_to_parent_category)
+		if ($this->move_entries_to_parent_category && $this->parent_id!=0)
```
tan-tan-kanarek: Please use getters, we try avoiding direct access to parent attributes.

PR #4018, opened 2016-03-28
add limit for admin page track entry list objects

yossipapi:


alpha/lib/model/TrackEntryPeer.php
```diff
@@ -10,6 +10,8 @@
  */ 
 class TrackEntryPeer extends BaseTrackEntryPeer implements IRelatedObjectPeer
 {
+	const PEER_GET_LIMIT = 30;
```
tan-tan-kanarek: I think the constant should be defined in EntryAdminService

yossipapi: thought about that at first as well but eventually moved it to the peer class... will move it back to the service again.

PR #4026, opened 2016-03-29
Wrap $event->consume with try and catch

Galmoss:
SUP-7683


Galmoss:
@MosheMaorKaltura  @tan-tan-kanarek updated pull request

PR #4034, opened 2016-03-29
support playerUrlType=embedIframeJs parameter to getFeed.php

MichalRadwantzor:


api_v3/web/getFeed.php
```diff
@@ -73,6 +73,7 @@ function getRequestParameter($paramName)
 $entryId = getRequestParameter('entryId');
 $limit = getRequestParameter('limit');
 $ks = getRequestParameter('ks');
+$playerUrlType = getRequestParameter('playerUrlType');
```
tan-tan-kanarek: Why don't we add this parameter to the feed instead of accepting it in the request?
This way we can define that the now feeds from now on will return the new player URL.

etameran: Additionally this should probably be the default from now on for every new feed.

MichalRadwantzor: ok, I opened this pull request instead: https://github.com/kaltura/server/pull/4036

MichalRadwantzor:
took a different approach for the fix, closing this one.

PR #4037, opened 2016-03-30
fix SUP-7792

asafrob:


asafrob:
HF on 11.11.0
adding to 11.12.0

PR #4040, opened 2016-04-03
allow listing entryServerNode without providing entry + add statusIn t…

yossipapi:
…o the filter


api_v3/lib/types/EntryServerNode/base/filters/KalturaEntryServerNodeFilter.php
```diff
@@ -21,12 +21,6 @@ protected function getCoreFilter()
 	 */
 	public function getListResponse(KalturaFilterPager $pager, KalturaDetachedResponseProfile $responseProfile = null)
 	{
-		if ($this->entryIdEqual == null &&
```
tan-tan-kanarek: I think this validation should remain for any non-internal partner.

yossipapi: Saw this after the commit ... I agree ... will add this now

yossipapi: I remembered there was a reason why I did not add this :)
This service is only open for partner -5 so there is no reason to add this validation.

PR #4048, opened 2016-04-06
use authenticated entryServerNode to know who is primary/backup

yossipapi:


alpha/lib/model/LiveEntry.php
```diff
@@ -936,4 +938,39 @@ public static function maxLiveEntryStatus($primaryMediaServerStatus, $secondaryM
 		else
 			return EntryServerNodeStatus::STOPPED;
 	}
+	
+	public function isStremaAlreadyBroadcasting()
```
kalturaguy: isStremaAlreadyBroadcasting  Stream not strema

PR #4062, opened 2016-04-11
Add logic to delete non-distributed entry distribution objects

hilak:


hilak:
@tan-tan-kanarek  @assafb  Please review.

plugins/content_distribution/lib/kContentDistributionFlowManager.php
```diff
@@ -1617,6 +1623,13 @@ public static function onEntryDeleted(entry $entry)
 				KalturaLog::log("Entry distribution [" . $entryDistribution->getId() . "] already flagged for deletion");
 				continue;
 			}
+			
+			if (!in_array ($entryDistribution->getStatus(), $validStatus))
+			{
+				$entryDistribution->setStatus(EntryDistributionStatus::DELETED);
+				$entryDistribution->setDirtyStatus(EntryDistributionDirtyStatus::NONE);
+				$entryDistribution->save ();
```
tan-tan-kanarek: Won't the submitDelete job fail for DELETED entry-distribution?

hilak: DELETED entry distributions are not returned because of the default criteria filter in EntryDistributionPeer class line 31.

tan-tan-kanarek: In that case you're breaking the current behavior, until now ready entry-distributions went through submitDelete and removed from the remote system. I think you should mark them as deleted only after the submitDelete finished (maybe even if they failed).

hilak: @tan-tan-kanarek  That is precisely why I added the 'validStatuses' array above and only delete the entry distribution if its status is not one of these.

tan-tan-kanarek: First I think that UPDATING should behave the same and you should also `continue;` after this save, because as you said, the submitDelete will fail anyway.

PR #4045, opened 2016-04-05
Kajam 11.12.0 plat 4645 new xslt

noam-arad:

PR #4070, opened 2016-04-12
Kajam 11.12.0 plat 4645 new xslt

noam-arad:


noam-arad:
see review remarks here:
https://github.com/kaltura/server/pull/4045

PR #4071, opened 2016-04-12
disregard capture_space for relevant clients

noam-arad:


configurations/generator.ini
```diff
@@ -38,6 +38,7 @@ subpackage = Client
 internal = true
 nopackage = true
 copyPath = batch/client
+excludepaths = plugins/capture_space
```
tan-tan-kanarek: Please move it to "all" section.

PR #4078, opened 2016-04-14
there is a synchronization issue if the convertprofile and the extrac…

noam-arad:
…tmedia both think the other isn't finished yet they hold this state in memory, to fix this the closer should also check weather the entry is ready.

SUP-7477


noam-arad:
see https://github.com/kaltura/server/pull/4079

api_v3/services/BatchService.php
```diff
@@ -560,4 +560,43 @@ function checkFileExistsAction($localPath, $size)
...
+	 * @return bool
+	 * @throws KalturaAPIException
+	 */
+	function checkEntryIsDone($batchJobId)
+	{
+		$ret_val = false;
+		$dbBatchJob = BatchJobPeer::retrieveByPK($batchJobId);
+		if (!$dbBatchJob)
+		{
+			throw new KalturaAPIException("batch job [$batchJobId] not found");//TODO: see if there is already a message defined if not define it.
```
tan-tan-kanarek: These are not the parameters that accepted by KalturaAPIException constructor.

deployment/permissions/service.batch.ini
```diff
@@ -160,3 +160,12 @@ permissionItem73.param4 =
...
 permissionItem73.permissions = BATCH_BASE, PARTNER_-1_GROUP_*_PERMISSION
+
+permissionItem84.service = batch
+permissionItem84.action = checkEntryIsDone
+permissionItem84.partnerId = -1
+permissionItem84.param3 =
+permissionItem84.param4 =
+permissionItem84.param5 =
+permissionItem84.tags =
+permissionItem84.permissions = BATCH_BASE, PARTNER_-1_GROUP_*_PERMISSION
```
tan-tan-kanarek: PARTNER_-1_GROUP_*_PERMISSION  is not needed here, BATCH_BASE is enough.

alpha/apps/kaltura/lib/batch2/bcdl/kBusinessPostConvertDL.php
```diff
@@ -342,7 +342,8 @@ public static function handleConvertFinished(BatchJob $dbBatchJob = null, flavor
 		
 		if(count($childJobs) > 1 && count($waitingFlavorAssets) < 1)
```
tan-tan-kanarek: In your use case, you won't get into this code because of the condition `count($waitingFlavorAssets) < 1`.
Since you check the flavors readyness in the closer, you can remove this entire code entirely.

PR #4079, opened 2016-04-14
Kajam 11.13.0 sup 7477 disney with locking

noam-arad:


noam-arad:
see https://github.com/kaltura/server/pull/4078

noam-arad:
Cannot use locking since requests might go to different DCs

PR #4096, opened 2016-04-21
trigger update user in db when changing only role

gotlieb:
PLAT-5354

@MosheMaorKaltura please review.

### DIFF
--- a/alpha/lib/model/kuser.php
+++ b/alpha/lib/model/kuser.php
@@ -86,7 +86,9 @@ class kuser extends Basekuser implements IIndexable, IRelatedObject
 					}
 				}
 			}
-			
+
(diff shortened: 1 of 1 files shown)


alpha/lib/model/kuser.php
```diff
@@ -86,7 +86,9 @@ public function preSave(PropelPDO $con = null)
 					}
 				}
 			}
-			
+
+			$this->setUpdatedAt(time());
```
yossipapi: move this after all the if validations have passed.
The next if statement may throw an exception so there is not point to set updated at before it passed.

gotlieb:
@yossipapi 

approve and ill merge it

PR #4099, opened 2016-04-24
Allow validate accessControl even if runing with AdminKS

yossipapi:


alpha/apps/kaltura/lib/kContextDataHelper.php
```diff
@@ -210,23 +205,22 @@ private function setContextDataFlavorAssets($flavorTags)
...
-				if($action->getType() == RuleActionType::LIMIT_FLAVORS)
-				{
-					/* @var $action kAccessControlLimitFlavorsAction */
-					$flavorParamsIds = explode(',', $action->getFlavorParamsIds());
-					$flavorParamsNotIn = $action->getIsBlockedList();
-				}
-			}	
-		}
+		
+		foreach ($this->contextDataResult->getActions() as $action)
```
MosheMaorKaltura: Must validate that contextDataResult is not null, in case of optimized_playback applyContext will not be called.

yossipapi: This is ok that in case of optimized_playback it wont run. if I remember correctly this was the entire point of this setup to avoid running ACL for calls made by the player but do run them for testme|clietnLibs|etc ...

Previously this was also the behavior if we were running with user KS and we had access control we would also have skipped the applyContext.

PR #4100, opened 2016-04-24
Kajam 11.13.0 sup 7911

noam-arad:


plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -257,7 +259,23 @@ protected function doSubmit(KalturaDistributionSubmitJobData $data, KalturaYoutu
...
+				$numOfTries = 0;
+				while (true) 
+				{
+					try
+					{
+						$ingestedVideo = $media->nextChunk($chunk);
+						break;
+					} catch (Google_IO_Exception $e)
+					{
+						KalturaLog::info("Uploading chunk to youtube failed with the message '".$e->getMessage()."' number of retries ".$numOfTries);
```
tan-tan-kanarek: Instead of implementing retry mechanism yourself, just throw kTemporaryException and the job will be retried.

MosheMaorKaltura: We don't want the Job to retry, but the current chunk upload to retry.

PR #4122, opened 2016-04-26
added node tests

gotlieb:
no-plat

### DIFF
--- a/generator/sources/node/tests/01.js
+++ b/generator/sources/node/tests/01.js
@@ -1,0 +1,33 @@
+var Unit = require('deadunit')
+var kc = require('../KalturaClient');
+var ktypes = require('../KalturaTypes');
+var vo = require ('../KalturaVO.js');
+var config = require ('./config.js');
+
+    var cb = function (results){
+	if(results){
+	    var test = Unit.test('Create -2 Admin session', function () {
+		this.count(1);
+		if(results.code && results.message){
+		    this.log(results.message);
+		    this.log(results.code);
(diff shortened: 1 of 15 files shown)


generator/sources/node/tests/config.js
```diff
@@ -0,0 +1,23 @@
+ var minus2_partner_id = -2;
```
tan-tan-kanarek: Please replace the secrets with tokens.

generator/sources/node/tests/01.js
```diff
@@ -0,0 +1,33 @@
+var Unit = require('deadunit')
```
tan-tan-kanarek: Please rename the numbered files to meaningful names.

generator/sources/node/tests/npm-debug.log
```diff
@@ -0,0 +1,202 @@
+0 info it worked if it ends with ok
```
tan-tan-kanarek: Please remove log file.

generator/sources/node/tests/testme.js.old
```diff
@@ -0,0 +1,46 @@
+var cb = function (results){
```
tan-tan-kanarek: Please remove this file.

generator/sources/node/tests/config.js
yossipapi: why not call this admin_console_partber_id? this would be the more appropriate name
Same comment for the secret params as well

PR #4147, opened 2016-04-27
update currentBroadcastStartTime when authentictae is called

yossipapi:


plugins/media/wowza/lib/model/WowzaMediaServerNode.php
```diff
@@ -134,8 +134,8 @@ public function getApplicationPrefix($mediaServerConfig)
 
 		$appPrefix = $this->getValueByField($mediaServerConfig, 'appPrefix', $appPrefix);
 		
-		if($this->getAppPrefix())
-			$appPrefix = $this->getApplicationPrefix();
+		if(is_null($this->getAppPrefix()))
+			$appPrefix = $this->getAppPrefix();
```
EladCohenKaltura: It seems that you will do this only in the case it is null - i think you want the opposite

yossipapi: You are right my bad... The not somehow disappeared... Will add it and update the pull request.

Thanks for the heads up.

alpha/lib/model/LiveEntryServerNode.php
```diff
@@ -44,9 +65,21 @@ public function postDelete(PropelPDO $con = null)
...
+			{
+				$liveEntry->setPrimaryServerNodeId(null);
+				
+				$entryServerNodes = EntryServerNodePeer::retrieveByEntryId($this->getEntryId());
+				if($liveEntry->getCurrentBroadcastStartTime())
+					$liveEntry->setCurrentBroadcastStartTime(0);
+			}
+			
+			if(!$liveEntry->save())
+				$liveEntry->indexToSearchIndex();
```
EladCohenKaltura: Save should generally work - I think that it is better to call $liveEntry->indexToSearchIndex(); within the if ($this->getServerType() === EntryServerNodeType::LIVE_PRIMARY) condition

yossipapi:
@EladCohenKaltura added the ! before the is_null as commented... did you review the entire code or only this section?

PR #4148, opened 2016-04-28
Kajam 11.3.0 tr 948

hilak:


alpha/apps/kaltura/lib/kCategoryEventHandler.php
```diff
@@ -0,0 +1,305 @@
...
+	}
+	
+	protected function addDeleteAggregationCategoryEntryJob (category $object, category $aggregationCategory)
+	{
+		$filter = new entryFilter();
+		$filter->set("_matchand_categories_ids", $object->getId().','.$aggregationCategory->getId());
+		$filter->set ("_notcontains_categories_ids", $aggregationCategory->getPublishingCategories());
+		
+		$additionalParameters = array();
+		$pair[self::PUBLIC_AGGREGATION_CATEGORY] = $aggregationCategory->getId();
```
rkreich: I think that you should initialize `$pair` first (as an array), otherwise it could throw a PHP notice

alpha/lib/model/category.php
```diff
@@ -49,6 +56,45 @@ class category extends Basecategory implements IIndexable, IRelatedObject
...
+	 * @return bool
+	 */
+	public function getIsAggregationCategory() {
+		return $this->getFromCustomData(self::IS_AGGREGATION_CATEGORY, null, false);
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getAggregationCategories() {
```
rkreich: It would be cleaner to return an array here instead of a string. If aggregation categories are not set, it should return an empty array and not null

alpha/lib/model/category.php
```diff
@@ -49,6 +56,45 @@ class category extends Basecategory implements IIndexable, IRelatedObject
...
+	 * @param bool $v
+	 */
+	public function setIsAggregationCategory($v) {
+		$this->putInCustomData(self::IS_AGGREGATION_CATEGORY, $v);
+	}
+
+	/**
+	 * @param string $value
+	 */
+	public function setAggregationCategories($value) {
```
rkreich: For `setAggregationCategories` & `setPublishingCategories` it would also make sense to pass an array instead of string.
It will make the code that is using it much cleaner and you won't need to duplicate the `implode(',', $xxx)`, `explode(',', $xxx)`

alpha/apps/kaltura/lib/kCategoryEventHandler.php
```diff
@@ -0,0 +1,305 @@
...
+			$aggregationCategory->setPublishingCategories(implode(',', $currentPublishingCategories));
+			$aggregationCategory->save();
+			
+			$this->addCopyJobToAggregationChannel ($object, $aggregationCategory->getId());
+		}
+	}
+	
+	protected function handleCategoryEntryCreated (categoryEntry $object)
+	{
+			$category = categoryPeer::retrieveByPK($object->getCategoryId());
```
rkreich: There is an extra tab indentation here

alpha/apps/kaltura/lib/kCategoryEventHandler.php
```diff
@@ -0,0 +1,305 @@
...
+			if (!$category)
+			{
+				KalturaLog::info("category [" . $object->getCategoryId() . "] does not exist in the system.");
+				return true;
```
rkreich: No need to return true if this method is not expected to return a boolean value. It can just be `return;`

alpha/apps/kaltura/lib/kCategoryEventHandler.php
```diff
@@ -0,0 +1,305 @@
...
+			if (!$category->getAggregationCategories())
+			{
+				KalturaLog::info("No aggregation categories found for category [" . $category->getId() . "]");
+				return true;
+			}
+			
+			$aggregationCategories = explode(',', $category->getAggregationCategories());
+			$aggregationCategoryEntries = categoryEntryPeer::retrieveActiveByEntryIdAndCategoryIds($object->getEntryId(), $aggregationCategories); 
+			
+			foreach ($aggregationCategoryEntries as $aggregationCategoryEntry)
```
rkreich: As discussed - let's add a comment here explaining the logic

api_v3/lib/types/KalturaCategory.php
```diff
@@ -506,6 +520,26 @@ private function validateCategory(category $sourceObject = null)
...
+			{
+				throw new KalturaAPIException(KalturaErrors::AGGREGATION_CATEGORY_WRONG_ASSOCIATION);
+			}
+			
+		if ($this->aggregationCategories)
+		{
+			$aggrCatIdsToCheck = explode (',' , $this->aggregationCategories);
+			foreach ($aggrCatIdsToCheck as $aggrCatIdToCheck)
+			{
+				$agrrCat = categoryPeer::retrieveByPK($aggrCatIdToCheck);
```
rkreich: If category not found the API will return `AGGREGATION_CATEGORY_WRONG_ASSOCIATION`? If so, we should add another validation and return `CATEGORY_NOT_FOUND`

batch/batches/Copy/Engines/KCopyingCategoryEntryEngine.php
```diff
@@ -0,0 +1,71 @@
...
+		foreach($categoryEntryList->objects as $categoryEntry)
+		{
+			$newCategoryEntry = $this->getNewObject($categoryEntry, $templateObject);
+			KBatchBase::$kClient->categoryEntry->add($newCategoryEntry);
+		}
+		
+		$results = KBatchBase::$kClient->doMultiRequest();
+		foreach($results as $index => $result)
+			if(is_array($result) && isset($result['code']))
+				unset($results[$index]);
```
rkreich: No logging of API errors?

batch/batches/Delete/Engines/KDeletingAggregationChannelEngine.php
```diff
@@ -0,0 +1,64 @@
...
+		KBatchBase::$kClient->startMultiRequest();
+		foreach($entriesList->objects as $entry)
+		{
+			/* @var $entry KalturaBaseEntry */
+			KBatchBase::$kClient->categoryEntry->delete($entry->id, $this->publicAggregationChannel);
+		}
+		$results = KBatchBase::$kClient->doMultiRequest();
+		foreach($results as $index => $result)
+			if(is_array($result) && isset($result['code']))
+				unset($results[$index]);
```
rkreich: Same comment about logging API errors

hilak: We don't log in any of the other engines either. Since this is a multi-request, there's no need to catch these errors - they will be displayed in regular batch logs.

PR #4153, opened 2016-05-01
Fix for BPM Signal Notifications

rkreich:
Discussed with @tan-tan-kanarek

### DIFF
--- a/plugins/event_notification/providers/bpm/lib/model/BusinessProcessSignalNotificationTemplate.php
+++ b/plugins/event_notification/providers/bpm/lib/model/BusinessProcessSignalNotificationTemplate.php
@@ -19,9 +19,44 @@ class BusinessProcessSignalNotificationTemplate extends BusinessProcessNotificat
 	 */
 	public function dispatch(kScope $scope)
 	{
-		return $this->dispatchPerCase($scope);
+		$jobData = $this->getJobData($scope);
+		/* @var $jobData kBusinessProcessNotificationDispatchJobData */
+		if(!$jobData->getObject())
+		{
+			return;
+		}
+
+		// check if event template is dispatching from a batch job event, if so, try to get the right case to signal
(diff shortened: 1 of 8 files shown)


plugins/event_notification/providers/bpm/lib/model/BusinessProcessSignalNotificationTemplate.php
```diff
@@ -19,9 +19,44 @@ public function __construct()
...
+		$jobId = null;
+		foreach($caseIds as $caseId)
+		{
+			// when we have $currentCaseId, we should only signal to that case
+			if (!is_null($currentCaseId) && $caseId !== $currentCaseId)
+				continue;
+
+			$currentJobData = clone $jobData;
+			$currentJobData->setCaseId($caseId);
+			$this->dispatchJob($scope, $currentJobData);
```
MosheMaorKaltura: Once case found - break the loop.

rkreich: @MosheMaorKaltura, the default behavior is actually to signal all cases (if $currentCaseId is null). So we should continue with the loop.
@tan-tan-kanarek, does it make sense to keep the code that will allow signaling all cases? Or should we always try to signal a specific case and throw an exception when case was not found on the event data?

tan-tan-kanarek: We might have a future use case to signal all cases, but I don't think it's very important for now as all our signaling will have a case id.

rkreich: OK. I'll leave the code as is, we might actually want to manually dispatch the template to signal all cases of a specific process.

PR #4159, opened 2016-05-03
Add new analytics service

eladBenedict:


api_v3/lib/types/report/KalturaAnalyticsFilter.php
```diff
@@ -9,13 +9,13 @@ class KalturaAnalyticsFilter extends KalturaObject
 	 * Query start time (in local time)
 	 * @var string
 	 */
-	public $from;
+	public $from_time;
```
tan-tan-kanarek: Please use camel case fromTime and toTime instead of underscores.

PR #4166, opened 2016-05-04
Added ability to set language on flavor asset object

NadavHarnik:
Kajam-11.14.0-PLAT5441 #time 3h


tan-tan-kanarek:
You should remove the language attribute from all classes that extend KalturaFlavorAsset and flavorAsset.

MosheMaorKaltura:
KalturaFlavorAsset  is extended by KalturaWidevineFlavorAsset  & KalturaLiveAsset.
KalturaLiveAsset make sense to have language attribute, while KalturaWidevineFlavorAsset  belongs to the old DRM mechanism (which is not really in use) , same for flavorAsset

PR #4182, opened 2016-05-09
Revert copy user id from template entry

Galmoss:
SUP-8057


Galmoss:
@tan-tan-kanarek  can you please confirm this is the change needed?

PR #4193, opened 2016-05-10
Kajam 11.15.0 plat 5476

NadavHarnik:


api_v3/lib/reflection/XmlClientGenerator.php
```diff
@@ -515,14 +515,20 @@ private function getServiceActionElement(KalturaActionReflector $actionReflector
...
+
 		$description = $actionInfo->description;
 		$description = $this->fixDescription($description);
 		$actionElement->setAttribute("description", kString::stripUtf8InvalidChars($description));
 		$actionElement->setAttribute("enableInMultiRequest", ($outputType === 'file' ? "0" : "1"));
-		
+
 		$actionElement->appendChild($resultElement);
+
+		if(!empty($actionInfo->actionAlias))
```
tan-tan-kanarek: Please add the alias as attribute of action element instead of child element.

PR #4195, opened 2016-05-10
Kajam 11.3.0 tr 948

hilak:


hilak:
Merging after code review, sanity and unit test

PR #4198, opened 2016-05-11
When bulkUploadXML engine is used to upload flavors with flavor param…

noam-arad:
…s that aren't in the conversion profile being used we create the flavor params and add the content, since these are not new flavor assets no "objectCreated" event is sent but objectDataChanged is raised. If the flavor params where in the conversion profile then we would have another job of ConvertProfile which will move the flavors to ready but since they aren't then the flavors remain in status QUEUED, so this fix should catch the objectDataChanged event and move the status of the flavors to "ready" (the entry will not be ready becasue the conversion profile will not be converted).

SUP-7695

### DIFF
--- a/alpha/apps/kaltura/lib/batch2/kFlowManager.php
+++ b/alpha/apps/kaltura/lib/batch2/kFlowManager.php
@@ -8,7 +8,7 @@
  * @subpackage Batch
  *
  */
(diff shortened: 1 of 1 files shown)


MosheMaorKaltura:
Reviewed

PR #4205, opened 2016-05-15
remove html tags from description

NadavHarnik:


plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -315,6 +315,11 @@ protected function doSubmit(KalturaDistributionSubmitJobData $data, KalturaYoutu
 		
 		return $distributionProfile->assumeSuccess;
 	}
+
+	private function sanitizeFromHtmlTags($filed)
+	{
```
MosheMaorKaltura: Since this function does not access an object (not access to this) make it static

PR #4247, opened 2016-05-23
Parsing dfxp captions correctly and not combining text word between e…

gotlieb:
…lements. This fix is to make sure sphinx index words correctly.

PLAT-5484

@MosheMaorKaltura


plugins/content/caption/base/lib/captionManagers/dfxpCaptionsContentManager.php
```diff
@@ -179,6 +179,9 @@ public function getContent($content)
 		try
 		{
 			$xml->loadXML($content);
```
tan-tan-kanarek: You call loadXML twice.

PR #4249, opened 2016-05-23
Support Live seamless failover when streamInfo is known

yossipapi:


alpha/lib/data/live/kLiveStreamParams.php
```diff
@@ -0,0 +1,85 @@
...
+
+	/**
+	 * @param int $height
+	 */
+	public function setHeight($height) { $this->height = $height; }
+
+	/**
+	 * @return the $codec
+	 */
+	public function getCode() { return $this->codec; }
```
EladCohenKaltura: missing c

yossipapi: Done

PR #4260, opened 2016-05-24
this is the minimum flow needed to support datachanged for videos in …

noam-arad:
…drop folder that are out of the conversion profile


MosheMaorKaltura:
Reviewed.

PR #4278, opened 2016-05-25
Kajam 11.15.0 plat 5476

NadavHarnik:


NadavHarnik:
After code review fixes.

PR #4280, opened 2016-05-25
Kajam 11.16.0 objc exclude services

MosheMaorKaltura:


generator/ObjCClientGenerator.php
```diff
@@ -954,7 +954,7 @@ function writePlugin($pluginName, $enumNodes, $classNodes, $serviceNodes, $servi
 		$services = array();
 		foreach($serviceNamesNodes as $serviceNode)
 		{
-			if($serviceNode->hasAttribute('plugin') && $pluginName == '')
+			if($serviceNode->hasAttribute('plugin') && $pluginName == '' || !$this->shouldIncludeService(strtolower($serviceNode->getAttribute("name"))) )
```
tan-tan-kanarek: Use service id instead of name, it may matter.

tan-tan-kanarek: Also, strtolower is not needed on the id.

PR #4282, opened 2016-05-25
Add manifest merging support for old Wowza partners

yossipapi:


alpha/lib/model/DeliveryProfileLiveAppleHttp.php
```diff
@@ -120,13 +120,61 @@ protected function isDvrContent($content) {
...
+				
+			$manifestLine = next($manifestLines);
+		}
+	}
+	
+	/**
+	 * Build the manifest from the rreported stream information and build all flavors array
+	 * @param string $url
+	 */
+	private function buildStramInfoFlavorsArray($url, array &$flavors, array $kLiveStreamParamsArray, $flavorBitrateInfo, $doaminPrefix)
```
EladCohenKaltura: Stream instead of Stram

yossipapi: Done.

alpha/lib/model/DeliveryProfileLiveAppleHttp.php
```diff
@@ -120,13 +120,61 @@ protected function isDvrContent($content) {
...
-	private function buildM3u8Flavors($url, array &$flavors, array $kLiveStreamParamsArray, $flavorBitrateInfo = array())
+	private function buildProxiedFlavorsArray($url, array &$flavors, $doaminPrefix)
 	{
-		$domainPrefix = $this->getDeliveryServerNodeUrl(true);
+		if($this->getDisableExtraAttributes())
+			$url = kDeliveryUtils::addQueryParameter($url, "attributes=off");
+		
+		$manifest = KCurlWrapper::getContent($url);
+		if(!$manifest)
+			return;
```
EladCohenKaltura: Don't you want to log that you did not get a manifest for this URL ?

alpha/lib/model/DeliveryProfileLiveAppleHttp.php
```diff
@@ -120,13 +120,61 @@ protected function isDvrContent($content) {
...
+								
+						case 'RESOLUTION':
+							list($flavor['width'], $flavor['height']) = explode('x', $attributeParts[1], 2);
+							break;
+					}
+				}
+				$flavors[] = $flavor;
+			}
+				
+			$manifestLine = next($manifestLines);
```
EladCohenKaltura: You are using the next twice in the same while - this can cause you to skip a line after the '#EXT-X-STREAM-INF' line

alpha/lib/model/DeliveryProfileLiveAppleHttp.php
yossipapi: yes maybe that would be a good idea foratter identifying issues that occurred in the past.

I will add the debug.

alpha/lib/model/DeliveryProfileLiveAppleHttp.php
yossipapi: The parsing is according to HLS and Wowoza standards and it worked for us on production from the beginning so I don't think changing it would be the right thing to do.

EladCohenKaltura: is this code copied from somewhere ?

yossipapi: Yes this is the code that was previously here for merging manifest and supporting seamless failover.

PR #4287, opened 2016-05-26
Update categoryUser table when update User

david-winder-kaltura:
PLAT4661 #time 1d

### DIFF
--- a/alpha/lib/model/categoryKuser.php
+++ b/alpha/lib/model/categoryKuser.php
@@ -43,7 +43,14 @@ class categoryKuser extends BasecategoryKuser implements IIndexable{
 	{
 		$this->setUpdateMethod(UpdateMethodType::MANUAL);
 	}
-	
+
+	public function updateUser($newPuserId = null, $newScreenName = null) {
+		if ($newPuserId)
+			parent::setPuserId($newPuserId);
+		if ($newScreenName)
+			parent::setScreenName($newScreenName);
+	}
+
 	public function setPuserId($puserId)
 	{
 		if ( $this->getPuserId() == $puserId )  // same value - don't set for nothing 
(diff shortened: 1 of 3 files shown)


alpha/lib/model/categoryKuser.php
```diff
@@ -43,7 +43,14 @@ public function applyDefaultValues()
 	{
 		$this->setUpdateMethod(UpdateMethodType::MANUAL);
 	}
-	
+
+	public function updateUser($newPuserId = null, $newScreenName = null) {
```
MosheMaorKaltura: Rename to updateKuser

MosheMaorKaltura: remove the new* from the var names

alpha/lib/model/categoryKuser.php
```diff
@@ -43,7 +43,14 @@ public function applyDefaultValues()
 	{
 		$this->setUpdateMethod(UpdateMethodType::MANUAL);
 	}
-	
+
+	public function updateKuser($PuserId = null, $ScreenName = null) {
```
MosheMaorKaltura: Coding conversion  PuserId & ScreenName should be lower cammel case

api_v3/services/UserService.php
```diff
@@ -150,6 +166,11 @@ public function updateAction($userId, KalturaUser $user)
 			}			
 			$dbUser = $user->toUpdatableObject($dbUser);
 			$dbUser->save();
+			if (($user->id || $user->screenName) &&
+				categoryKuserPeer::isCategroyKuserExistsForKuser($dbUser->getId()))
+			{
+				$this->updateCategoryKuser($dbUser->getId(),$user);
```
tan-tan-kanarek: That should be triggered from the kuser postUpdate and not from the service, a user id or screen name could be changed from PS2, internal scripts or maintenance scripts.

PR #4289, opened 2016-05-26
Validate filter is not empty and contains the mandatory parameters

MosheMaorKaltura:


plugins/like/services/LikeService.php
```diff
@@ -129,16 +129,14 @@ public function checkLikeExistsAction ( $entryId , $userId = null )
...
-		if(!$filter)
-			$filter = new KalturaLikeFilter();
-		else	
-		{			
-			if($filter->entryIdEqual && !entryPeer::retrieveByPK($filter->entryIdEqual))			
-				throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $filter->entryIdEqual);			
-			if($filter->userIdEqual && !kuserPeer::getActiveKuserByPartnerAndUid(kCurrentContext::$ks_partner_id, $filter->userIdEqual))
-				throw new KalturaAPIException(KalturaErrors::USER_NOT_FOUND);			
-		}
+		if( !$filter->entryIdEqual && !$filter->userIdEqual && !$filter->entryIdIn)
```
shpiNi: @MosheMaorKaltura need to check if there's a filter. if not, you get an exception for checking a field of a non-existing object.

PR #4296, opened 2016-05-29
add exceptions needed for html purifier

noam-arad:
PLAT-5474


MosheMaorKaltura:
Also,
Enable the purifier.

PR #4299, opened 2016-05-29
Kajam 11.16.0 export thumbs plat 2325

noam-arad:


alpha/apps/kaltura/lib/storage/kStorageExporter.php
```diff
@@ -62,11 +62,35 @@ public function objectChanged(BaseObject $object, array $modifiedColumns)
...
+		// if changed object is thumb asset
+		if( $object instanceof thumbAsset && PermissionPeer::isValidForPartner(PermissionName::FEATURE_REMOTE_STORAGE, $object->getPartnerId()) && $object->isLocalReadyStatus())
+			return true;
+		
+	}
+
+	public function objectAdded(BaseObject $object, BatchJob $raisedJob = null)
+	{
+		$entry = $object->getentry();
+
```
MosheMaorKaltura: Should limit this event handler explicitly to the object it should handle

yossipapi:
@noam-arad  What about adding he export method to the service itself?
Also was it verified that when calling thumbAsset->getUrl() for exported content that the actual URL returned is one pointing to the remote storage?

noam-arad:
Hi @yossipapi , you're right.
It would have exported the thumb via "baseEntry->export" but would return an error, please see https://github.com/kaltura/server/pull/4308 with the fix.

Regarding thumbAsset->getUrl() - works out of the box

PR #4313, opened 2016-05-31
added missing parentheses

gotlieb:
PLAT-5488


tan-tan-kanarek:
Please delete merged branches.

PR #4316, opened 2016-05-31
change email body to send origin entry and not temporary entry

david-winder-kaltura:
SUP8208 #time 2d


tan-tan-kanarek:
We can't do that, this event-notification infrastructure is used also for BPM, HTTP and Push notifications that might want to react to temporary entry events.
You can add a configuration to the notification template to decide if this transformation should be applied.
Also, I recommend to check if the object is instance of entry and then use the entry getters and not access directly the custom-data methods that are not implemented in all propel objects.

david-winder-kaltura:
Moving the email template logic to eventNotificationTemplate

PR #4334, opened 2016-06-06
reduce logger memory usage

erankor:
backtrace is only used to get the calling function name, no need for the args / objects

### DIFF
--- a/infra/log/KalturaLog.php
+++ b/infra/log/KalturaLog.php
@@ -189,7 +189,7 @@ class LogMethod
 	public function __toString()
 	{
 		$backtraceIndex = 3;
-		$backtrace = debug_backtrace();
+		$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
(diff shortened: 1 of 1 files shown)


infra/log/KalturaLog.php
```diff
@@ -189,7 +189,7 @@ class LogMethod
 	public function __toString()
 	{
 		$backtraceIndex = 3;
-		$backtrace = debug_backtrace();
+		$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
```
noam-arad: Since this is relevant for php 5.3.6 and up and on CentOs the version (unless compiled manually) is 5.3.3 I think you should change to 
$backtrace = @debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
This way the log file won't be full of NOTICES

erankor: Already merged by @yossipapi

PR #4347, opened 2016-06-06
Support manifest merging for kalturaLive streams + audioOnly support

yossipapi:


alpha/lib/model/DeliveryProfileLiveAppleHttp.php
```diff
@@ -324,5 +256,26 @@ private function buildFlavorBitrateInfoArray($primaryServerStreams)
...
 		return $flavorBitrateInfo;
 	}
+	
+	private function isAudioStream($primaryServerStreams, $backupServerStreams)
+	{
+		if(count($primaryServerStreams))
+		{
+			/* @var $kLiveStreamParams kLiveStreamParams */
+			$kLiveStreamParams = reset($primaryServerStreams);
+			if(!$kLiveStreamParams->getHeight() && !$kLiveStreamParams->getWidth())
```
EladCohenKaltura: in this case 0 is not a bool - should not be handled as such 
would use getHeight == 0 && getWidth ==  0

yossipapi: In case of audio stream the width and height will a;ways be undefined so no case to check for cases where value is 0.

The use case of streaming different streams to primary and backup is not a normal situation and is not supported

PR #4351, opened 2016-06-07
Added schedule events consumer

gotlieb:
PLAT-5488
PLAT-5484

### DIFF
--- a/plugins/schedule/base/lib/kScheduleEventsConsumer.php
+++ b/plugins/schedule/base/lib/kScheduleEventsConsumer.php
@@ -1,0 +1,108 @@
+<?php
+class kScheduleEventsConsumer implements kObjectChangedEventConsumer, kObjectDeletedEventConsumer, kObjectCreatedEventConsumer
+{
+    public function shouldConsumeCreatedEvent(BaseObject $object)
+    {
+        if ($object instanceof categoryEntry)
+        {
+            return true;
+        }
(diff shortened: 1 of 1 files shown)


plugins/schedule/base/lib/kScheduleEventsConsumer.php
```diff
@@ -0,0 +1,108 @@
+<?php
+class kScheduleEventsConsumer implements kObjectChangedEventConsumer, kObjectDeletedEventConsumer, kObjectCreatedEventConsumer
+{
+    public function shouldConsumeCreatedEvent(BaseObject $object)
+    {
+        if ($object instanceof categoryEntry)
```
tan-tan-kanarek: Please check the status, if it's not active you can ignore it.

plugins/schedule/base/lib/kScheduleEventsConsumer.php
```diff
@@ -0,0 +1,108 @@
...
+
+        return false;
+    }
+
+    /* (non-PHPdoc)
+     * @see kObjectChangedEventConsumer::objectChanged()
+     */
+    public function objectChanged(BaseObject $object, array $modifiedColumns)
+    {
+        try
```
tan-tan-kanarek: No need to wrap with try and catch because the events manager already does it.
No need to check the object type because the should method already does it.

PR #4353, opened 2016-06-07
modifications in kscheduleEventConsumer - remove try/catch and handle…

gotlieb:
… events. Also modifications in templateEntry sphinx index.

PLAT-5488


plugins/schedule/base/lib/model/ScheduleEvent.php
```diff
@@ -270,21 +270,22 @@ public function getTemplateEntryCategoriesIdsForIndex()
 		return '';
 	}
 
-	public function getTemplateEntryIdForIndex()
```
tan-tan-kanarek: You should implement empty getTemplateEntryId instead.

PR #4355, opened 2016-06-07
when checking if entry is done return true is entry status is already…

yossipapi:
… set to READY


yossipapi:
@noam-arad please review.

noam-arad:
Looks good

MosheMaorKaltura:
The action ‘checkEntryIsDoneAction’ is not only checking but also calling kBusinessPostConvertDL::handleConvertFinished.
Did you verify this function will be called in the flow that you covered also when entry becomes ready.

noam-arad:
handleConvertFinished should be called only if the entry is stuck in a non-ready state. Thus if the entry is ready there is no need for handleConvertFinished

PR #4356, opened 2016-06-07
when exporting replacing entry the remoteStorage rules should be vali…

yossipapi:
…dated on the replaced entry


MosheMaorKaltura:
reviewed

PR #4352, opened 2016-06-07
Kajam 11.17.0 export images

noam-arad:

PR #4363, opened 2016-06-08
Kajam 11.17.0 export images plat 3849

noam-arad:


noam-arad:
already discussed in https://github.com/kaltura/server/pull/4352

PR #4367, opened 2016-06-08
add delete direct member only to filter in enable inherit

david-winder-kaltura:
PLAT-5433 #time 1.5d


MosheMaorKaltura:
Reviewed

PR #4394, opened 2016-06-13
Update kaltura_sphinx.template

DBezemer:
Correct service name

### DIFF
--- a/configurations/logrotate/kaltura_populate.template
+++ b/configurations/logrotate/kaltura_populate.template
@@ -8,7 +8,7 @@
  notifempty
  sharedscripts
  postrotate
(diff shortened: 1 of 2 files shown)


DBezemer:
Please merge

DBezemer:
@yossipapi all of the init scripts are renamed about 2 years ago, currently not having this change breaks all installs out there. If you are updating from git in your installation it should be sufficient to update the symlink for kaltura_populate

jessp01:
This was since pushed here:
d4af644e584d67ec15b2d3c9408c2c988bb7b8d0
d4af644e584d67ec15b2d3c9408c2c988bb7b8d0

PR #4403, opened 2016-06-15
TR-1144 - Raise kObjectDataChangedEvent

rkreich:
- Raise kObjectDataChangedEvent when asset data is updated in Cielo24 and VoiceBase integration providers

**Opening for review only, will be merged to 11.18.0**


rkreich:
@tan-tan-kanarek, can you give it a look?
`kObjectDataChangedEvent` is raised in the API when updating asset's content (https://github.com/kaltura/server/blob/Kajam-11.17.0/plugins/content/attachment/services/AttachmentAssetService.php#L127)
Is it OK to raise it here (from the flow manager)?

(The exta diff is normalization of incorrect line endings)

tan-tan-kanarek:
I would prefer if we could have it raised once in the core, maybe in postUpdate and only if version was changed. But I can live with it as is.

PR #4405, opened 2016-06-15
fixed objc client generation - cause by duplicate templateEntryIdEqual param in scheduling filters.

gotlieb:
Kajam 11.17.0 plat 5650
@MosheMaorKaltura


plugins/schedule/base/lib/api/filters/base/KalturaEntryScheduleEventBaseFilter.php
```diff
@@ -8,7 +8,6 @@
 {
 	static private $map_between_objects = array
 	(
-		"templateEntryIdEqual" => "_eq_template_entry_id",
```
tan-tan-kanarek: This is generated code and shouldn't be changed manually.
If you have duplication, remove the other occurrence.

PR #4410, opened 2016-06-16
changed audio_only tag to alt_audio on playManifest

NadavHarnik:
SUP-8352 #time 1h


MosheMaorKaltura:
@erankor  please validate it since we may want to upload in the current release.

PR #4434, opened 2016-06-21
PLAT-5714

EladCohenKaltura:
PLAT-5714
change to support always free trailers - trailer files will always be set with defualt PPVmodule of none - free


MosheMaorKaltura:
Merge afters testing conducted by PS OTT + validation of product manager

PR #4440, opened 2016-06-23
Move generator to different repo

tan-tan-kanarek:


tan-tan-kanarek:
I cannot maintain two repos.

PR #4441, opened 2016-06-23
Correctly set default

DBezemer:
and not debug value for CURLOPT_SSL_VERIFYHOST


MosheMaorKaltura:
reviewed

DBezemer:
@MosheMaorKaltura after reviewing is there any reason not to merge? if so let me know.

PR #4443, opened 2016-06-24
Kajam 11.18.0 tr 1029

hilak:


plugins/content/attachment/services/AttachmentAssetService.php
```diff
@@ -494,11 +494,11 @@ public function getAction($attachmentAssetId)
 	 * List attachment Assets by filter and pager
 	 * 
 	 * @action list
-	 * @param KalturaAssetFilter $filter
+	 * @param KalturaAttachmentAssetFilter $filter
 	 * @param KalturaFilterPager $pager
 	 * @return KalturaAttachmentAssetListResponse
 	 */
-	function listAction(KalturaAssetFilter $filter = null, KalturaFilterPager $pager = null)
+	function listAction(KalturaAttachmentAssetFilter $filter = null, KalturaFilterPager $pager = null)
```
tan-tan-kanarek: It breaks compatibility.

plugins/content/attachment/lib/api/filters/KalturaAttachmentAssetFilter.php
```diff
@@ -4,7 +4,33 @@
  * @subpackage api.filters
  */
 class KalturaAttachmentAssetFilter extends KalturaAttachmentAssetBaseFilter
-{
+{	
+	/**
+	 * @dynamicType KalturaAssetType
+	 * @var string
+	 */
+	public $typeNotIn;
```
rkreich: It's a bit problematic to put `typeNotIn` on `KalturaAttachmentAssetFilter`, type is a property of Asset.
That's why we wanted to add a new type to allow filtering by "not in".

plugins/content/attachment/services/AttachmentAssetService.php
hilak: There's no compatibility today. If you pass anything other than the KalturaAttachmentFilter, and error is thrown. Line 509 attempts to cast a filter which is not of type KalturaAttachmentAssetFilter into the type. This will throw an error with any class but KalturaAttachmentAssetFilter  and KalturaTranscriptAssetFilter

tan-tan-kanarek: That line exists because compat-check in the past showed that there is a customer that uses KalturaAssetFilter as input. If you'll apply your change you will break the application of that customer, I can't remember who it was.

PR #4458, opened 2016-06-30
Update function variable typedef

DBezemer:
PHP Warning:  Declaration of BulkUploadEngineXml::updateObjectsResults($requestResults, $bulkUploadResults) should be compatible with KBulkUploadEngine::updateObjectsResults(array $requestResults, array $bulkUploadResults) in /opt/kaltura/app/plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php


DBezemer:
Any progress on this?

PR #4461, opened 2016-07-04
SUP-8526 changed float formatting to avoid OS locale settings affect …

gonenradai:
…the number representation


gonenradai:
zip file in the respective support ticket can be used to reproduce the issue (just need to make sure Windows locale setting use comma as decimal separator instead of dot).

tan-tan-kanarek:
This client generator is not used anymore.
Please apply the fix in https://github.com/kaltura/clients-generator

gonenradai:
@tan-tan-kanarek , @MosheMaorKaltura  
https://github.com/kaltura/clients-generator doesn't have a 11.19 branch yet...

tan-tan-kanarek:
Created

PR #4468, opened 2016-07-06
In case entry with parent ID set , set display_in_search to system

MosheMaorKaltura:
SUP-8124#time 1d


tan-tan-kanarek:
I think it's wrong to change the display-in-search to existing entries.

MosheMaorKaltura:
1. In case of template entry, the entry is first created and then updated.
2. Once entry is set to display in search = system , it will never return back.

PR #4476, opened 2016-07-07
Kajam 11.19.0 plat 5414

shpiNi:


plugins/content_distribution/providers/pushToNews/lib/PushToNewsDistributionEngine.php
```diff
@@ -0,0 +1,184 @@
...
+	{
+		$this->validateJobObjects($data);
+
+		$distributionProfile = $data->distributionProfile;
+		$host = $distributionProfile->host;
+		$port = $distributionProfile->port;
+		$password = $distributionProfile->password;
+		$basePath = $distributionProfile->basePath;
+		
+		$url = $host;
```
rkreich: Need to move this to a helper function that will receive `$distributionProfile` as a parameter and return the URL.
Same goes for https://github.com/kaltura/server/pull/4476/files#diff-89ba83b8ce3c12e749ec8357f423036dR74

PR #4488, opened 2016-07-12
adding function for constant value increment

shpiNi:


MosheMaorKaltura:
Why not just use incrementSubmitDataVersion ?

PR #4489, opened 2016-07-12
Add KalturaSystemPartnerPackage to adminConsole

MosheMaorKaltura:


tan-tan-kanarek:
That change is redundant, once the plugin enable, the service will be  included and the used objects will be generated.

MosheMaorKaltura:
It was reverted.

PR #4490, opened 2016-07-12
add include o admin console client

MosheMaorKaltura:


tan-tan-kanarek:
1. This code shouldn't be changed, it should be deleted, see https://github.com/kaltura/server/pull/4462.
2. This code will cause that only this service will be generated.
3. To include this service, simply enable the plugin.

PR #4501, opened 2016-07-13
add toObject to KalturaBulkUploadJobData to avoid passing null  to ba…

yossipapi:
…se class


tan-tan-kanarek:
KalturaBulkUploadJobData should never be used as well as kBulkUploadJobData, the reason they're not abstract is for backward compatibility. If someone used this class, he should get an exception.

MosheMaorKaltura:
Class KalturaBulkUploadJobData is being instantiated by batch

MosheMaorKaltura:
Reviewed. Let's push it into  Kajam-11.21.0

PR #4502, opened 2016-07-13
fix Memory Fatal error caused when entry is associated with large amo…

yossipapi:
…unt of cue-points


plugins/cue_points/base/CuePointPlugin.php
```diff
@@ -277,37 +279,39 @@ public static function getCuePointSearchData(entry $entry)
...
+			}
 			
-			$dataByType[$cuePointType][] = $contributedData;
-		}
-		
-		$data = array();
-		foreach ($dataByType as $type => $typeData)
-		{
-			$data = array_merge($data, $typeData);
+			$offset += self::CUE_POINT_FETCH_LIMIT;
```
MosheMaorKaltura: Offset should increase in the number of objects that returned in the response.

PR #4511, opened 2016-07-19
Increase php memorty limit for report with large result set

olampert:


MosheMaorKaltura:
Current max allowed size is 128MB.
You are trying to request 1024MB, was it evaluated ?

olampert:
Changed to 256MB after testing some reports

PR #4515, opened 2016-07-19
support disabling the inheritance of security

erankor:
settings from parent to child


alpha/lib/enums/PermissionName.php
```diff
@@ -61,6 +61,7 @@
 	const FEATURE_IGNORE_ENTRY_SEO_LINKS = 'FEATURE_IGNORE_ENTRY_SEO_LINKS';
 	
 	const FEATURE_DISABLE_CATEGORY_LIMIT = 'FEATURE_DISABLE_CATEGORY_LIMIT';
+	const FEATURE_DISABLE_PARENT_ENTRY_SECURITY_INHERITANCE = 'FEATURE_DISABLE_PARENT_ENTRY_SECURITY_INHERITANCE';
```
tan-tan-kanarek: Where is this constant used? Why is it here?

tan-tan-kanarek:
Generally this change will break production entries that already counting on the parent-id, a migration script is needed to move to this code.

alpha/lib/enums/PermissionName.php
erankor: used below in this pull, in the new entry::getSecurityParentId function

erankor:
I don't see why, it will only affect partners for which FEATURE_DISABLE_PARENT_ENTRY_SECURITY_INHERITANCE will be enabled. 
The reason for this change was that in MediaPrep, a parent-child relationship was used to hold the relation between a movie and its trailers. The problem was that the trailers need to have a non-restricted access control, while the main video has a DRM access control. This change makes it possible to use two different access controls for the parent and child.

PR #4544, opened 2016-07-27
ensure acl for remote assets starts with slash

etameran:


rkreich:
Compat check passed successfully

PR #4561, opened 2016-08-01
Kajam 11.20.0 tr 1029

hilak:


api_v3/lib/types/conversionProfile/filters/KalturaAssetFilter.php
```diff
@@ -56,6 +72,17 @@ protected function doGetListResponse(KalturaFilterPager $pager, array $types = n
 		$c = new Criteria();
 		$flavorAssetFilter->attachToCriteria($c);
 		
+		if ($flavorAssetFilter->get('_in_type'))
+        {
```
rkreich: @hilak, need to fix indents here

PR #4569, opened 2016-08-03
retrieve parents resources id's in case of a recurrences event retrieval

gotlieb:
PLAT-5878

@MosheMaorKaltura


plugins/schedule/base/lib/iCal/kSchedulingICalEvent.php
```diff
@@ -249,7 +257,6 @@ public static function fromObject(KalturaScheduleEvent $event)
 		}
 
 		$resources = ScheduleEventResourcePeer::retrieveByEventId($event->id);
-		$resourceIds = array();
```
tan-tan-kanarek: I think that child resources should replace the parent resources if there are any, not to be added to the parent resources, but it should be decided by Renan.

PR #4580, opened 2016-08-07
Kajam 11.21.0 play server flavot get permission

EladCohenKaltura:


api_v3/services/FlavorAssetService.php
```diff
@@ -22,7 +22,8 @@ protected function kalturaNetworkAllowed($actionName)
 			$actionName == 'getFlavorAssetsWithParams' ||
 			$actionName == 'convert' ||
 			$actionName == 'reconvert' ||
-			$actionName == 'setContent'
+			$actionName == 'setContent' ||
```
noam-arad: don't you want to add permission to partner id -6 for both these actions?

EladCohenKaltura: no - currently only the get is additionally needed

noam-arad:
Approved

PR #4586, opened 2016-08-07
moving handling cuepoint filter typeIn and typeEqual to the sphinx applyingFilters part PLAT-5742

gotlieb:
@noam-arad


noam-arad:
Approved

PR #4596, opened 2016-08-09
restore commented code from schudle consumer thast was commented out …

gotlieb:
…to allow sphinx indexing on previous deployoment.

PLAT-5673

@noam-arad


tan-tan-kanarek:
Please add to deployment instructions a guideline to run populate script after code deployment.

gotlieb:
no need to run the populate. currently there are 12 events in sphinx in production and it is meaningful according to Eran E

PR #4611, opened 2016-08-15
KalturaBulkUploadJobData is abstract and cannot be initiated

yossipapi:


yossipapi:
See old pull request for comments and reviews.

https://github.com/kaltura/server/pull/4501

PR #4618, opened 2016-08-16
fixed remove entry from remote distribution

NadavHarnik:
PLAT-5761 #time 2.5d


yossipapi:
@NadavHarnik looks OK.

PR #4619, opened 2016-08-17
handle the case where playManifest request has no flavors

NadavHarnik:
PLAT-5950 #time 4h

### DIFF
--- a/alpha/lib/model/DeliveryProfileDynamicAttributes.php
+++ b/alpha/lib/model/DeliveryProfileDynamicAttributes.php
@@ -485,11 +485,14 @@ class DeliveryProfileDynamicAttributes {
 		return array();
 	}
 
-	private function addAltAudioFlavors(&$flavors, &$curFlavors)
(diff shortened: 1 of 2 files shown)


alpha/lib/model/DeliveryProfileDynamicAttributes.php
```diff
@@ -477,11 +477,23 @@ public function filterFlavorsByTags($flavors)
...
+			}
 		}
 		return array();
 	}
 
+	private function addAltAudioFlavors(&$flavors, &$curFlavors)
+	{
+		foreach($flavors as $flavor)
+		{
+			if(!in_array($flavor, $curFlavors) && $flavor->hasTag(assetParams::TAG_ALT_AUDIO) && playManifestAction::shouldAddAltAudioFlavors($this->getFormat()))
```
yossipapi: check "playManifestAction::shouldAddAltAudioFlavors($this->getFormat()" once before the loop

PR #4627, opened 2016-08-19
Update kBusinessConvertDL.php

EladCohenKaltura:
Disabled Filler logic on ad stitching transcoding command


MosheMaorKaltura:
This is a temporary delivered code so we can move on with ad stitch tests.
Code will not be used by anyone other than playserver

PR #4636, opened 2016-08-25
make accessControlId updatebale using user KS

Galmoss:
SUP-9003


MosheMaorKaltura:
Did you validated the permission that are set for the specific action, so it will not allow widget KS to call this action.?

PR #4646, opened 2016-08-29
support play-server in delivery profile

NadavHarnik:
PLAT-6012 #time 1d

### DIFF
--- a/alpha/lib/model/DeliveryProfileVod.php
+++ b/alpha/lib/model/DeliveryProfileVod.php
@@ -13,6 +13,22 @@ abstract class DeliveryProfileVod extends DeliveryProfile {
 	/** -------------------
 	 * Functionality 
 	 * --------------------*/
+
+	protected function getPlayServerUrlPrefix()
+	{
+		return '';
+	}
+
+	protected  function generatePlayServerUrlPrefix()
+	{
+		$prefix = '/usePlayServer/1';
+		if($this->getDynamicAttributes()->getUiConfId())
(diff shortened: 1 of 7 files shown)


alpha/lib/model/DeliveryProfileVod.php
```diff
@@ -13,6 +13,22 @@
...
+		return '';
+	}
+
+	protected  function generatePlayServerUrlPrefix()
+	{
+		$prefix = '/usePlayServer/1';
+		if($this->getDynamicAttributes()->getUiConfId())
+			$prefix .= '/uiConfId/'.$this->getDynamicAttributes()->getUiConfId();
+		if($this->getDynamicAttributes()->getSessionID())
+			$prefix .= '/sessionId/'.$this->getDynamicAttributes()->getSessionID();
```
MosheMaorKaltura: We need to consider to generate the session ID here , rather generate it on any request and use it only here. due to performance considerations.

alpha/lib/model/DeliveryProfileVod.php
```diff
@@ -23,8 +39,12 @@ protected function getBaseUrl(asset $flavorAsset) {
 		$partnerId = $flavorAsset->getPartnerId();
 		$subpId = $entry->getSubpId();
 		$partnerPath = myPartnerUtils::getUrlForPartner($partnerId, $subpId);
-		
-		$url = "$partnerPath/serveFlavor/entryId/".$entry->getId();
+
+		$url = '';
```
MosheMaorKaltura: to make the code thinner , please use the binary operator ?

PR #4659, opened 2016-09-01
if thumbnail addResource fails the entry as a whole is successfully u…

noam-arad:
…ploaded so we shouldn't fail the job, instead in the log we will mention this issue

### DIFF
--- a/plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
+++ b/plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
@@ -614,10 +614,10 @@ class BulkUploadEngineXml extends KBulkUploadEngine
 		switch($contentAssetsAction)
 		{
 			case self::$actionsMap[KalturaBulkUploadAction::UPDATE]:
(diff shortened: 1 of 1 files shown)


plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
```diff
@@ -718,6 +719,20 @@ protected function sendItemReplaceData($entryId, KalturaBaseEntry $entry ,Kaltur
 			$flavor = KBatchBase::$kClient->flavorAsset->add($updatedEntryId, $flavorAsset);
 			KBatchBase::$kClient->flavorAsset->setContent(KBatchBase::$kClient->getMultiRequestResult()->id, $flavorResource);		// TODO: use flavor instead of getMultiRequestResult
 		}
+
+		$requestResults = KBatchBase::$kClient->doMultiRequest();
+		if(is_array($requestResults))
+			foreach($requestResults as $requestResult)
+			{
+				if(is_array($requestResult) && isset($requestResult['code']))
```
yossipapi: I see this same validation is happening few times in the code so why not move it into a function?

PR #4680, opened 2016-09-11
Lynx 12.1.0 sup7815

inbal-ben-david:


MosheMaorKaltura:
1. Load the cuepoint list in pages.
2. Break the script into 2 - one that only handles cue point restoration 
retoreCuepoint <entryId> <updatedAtGreaterThan>
3. Server code fix for entryDelete-cuepoints move to other pull request

PR #4684, opened 2016-09-12
Lynx 12.0.0 webc 710 archive

asafrob:
https://kaltura.atlassian.net/browse/PLAT-5963
https://kaltura.atlassian.net/browse/PLAT-5964
https://kaltura.atlassian.net/browse/PLAT-5965
https://kaltura.atlassian.net/browse/PLAT-5966


alpha/lib/data/live/kLiveEntryRecordingOptions.php
```diff
@@ -23,7 +23,12 @@ class kLiveEntryRecordingOptions
 	 * @var boolean
 	 */
 	protected $shouldCopyThumbnail;
-	
+
+	/**
+	 * @var boolean
```
yossipapi: Please use KalturaNullableBoolean instead of boolean and in your code make sure that in all places null will be handled as identical to false.

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -341,7 +341,12 @@ private function createRecordedEntry(LiveEntry $dbEntry, $mediaServerIndex)
 			$recordedEntry->setModerationStatus($dbEntry->getModerationStatus());
 			$recordedEntry->setIsRecordedEntry(true);
 			$recordedEntry->setTags($dbEntry->getTags());
-			
+
+			// make the recorded entry to be "hidden" in search so it won't return in entry list action
+			if ($dbEntry->getRecordingOptions() && $dbEntry->getRecordingOptions()->getShouldMakeHidden())
```
yossipapi: If its nor very complicated think about moving this check to mySearchUtils::setDisplayInSearch which contains the logic of what the displayInSearch value should be for each object

alpha/lib/data/live/kLiveEntryRecordingOptions.php
asafrob: Why should this be different than the other 3 boolean members of this class?

api_v3/lib/KalturaLiveEntryService.php
asafrob: this is specific to live entries, so doing it somewhere else will generate more code lines for checks and will be less object oriented since I'll have to check types

alpha/lib/data/live/kLiveEntryRecordingOptions.php
yossipapi: Sorry overlooked the fact that this is the core class and not the API one.

PR #4698, opened 2016-09-13
Lynx 12.2.0 plat5967 plat5968

asafrob:


alpha/scripts/utils/migrationWebcastUpdateMakeHiddenAndDisplayInSearch.php
```diff
@@ -0,0 +1,109 @@
...
+require_once(__DIR__ . '/../bootstrap.php');
+
+if (count($argv) < 2)
+{
+    die('pleas provide partner id as input' . PHP_EOL .
+        'to run script: ' . basename(__FILE__) . ' X' . PHP_EOL .
+        'whereas X is the path to the partner ids file' . PHP_EOL);
+}
+
+$dryRun = !(count($argv) == 3 && @$argv[2] == "realrun");
```
yossipapi: Don't use @ since it silences error if there are any.

alpha/scripts/utils/migrationWebcastUpdateMakeHiddenAndDisplayInSearch.php
```diff
@@ -0,0 +1,109 @@
...
+if ($dryRun)
+{
+    KalturaLog::debug('this is a dry run. pass "realrun" as 2nd param to make it actually do stuff');
+}
+
+$partners = file(@$argv[1]);
```
yossipapi: Use something like this for fetching the file:
$entriesFilePath = $argv[1];
$entries = file ( $entriesFilePath ) or die ( 'Could not read file [$entriesFilePath]!' );

alpha/scripts/utils/migrationWebcastUpdateMakeHiddenAndDisplayInSearch.php
```diff
@@ -0,0 +1,109 @@
...
+        die('no such partner.'.PHP_EOL);
+    }
+
+    $liveEntries = array();
+
+    $c = new Criteria();
+    $c->add(entryPeer::PARTNER_ID, $partnerId);
+    $c->add(entryPeer::SOURCE, EntrySourceType::LIVE_STREAM);
+
+    $con = myDbHelper::getConnection(myDbHelper::DB_HELPER_CONN_PROPEL2);
```
yossipapi: Remove this line and don't send con to the select.
Applies to second select as well.

alpha/scripts/utils/migrationWebcastUpdateMakeHiddenAndDisplayInSearch.php
```diff
@@ -0,0 +1,109 @@
...
+
+$partners = file(@$argv[1]);
+foreach ($partners as $partnerId)
+{
+    $partnerId = trim($partnerId);
+    KalturaLog::debug("running script for partner [" . $partnerId . "]");
+    $partner = PartnerPeer::retrieveByPK($partnerId);
+    if(!$partner)
+    {
+        die('no such partner.'.PHP_EOL);
```
yossipapi: Maybe think of using KalturaLog to log that the partner id is not found and than continue to next one instead of die.

alpha/scripts/utils/migrationWebcastUpdateMakeHiddenAndDisplayInSearch.php
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    $liveEntries = array();
+
+    $c = new Criteria();
+    $c->add(entryPeer::PARTNER_ID, $partnerId);
+    $c->add(entryPeer::SOURCE, EntrySourceType::LIVE_STREAM);
```
yossipapi: Select where status is not deleted.

PR #4728, opened 2016-09-20
validate single event duration doesn't exceed 24 hours. this is right…

gotlieb:
… for every single event and any event recurrence.

PLAT-5485
@MosheMaorKaltura 
merge this in Lynx-12.3.0

### DIFF
--- a/plugins/schedule/base/lib/api/KalturaScheduleEvent.php
+++ b/plugins/schedule/base/lib/api/KalturaScheduleEvent.php
@@ -290,7 +290,7 @@ abstract class KalturaScheduleEvent extends KalturaObject implements IRelatedFil
 		$this->validate($this->startDate, $this->endDate);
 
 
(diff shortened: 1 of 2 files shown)


plugins/schedule/base/lib/api/KalturaScheduleEvent.php
```diff
@@ -288,9 +288,23 @@ public function validateForInsert($propertiesToSkip = array())
 		$this->validatePropertyNotNull('startDate');
 		$this->validatePropertyNotNull('endDate');
 		$this->validate($this->startDate, $this->endDate);
-		
+
+
+		$maxSingleEventDuration = SchedulePlugin::getScheduleEventmaxDuration();
```
MosheMaorKaltura: Change to getSingleScheduleEventMaxDuration

gotlieb: done - and tested

PR #4733, opened 2016-09-22
better check if entry is of type live than checking the source type

yossipapi:


alpha/lib/model/DeliveryProfilePeer.php
```diff
@@ -186,7 +186,8 @@ public static function getLocalDeliveryByPartner($entryId, $streamerType = Playb
 		}
 
 		$isSecured = $checkSecured ? self::isSecured($partner, $entry) : false;
-		$isLive = in_array($entry->getSource(), LiveEntry::$liveSourceTypes) ? true : false;
+		
+		$isLive = $entry->getType() == entryType::LIVE_STREAM ? true : false;
```
MosheMaorKaltura: $isLive = $entry->getType() == entryType::LIVE_STREAM; 
will do the same

PR #4755, opened 2016-09-28
make sure resource parent id is valid also on update

gotlieb:
PLAT-5513


tan-tan-kanarek:
I believe we can change the parent-id to be insert only instead.

PR #4764, opened 2016-09-29
when bitrate is low but has width and height should not be considered…

noam-arad:
… audio


noam-arad:
Full compatCheck was run on playmanifest requets and the correct diff was seen.

PR #4777, opened 2016-10-10
add sort on summary field in schedule events

david-winder-kaltura:
PLAT6164 #time 1d


tan-tan-kanarek:
+@yossipapi 
Summary in the sphinx is field and not attribute therefore the sorting by summary won't be done on the sphinx but on the DB and therefore it will fail with pagination.
Note that also after adding summary attribute to the sphinx you still will have index size issue and you should consider indexing only the first X characters into the summary attribute.

yossipapi:
He is gong to add the summery field as an rt_attr_string same as we do in category|entry so we will be able to do the pagination in sphinx.

As to the size limit this is something that should be decided by product of far into the string we should sort.

@david-winder-kaltura I also see you did not update the sphinx config files please do so if not this entire change will not work as mentioned by @tan-tan-kanarek

PR #4778, opened 2016-10-10
Allow more than 3 KAC users

DBezemer:


jessp01:
@DBezemer, can you also change: https://github.com/kaltura/server/blob/Lynx-12.4.0/deployment/base/scripts/init_data/01.Partner.template.ini#L83 from 10 to -1? 
Thanks,

PR #4811, opened 2016-10-25
Lynx 12.4.0 plat 6093

NadavHarnik:


plugins/event_notification/providers/bpm/lib/api/KalturaBusinessProcessServer.php
```diff
@@ -69,6 +69,13 @@
 	public $type;
 	
 	/**
+	 * The dc of the server 0,1,null(-1)
+	 * @var KalturaNullableBoolean
+	 * @filter eq,gt,lt
```
tan-tan-kanarek: None of these filters is relevant, you can remove them.

plugins/event_notification/providers/bpm/lib/api/KalturaBusinessProcessServer.php
```diff
@@ -69,6 +69,13 @@
 	public $type;
 	
 	/**
+	 * The dc of the server 0,1,null(-1)
+	 * @var KalturaNullableBoolean
```
tan-tan-kanarek: The type should be int.

plugins/event_notification/providers/bpm/lib/api/filters/KalturaBusinessProcessServerFilter.php
```diff
@@ -5,11 +5,40 @@
...
+	 * @see KalturaFilter::toObject()
+	 */
+	public function toObject ( $object_to_fill = null, $props_to_skip = array() )
+	{
+		if(!$this->isNull('currentDc') && KalturaNullableBoolean::toBoolean($this->currentDc))
+			$this->dcEqual = kDataCenterMgr::getCurrentDcId();
+
+		elseif(!$this->isNull('currentDcOrNull') && KalturaNullableBoolean::toBoolean($this->currentDcOrNull))
+		{
+			if(kDataCenterMgr::getCurrentDcId())
```
tan-tan-kanarek: You can use more simple formula:
$this->dcGreaterThenOrNull = 1 - kDataCenterMgr::getCurrentDcId ()

jessp01:
Note that you should also add the 'dc' column in https://github.com/kaltura/server/blob/Lynx-12.4.0/deployment/base/sql/01.kaltura_ce_tables.sql#L2349

plugins/event_notification/providers/bpm/lib/api/filters/KalturaBusinessProcessServerFilter.php
```diff
@@ -6,12 +6,12 @@
 class KalturaBusinessProcessServerFilter extends KalturaBusinessProcessServerBaseFilter
 {
 	/**
-	 * @var KalturaNullableBoolean
+	 * @var bool
```
tan-tan-kanarek: Boolean can't be null and therefor it will be false if it wasn't supplied, I think that using KalturaNullableBoolean was better.

PR #4815, opened 2016-10-25
add field in XML bulkupload to keep-Manual-Thumbnails

david-winder-kaltura:
PLAT-6279 # time 5h

### DIFF
--- a/plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
+++ b/plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
@@ -609,9 +609,7 @@ class BulkUploadEngineXml extends KBulkUploadEngine
 			}
 		}
 
-		$keepManualThumbnails = false;
-		if(isset($item->keepManualThumbnails) && $item->keepManualThumbnails == 1)
-			$keepManualThumbnails = true;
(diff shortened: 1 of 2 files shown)


plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
```diff
@@ -608,7 +608,11 @@ protected function handleItemUpdate(SimpleXMLElement $item)
 				$resource = null;
 			}
 		}
-		
+
+		$keepManualThumbnails = false;
+		if(isset($item->keepManualThumbnails) && $item->keepManualThumbnails == 1)
```
MosheMaorKaltura: keepManualThumbnails  should be string, when value is strtolow(true) then set it

plugins/bulk_upload/xml/xml/ingestion.xsd
```diff
@@ -141,6 +141,11 @@
 					<xs:documentation>Text that describes the entry</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element maxOccurs="1" minOccurs="0" name="keepManualThumbnails" type="xs:string">
```
tan-tan-kanarek: The element should be boolean.

PR #4836, opened 2016-11-01
TR-1663 - Added option to transform DFXP files to a more "common" format

rkreich:


plugins/integration/providers/voicebase/lib/kVoicebaseFlowManager.php
```diff
@@ -161,5 +166,95 @@ private function setObjectContent($assetObject, $content, $accuracy = null, $for
...
+	}
+
+	private function formatDuration($int)
+	{
+		$durationSeconds = floor($int / 1000);
+		$durationMilliseconds = $int % 1000;
+		return $durationSeconds . '.' . str_pad($durationMilliseconds, 3, 0, STR_PAD_LEFT);
+	}
+
+	private function integerToTime($int)
```
MosheMaorKaltura: This function can be achieved by using DateTime(). 
Preferred than implementing it.

PR #4850, opened 2016-11-03
TR-1663 - Voicebase DFXP Transformation

rkreich:
Fixed integerToTime method


rkreich:
Closing. Will be reopened against Lynx-12.5.0

PR #4880, opened 2016-11-09
added label support to manifest

NadavHarnik:
PLAT-6338 #time 0.5d

### DIFF
--- a/alpha/apps/kaltura/lib/kManifestRenderers.php
+++ b/alpha/apps/kaltura/lib/kManifestRenderers.php
@@ -781,9 +781,9 @@ class kM3U8ManifestRenderer extends kMultiFlavorManifestRenderer
 		foreach($this->flavors as $flavor)
 		{
 			// Sperate audio flavors from video flavors
-			if ( isset($flavor['audioLanguage']) || isset($flavor['label']) ) {
+			if ( isset($flavor['audioLanguage']) || isset($flavor['$audioLabel']) ) {
 				$language = (isset($flavor['audioLanguage'])) ? $flavor['audioLanguage'] : 'und';
-				$languageName = (isset($flavor['label'])) ? $flavor['label'] : $flavor['audioLanguageName'];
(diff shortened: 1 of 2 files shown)


alpha/apps/kaltura/lib/kManifestRenderers.php
```diff
@@ -781,9 +781,9 @@ protected function getManifestFlavors()
 		foreach($this->flavors as $flavor)
 		{
 			// Sperate audio flavors from video flavors
-			if (isset($flavor['audioLanguage'])) {
-				$language = $flavor['audioLanguage'];
-				$languageName = $flavor['audioLanguageName'];
+			if ( isset($flavor['audioLanguage']) || isset($flavor['label']) ) {
+				$language = (isset($flavor['audioLanguage'])) ? $flavor['audioLanguage'] : 'und';
+				$languageName = (isset($flavor['label'])) ? $flavor['label'] : $flavor['audioLanguageName'];
```
yossipapi: Will this always be set?  $flavor['audioLanguageName']

alpha/lib/model/DeliveryProfile.php
```diff
@@ -315,12 +315,16 @@ protected function getFlavorAssetInfo($url, $urlPrefix = '', $flavor = null)
...
 		$audioLanguageName = null;
+		$label = null;
 		if ($flavor) {
 			if (is_callable(array($flavor, 'getFileExt'))) {
 				$ext = $flavor->getFileExt();
 			}
 			//Extract the audio language code from flavor
 			if ($flavor->hasTag(assetParams::TAG_AUDIO_ONLY)) {
+				if(is_callable(array($flavor, 'getLabel'))) {
+					$label = $flavor->getLabel();
```
yossipapi: If getLabel is not callable that you will remain with a null value in the label param, shouldn't there be some kind of a default value? (like you have for "audioLanguage")

PR #4884, opened 2016-11-09
Lynx 12.5.0 plat 6357

noam-arad:


noam-arad:
@anatolkaltura  according to @erankor  we should remove the sps-id=27 but I'm not sure if for all use-cases.

anatolkaltura:
@noam-arad, @erankor 
Removal of 'sps-id=27' contradicts the existing code in - https://github.com/kaltura/play-server/blob/v3.0.0/lib/media/KalturaFfmpegParams.js#L316

erankor:
To clarify - sps-id=27 MUST be used for live, and MUST NOT be used for vod. Not sure if this is currently the case or not.

noam-arad:
verified that this doesn't affect live.

PR #4885, opened 2016-11-10
move decision logic for copy asset into asset object

david-winder-kaltura:


alpha/lib/model/asset.php
```diff
@@ -767,5 +767,10 @@ public function getentry(PropelPDO $con = null)
 	public function getassetParams(PropelPDO $con = null)
 	{
 		return assetParamsPeer::retrieveByPk($this->flavor_params_id);
-	}	
+	}
+
+	public function shouldCopyAssetInEntry()
```
MosheMaorKaltura: rename to - shouldCopyOnEntryCopy

PR #4892, opened 2016-11-13
Enable audioPassthrough streams and configured source audio bitrate

gadyaari:


plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -199,8 +199,8 @@ protected function appendLiveParams(LiveStreamEntry $entry, WowzaMediaServerNode
 		{
 			$video->addChild('Codec', $videoCodec);
 			$audio->addChild('Codec', $audioCodec);
-			$audio->addChild('Bitrate', 96000);
-			
+			if ($audioCodec === 'AAC')
```
kalturaguy: better to do != 'PassThru'

PR #4899, opened 2016-11-14
Lynx 12.5.0 sphinx to db

david-winder-kaltura:


david-winder-kaltura:
Add back the apply filter because 'doCount' don't call the applyFilter for sphinx criteria

PR #4904, opened 2016-11-15
Lynx 12.4.0 ps 2666

hilak:


deployment/updates/sql/2016_11_11_alter_category_entry_table_add_creator_kuser_id_table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE category_entry ADD creator_kuser_id INT NULL AFTER privacy_context;
```
tan-tan-kanarek: Please update also 01.kaltura_ce_tables.sql.

api_v3/lib/types/KalturaCategoryEntry.php
```diff
@@ -47,13 +47,23 @@ class KalturaCategoryEntry extends KalturaObject implements IRelatedFilterable
 	 */
 	public $status;
 	
+	/**
+	 * 
+	 * CategroyEntry creator puser ID
+	 * @var string
+	 * @readonly
+	 * @filter eq,in
```
rkreich: @hilak @erankor - I think that `not` and `notin` might be useful here.

PR #4905, opened 2016-11-16
scheduled task - modify entry engine

shpiNi:


plugins/scheduled_task/lib/object_task_engine/KObjectTaskModifyEntryEngine.php
```diff
@@ -0,0 +1,154 @@
...
+			}
+			else
+				KalturaLog::info("found no input metadata objects for entry $entryId");
+		}
+
+		if(isset($objectTask->inputUserId) && $objectTask->inputUserId != '')
+			$entryObj->userId = $objectTask->inputUserId != 'null' ? $objectTask->inputUserId : null;
+		
+		if(isset($objectTask->inputUserId) && $objectTask->inputEntitledUsersEdit != '')
+			$entryObj->entitledUsersEdit = $objectTask->inputEntitledUsersEdit != 'null' ? $objectTask->inputEntitledUsersEdit : '';
```
rkreich: Why use`'null'`?

plugins/scheduled_task/lib/api/types/object_tasks/KalturaModifyEntryObjectTask.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @var int
+	 */
+	public $inputMetadataProfileId;
+
+	/**
+	 * array of {input metadata xpath location,entry field} objects
+	 *
+	 * @var KalturaKeyValueArray
+	 */
+	public $inputMetadata;
```
tan-tan-kanarek: What should be in the key and what in the value?

plugins/scheduled_task/lib/api/types/object_tasks/KalturaModifyEntryObjectTask.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+/**
+ * @package plugins.scheduledTask
+ * @subpackage api.objects.objectTasks
+ */
+class KalturaModifyEntryObjectTask extends KalturaObjectTask
```
tan-tan-kanarek: Please rename to KalturaModifyEntryTask

rkreich: This is actually consistent with other objects https://github.com/kaltura/server/tree/Lynx-12.5.0/plugins/scheduled_task/lib/api/types/object_tasks

plugins/scheduled_task/lib/object_task_engine/KObjectTaskModifyEntryEngine.php
```diff
@@ -0,0 +1,158 @@
...
+			$client->baseEntry->update($entryId, $entryObj);
+		}
+		catch(Exception $e)
+		{
+			KalturaLog::notice("problem with entry update entry id $entryId - " . $e->getMessage());
+		}
+		KBatchBase::unimpersonate();
+	}
+	
+	private static function updateMetadataObj(KalturaBaseEntry $entryResultForMetadataUpdate, &$metadataPlugin, $outputMetadataArr, $outputMetadataProfileId, $metadataFilter)
```
rkreich: Why static? (same goes for other methods)
@MosheMaorKaltura @tan-tan-kanarek agree that we usually don't use static methods in such cases?

plugins/scheduled_task/lib/object_task_engine/KObjectTaskModifyEntryEngine.php
```diff
@@ -0,0 +1,158 @@
...
+		{
+			$metadataInputXmlStr = $metadataInputResult->objects[0]->xml;
+			$xmlObj = new SimpleXMLElement($metadataInputXmlStr);
+			
+			foreach($inputMetadataArr as $inputMetadataItem)
+			{					
+				$xpathName = $inputMetadataItem->key;
+				$fieldName = $inputMetadataItem->value;
+	
+				$metadataValue = $xmlObj->$xpathName;
```
rkreich: To better explain my comment that was removed by last commit.
Lets assume that we have the following metadata on the entry:
```xml
<metadata>
<FieldA>ValueInField1</FieldA>
<FieldC>ValueInField3</FieldC>
</metadata>
```

We want to append `FieldB`.
The XSD defines the following order:
- FieldA
- FieldB
- FieldC

Current code will produce and try to save:
```xml
<metadata>
<FieldA>ValueInField1</FieldA>
<FieldC>ValueInField3</FieldC>
<FieldB>ValueInField2</FieldB>
</metadata>
```
This will not pass XSD validation.

plugins/scheduled_task/lib/object_task_engine/KObjectTaskModifyEntryEngine.php
```diff
@@ -0,0 +1,217 @@
...
+					$emptyXmlObjItem = (string)$entryResultForMetadataUpdate->$entryFieldName;
+					$tempXmlObj->$metadataFieldName = (string)$entryResultForMetadataUpdate->$entryFieldName;
+					break;
+				}
+			}
+		}
+		return $tempXmlObj->asXml();
+	}
+	
+	private function updateEntryFromMetadata(&$metadataPlugin, array $inputMetadataArr, KalturaBaseEntry $entryObj, $metadataFilter)
```
rkreich: Objects are always passed by reference, remove `&` and add the type of the object on method's signature

plugins/scheduled_task/lib/object_task_engine/KObjectTaskModifyEntryEngine.php
```diff
@@ -0,0 +1,217 @@
...
+		$inputMetadataProfileId = $objectTask->inputMetadataProfileId;
+		$inputMetadataArr = $objectTask->inputMetadata;
+		
+		$metadataFilter = new KalturaMetadataFilter();
+		$metadataFilter->metadataObjectTypeEqual = KalturaMetadataObjectType::ENTRY;
+		$metadataFilter->objectIdEqual = $entryId;
+		
+		if($outputMetadataProfileId != 0 && !empty($outputMetadataArr))
+		{
+			$entryResultForMetadataUpdate = $client->baseEntry->get($entryId);
```
noam-arad: You should surround all API requests with try/catch to avoid an uncaught exception, this way you'll always be able to unimpersonate.

noam-arad: what if the entry is not found for some reason? Please verify this

rkreich: There is a generic unimpersonate in https://github.com/kaltura/server/blob/Lynx-12.6.0/plugins/scheduled_task/batch/ScheduledTaskRunner/KScheduledTaskRunner.class.php#L133

plugins/scheduled_task/lib/object_task_engine/KObjectTaskModifyEntryEngine.php
```diff
@@ -0,0 +1,225 @@
...
+			KalturaLog::notice("problem with entry update entry id $entryId - " . $e->getMessage());
+		}
+		KBatchBase::unimpersonate();
+	}
+	
+	private function updateMetadataObj(KalturaBaseEntry $entryResultForMetadataUpdate, &$metadataPlugin, $outputMetadataArr, $outputMetadataProfileId, $metadataFilter)
+	{
+		$entryId = $entryResultForMetadataUpdate->id;
+		
+		try
```
MosheMaorKaltura: try catch is not required here , if this action will fail, then we rather see error in the batch logger and fail the action.

plugins/scheduled_task/lib/object_task_engine/KObjectTaskModifyEntryEngine.php
```diff
@@ -0,0 +1,225 @@
...
+				$metadataPlugin->metadata->add($outputMetadataProfileId, KalturaMetadataObjectType::ENTRY, $entryId, $xmlData);
+			}
+			catch(Exception $e)
+			{
+				KalturaLog::notice("problem with metadata add entry id $entryId - " . $e->getMessage());
+			}
+		}
+	}
+	
+	private function getUpdatedMetadataXmlStrFromEntry(KalturaBaseEntry $entryResultForMetadataUpdate, SimpleXMLElement $emptyXmlObj, SimpleXMLElement $currentXmlObj, array $outputMetadataArr)
```
MosheMaorKaltura: emptyXmlObj rename to templateXmlObj

plugins/scheduled_task/lib/object_task_engine/KObjectTaskModifyEntryEngine.php
```diff
@@ -0,0 +1,225 @@
...
+	{
+		$tempXmlObj = $emptyXmlObj;		
+		
+		KalturaLog::debug("current xml object - " . print_r($currentXmlObj, true));
+		KalturaLog::debug("output metadata array - " . print_r($outputMetadataArr, true));
+		
+		foreach($tempXmlObj as $metadataFieldName => $emptyXmlObjItem)
+		{
+			foreach($currentXmlObj as $currMetadataFieldName => $currentXmlObjItem)
```
MosheMaorKaltura: Instead of the internal loop , check if value exist in the currentXmlObjItem mand it is not empty

MosheMaorKaltura:
Reviewed

PR #4922, opened 2016-11-21
Support constant bit rate for transcoded flavors

gadyaari:


plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -177,9 +177,10 @@ public function serveAction($streamName, $hostname = null)
 
 	protected function appendLiveParams(LiveStreamEntry $entry, WowzaMediaServerNode $mediaServer = null, SimpleXMLElement $encodes, liveParams $liveParams)
 	{
+		$conversionExtraParam = json_decode($liveParams->getConversionEnginesExtraParams());
 		$streamName = $entry->getId() . '_' . $liveParams->getId();
 		$videoCodec = 'PassThru';
-		$audioCodec = (strpos($liveParams->getConversionEnginesExtraParams(), 'audioPassthrough') !== false) ? 'PassThru' : 'AAC';
+        $audioCodec = ($conversionExtraParam !== null && $conversionExtraParam->audioPassthrough !== null) ? 'PassThru' : 'AAC';
```
yossipapi: If audioPassthrough is defined check its not false

yossipapi: To avoid writing long if you can simply do
($conversionExtraParam && $conversionExtraParam->audioPassthrough)

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -285,13 +286,20 @@ protected function appendLiveParams(LiveStreamEntry $entry, WowzaMediaServerNode
...
-		if ($skipFrameCountPos !== false)
-		{
-			$skipFrameCount = $video->addChild('SkipFrameCount');
-			preg_match('/SkipFrameCount=(\d+)/', $liveParams->getConversionEnginesExtraParams(), $skipFrameValue, $skipFrameCountPos);
-			$skipFrameCount->addChild('Value', (int)$skipFrameValue[1]);
-		}
+        if ($conversionExtraParam !== null && $conversionExtraParam->skipFrameCount !== null)
+        {
+            $skipFrameCount = $video->addChild('SkipFrameCount');
+            $skipFrameCount->addChild('Value', (int)$conversionExtraParam->skipFrameCount->value);
```
yossipapi: Why not simply use $conversionExtraParam->skipFrameCount and json that looks like {"skipFrameCount":8} 

Also don't need to casting to int

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -285,13 +286,20 @@ protected function appendLiveParams(LiveStreamEntry $entry, WowzaMediaServerNode
...
+        }
+
+		if ($conversionExtraParam !== null && $conversionExtraParam->constantBitrate !== null)
+        {
+            $parameters = $video->addChild('Parameters');
+            $parameter = $parameters->addChild('Parameter');
+            $parameter->addChild('Name', 'mainconcept.bit_rate_mode');
+            $parameter->addChild('Value', 0);
```
yossipapi: Why not use the value of constantBitrate as we discussed this value represents a Wowza enum.

PR #4936, opened 2016-11-22
Don't update cusomData if value hasn't changed

yossipapi:


tan-tan-kanarek:
removeFromCustomData also adds key to $this->oldCustomDataValues without checking if the key existed.
Same with incInCustomData, and more weirdly, decInCustomData doesn't affect $this->oldCustomDataValues at all.

yossipapi:
I will give all the above a look and will change accordingly.

yossipapi:
in incInCustomData there is no point of checking if old value existed since it will always end with incrementing the value or if not existed setting it to the delta value so it will always end with the column being modified.

Regarding decInCustomData you are absolutely right it should affect the oldCustomDataValues since if this was the only change made the save will never happen. I will add this change. Again here as it will always end with a value change no need to do the verification.

AS to removeFromCustomData you are correct it should have the validation if the value was previously set or not. I will add a similar validation to what I did in the putInCustomData.

PR #4942, opened 2016-11-23
verify no use of empty privacy_context

noam-arad:


plugins/tag_search/sphinx/kTagFlowManager.php
```diff
@@ -73,7 +79,8 @@ public function objectCreated (BaseObject $object)
     			/* @var $object categoryEntry */
      			$privacyContexts = $object->getPrivacyContext() != "" ? self::trimObjectTags($object->getPrivacyContext()) : array();
     			if (!count($privacyContexts))
-    				$privacyContexts[] = kEntitlementUtils::DEFAULT_CONTEXT; 
+    				$privacyContexts[] = kEntitlementUtils::DEFAULT_CONTEXT;
+			    $privacyContexts = array_map('addNullPC', $privacyContexts);
```
MosheMaorKaltura: array_map should get parameter (array("self","addNullPC"),array)

plugins/tag_search/sphinx/kTagFlowManager.php
```diff
@@ -73,7 +79,8 @@ public function objectCreated (BaseObject $object)
     			/* @var $object categoryEntry */
      			$privacyContexts = $object->getPrivacyContext() != "" ? self::trimObjectTags($object->getPrivacyContext()) : array();
```
MosheMaorKaltura: If we change this code then need to refactor that the fallback will be to set $privacyContexts to be kEntitlementUtils::DEFAULT_CONTEXT; , hence no need to check count

PR #4944, opened 2016-11-23
Lynx 12.6.0 ps 2666

hilak:


hilak:
merging after review by @erankor .

configurations/admin.template.ini
```diff
@@ -905,7 +905,7 @@ moduls.LikeUsersSearch.group = GROUP_ENABLE_DISABLE_FEATURES
 moduls.PeerCategoryLevelModeration.enabled = true
 moduls.PeerCategoryLevelModeration.permissionType = 2
 moduls.PeerCategoryLevelModeration.label = "Category-Level Peer Moderation"
```
rkreich: @hilak, permission name is `FEATURE_BLOCK_CATEGORY_MODERATION_SELF_APPROVE`
Shouldn't this label be similar? something like _Block Self Category Moderation Approve_?
@erankor, what do you think?

PR #4946, opened 2016-11-23
reduce content-dist service cache time

shpiNi:


shpiNi:
@erankor please review.

PR #4947, opened 2016-11-24
save only last 100 bpm starts to save space in custom_data

noam-arad:


plugins/event_notification/providers/bpm/lib/model/BusinessProcessNotificationTemplate.php
```diff
@@ -149,6 +149,7 @@ public function addCaseId(BaseObject $object, $caseId, $processId = null)
 			}
 			
 			$values = $this->getCaseValues($object, true, $processId);
+			$values = array_slice($values, count($values) - 99);
```
MosheMaorKaltura: To get the last 100 shoud be  array_slice($values, -100);

noam-arad: The next line adds the new value so in all the array will have 100 values

PR #4956, opened 2016-11-28
addContent- extract ext from redirect url

shpiNi:
@MosheMaorKaltura  please review.


configurations/base.ini
```diff
@@ -288,6 +288,8 @@ audio_file_ext[]=m4a
 audio_file_ext[]=aiff
 audio_file_ext[]=aif
 
+video_curl_content_type["video/x-flv"]=flv
```
rkreich: Only FLV?
Can't we find a better mapping of mime type to ext?

rkreich: OK

PR #4960, opened 2016-11-29
minimal template needed in order for Live to work.

jessp01:


jessp01:
@yossipapi,

I'm afraid that's not true.
Without /opt/kaltura/app/configurations/media_servers.ini in place, the request for playManifest with return HTTP 500. From Apache's log:

[Fri Dec 02 11:46:25.947820 2016] [authz_core:error] [pid 3392] [client 188.28.130.203:42253] AH01630: client denied by server configur
ation: /usr/share/pear/data, referer: http://54.165.75.49/p/101/sp/10100/playManifest/entryId/0_ukxnmrps/format/applehttp/protocol/http
/uiConfId/23448188/a.m3u8?referrer=aHR0cDovLzU0LjE2NS43NS40OQ==&playSessionId=1d64a76c-e663-ceda-d3f5-d4e9ed31837f
[Fri Dec 02 11:47:05.412729 2016] [:error] [pid 3398] [client 54.165.75.49:41332] Cannot find configuration file [/opt/kaltura/app/conf
igurations/media_servers.ini]

Feel free to test it yourself. This is true to 12.6.0 as well as older versions.

yossipapi:
If this is the case than I will take a look why this happened and resolve.
Basically this file is no longer relevant and I would prefer to no longer use it.

jessp01:
So long as it stops generating HTTP 500 when calling playManifest I'm fine with that but for now it's mandatory,

PR #4968, opened 2016-12-01
this is for the 2-factor auth feature and should not be enabled for C…

jessp01:
…E/OnPrem by default.

### DIFF
--- a/configurations/local.template.ini
+++ b/configurations/local.template.ini
@@ -137,8 +137,8 @@ metadata_sphinx_num_of_int_fields = 10
 ;the range is used when checking service actions permissions and allowing to access certain 
 ;services without KS from the internal servers
 
-partner_otp_internal_ips = @IP_RANGE@
-otp_required_partners[] = -2
+;partner_otp_internal_ips = @IP_RANGE@
+;otp_required_partners[] = -2
 
 internal_ip_range = @IP_RANGE@
 ;Boolean flag signifying whether private IPs can be accepted as load-balancers in the environment 


configurations/local.template.ini
```diff
@@ -137,8 +137,8 @@ metadata_sphinx_num_of_int_fields = 10
 ;the range is used when checking service actions permissions and allowing to access certain 
 ;services without KS from the internal servers
 
-partner_otp_internal_ips = @IP_RANGE@
-otp_required_partners[] = -2
+;partner_otp_internal_ips = @IP_RANGE@
```
yossipapi: Wouldn't it be better to leave this configuration and simply provide ip range that matches all and add the missing settings.remoteAddrHeaderSalt? 

what do you think?

DBezemer: @yossipapi in that case the OTP box will still be shown, and there are cases it prevents logging in. Better to have it disabled by default and use the SaaS local.ini (that already has these settings) to set this for SaaS only.

jessp01: We have the settings.remoteAddrHeaderSalt in place. However, since we do not set up a 2 factor auth by default, there is no reason at all why these directives should be enabled..
The internal_ip_range remains enabled of course but partner_otp_internal_ips relates ONLY to the OTP mechanism, which again, is not activated by default.

yossipapi: Good enough reason.
Will do the merge.

yossipapi:
Opened https://github.com/kaltura/server/pull/4983 for merging to latest branch

PR #4986, opened 2016-12-06
update base.ini

MosheMaorKaltura:


MosheMaorKaltura:
The changes will be only in local.ini to avoid ce and onprem issues.

PR #5011, opened 2016-12-12
have the re-authenticate link show only when re-authenticate is press…

noam-arad:
…ed, this creates a much cleare UI


noam-arad:
This was created to have a better UI for youtube_api re-authentication

PR #5026, opened 2016-12-14
when serilzing json reponse we should enforce declared types so thatq…

noam-arad:
… clients can count on the correct types being returned


tan-tan-kanarek:
Seems like a significant performance penalty, I think that there are only few places that we return the wrong types, we can find them using scripts and fix them one by one.

PR #5037, opened 2016-12-18
In recording flow don't delete old assets

yossipapi:


MosheMaorKaltura:
Verified , also automatic tests pass.

PR #5040, opened 2016-12-19
Receive JSON object form Wowza and create "smart" transcode.xml

gadyaari:


plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -24,21 +24,39 @@ public function initService($serviceId, $serviceName, $actionName)
...
+			'videodatarate' => 0,
+			'width' => 0,
+			'height' => 0,
+			'framerate' => 0,
+			'videocodecidstring' => "",
+			'audiocodecidstring' => ""
+		);
+		if ($extraParams !== "")
+		{
+			$streamParametersArray = array_merge($streamParametersArray, json_decode($extraParams, true));
```
yossipapi: You need to validate that the string is a valid Json before using it.
You can do it with this function for example:
function isValidJson($strJson) { 
    json_decode($strJson); 
    return (json_last_error() === JSON_ERROR_NONE); 
}

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -24,21 +24,39 @@ public function initService($serviceId, $serviceName, $actionName)
...
+		$streamParametersArray = array(
+			'streamName' => $streamName,
+			'hostname' => $hostname,
+			'audiodatarate' => 0,
+			'videodatarate' => 0,
+			'width' => 0,
+			'height' => 0,
+			'framerate' => 0,
+			'videocodecidstring' => "",
+			'audiocodecidstring' => ""
```
yossipapi: videocodecidstring => videoCodecId why do you need the string postfix.
Same goes for audiocodecidstring

Also stay consistent with upper case in long variables you have "streamName" but also "videodatarate" applies for other long variables as well.

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -174,16 +192,105 @@ public function serveAction($streamName, $hostname = null)
...
+			case 'fit-height':
+				$flavorHeight = $flavorResolution['height'];
+				break;
+			case 'fit-width':
+				// Flavor's height is not defined in KMC, calculate it according to ingest/flavor ratio
+				$flavorHeight = $this->calculateFlavorHeight($flavorResolution, $ingestParameters);
+				break;
+		}
+		
+		if ($ingestParameters['height'] !== 0 && $ingestParameters['height'] < $flavorHeight)
```
yossipapi: Why check $ingestParameters['height'] !== 0 
I see that you set it to default value of 0 when initializing streamParametersArray

same goes for $ingestParameters['videodatarate'] !== 0

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -174,16 +192,105 @@ public function serveAction($streamName, $hostname = null)
...
+		
+		return true;
+	}
+	
+	private function getResolutionParameters($liveParams)
+	{
+		$resolutionObject = array(
+			'fitMode' => ''
+		);
+		if(!$liveParams->getWidth() && !$liveParams->getHeight())
```
yossipapi: you can loose this check by simply defining the default value to 'match-source' instead of ''

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -174,16 +192,105 @@ public function serveAction($streamName, $hostname = null)
...
+	}
+	
+	private function getIngestAudioCodec($ingestParameters)
+	{
+		return ($ingestParameters['audiocodecidstring'] === 'AAC') ? 'PassThru' : 'AAC';
+	}
+	
+	protected function appendLiveParams(LiveStreamEntry $entry, WowzaMediaServerNode $mediaServer = null, SimpleXMLElement $encodes, liveParams $liveParams, $streamParametersArray)
 	{
 		$conversionExtraParam = json_decode($liveParams->getConversionEnginesExtraParams());
```
yossipapi: Same comment for json_decode as mentioned above... this could fail if Json is not valid

plugins/media/wowza/services/LiveConversionProfileService.php
gadyaari: done

plugins/media/wowza/services/LiveConversionProfileService.php
gadyaari: Because if the value is still 0 it means that we didn't receive any information from the encode (meaning wowza) and we want to ignore this checkup. Otherwise it will reject all the flavors

plugins/media/wowza/services/LiveConversionProfileService.php
gadyaari: done

plugins/media/wowza/services/LiveConversionProfileService.php
gadyaari: It returns false, which is consistent with the checkups I'm performing later on

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -174,16 +192,107 @@ public function serveAction($streamName, $hostname = null)
...
+	
+	private function isValidJson($strJson)
+	{
+		json_decode($strJson);
+		return (json_last_error() === JSON_ERROR_NONE);
+	}
+	
+	private function calculateFlavorHeight($flavorResolution, $ingestParameters)
+	{
+		return ($ingestParameters['width'] != 0) ? (($flavorResolution['width'] * $ingestParameters['height']) / $ingestParameters['width']) : 0;
```
yossipapi: the strings width | hight | match-source are returning few times in the code why not use constants instead of strings?

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -174,16 +192,107 @@ public function serveAction($streamName, $hostname = null)
...
+	}
+	
+	private function isFlavorCompatibile($ingestParameters, $flavorBitrate, $flavorResolution, $flavorId)
+	{
+		$flavorHeight = 0;
+		switch ($flavorResolution['fitMode'])
+		{
+			case 'match-source':
```
yossipapi: there is no reason to insert case that does nothing.

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -257,39 +366,31 @@ protected function appendLiveParams(LiveStreamEntry $entry, WowzaMediaServerNode
...
 		$video->addChild('Codec', $videoCodec);
 		$video->addChild('Profile', $profile);
-		$video->addChild('Bitrate', $liveParams->getVideoBitrate() ? $liveParams->getVideoBitrate() * 1024 : 240000);
+		$video->addChild('Bitrate', $flavorBitrateValue);
 		$keyFrameInterval = $video->addChild('KeyFrameInterval');
 		$keyFrameInterval->addChild('FollowSource', 'true');
 		$keyFrameInterval->addChild('Interval', 60);
 		
-		if ($conversionExtraParam && $conversionExtraParam->skipFrameCount)
+		$skipFrameCountByMaxFramerate = $this->checkMaxFramerate($streamParametersArray['framerate'], $liveParams->getMaxFrameRate());
```
yossipapi: when looking at:
if ($configuredSkipFrameRate || $skipFrameCountByMaxFramerate

you don't really need to calculate both values for "skipFrameCountByMaxFramerate" & "configuredSkipFrameRate" check the second only f the first is not defined.

PR #5073, opened 2016-12-22
add delivery profile id filtering to playManifest

NadavHarnik:


alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
```diff
@@ -726,6 +727,23 @@ protected function initFlavorAssetArray()
 		}
 	}
 
+	private function shouldIncludeStorageProfile($storageProfile)
+	{
+		if($this->deliveryProfileId)
+		{
+			$deliveryExist = false;
```
MosheMaorKaltura: Flag is not needed , if found one return true

alpha/lib/model/DeliveryProfilePeer.php
```diff
@@ -433,6 +432,27 @@ protected static function selectByDeliveryAttributes($deliveries, DeliveryProfil
...
 		return $partialSupport;
 	}
+
+	/**
+	 * filter deliveries by delivery Profile Id
+	 * @param $deliveries list of deliveries
+	 * @param $deliveryId
+	 * @return array
+	 */
+	protected static function filterByDeliveryIdAttribute($deliveries, $deliveryId)
```
MosheMaorKaltura: rename to getDeliveryById

PR #5077, opened 2016-12-22
when updating entitledUserPuserEdit/Publish we should update all chil…

noam-arad:
…dren,parent and sibling entries

### DIFF
--- a/alpha/lib/model/entry.php
+++ b/alpha/lib/model/entry.php
@@ -2796,8 +2796,9 @@ class entry extends Baseentry implements ISyncableFile, IIndexable, IOwnable, IR
 		}
 
 		//Has entitledUserPuserEdit/Publish changed for current entry
-		if ( ($this->oldCustomDataValues['']['entitledUserPuserEdit'] != $this->getEntitledPusersEdit()) ||
-			($this->oldCustomDataValues['']['entitledUserPuserPublish'] != $this->getEntitledPusersPublish()) )
+		if (  isset($this->oldCustomDataValues['']) &&
+			($this->oldCustomDataValues['']['entitledUserPuserEdit'] != $this->getEntitledPusersEdit() ||
+			$this->oldCustomDataValues['']['entitledUserPuserPublish'] != $this->getEntitledPusersPublish()) )
 		{
 			//Change for parent entry (and thus changing all brother entries.
 			$parentEntry = $this->getParentEntry();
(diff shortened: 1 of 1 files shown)


alpha/lib/model/entry.php
```diff
@@ -2794,7 +2794,39 @@ public function postUpdate(PropelPDO $con = null)
 				$redirectEntry->save();
 			}
 		}
-		
+
+		//Has entitledUserPuserEdit/Publish changed for current entry
+		if ( ($this->oldCustomDataValues['']['entitledUserPuserEdit'] != $this->getEntitledPusersEdit()) ||
```
MosheMaorKaltura: Please add tiny method getOldCustomDataValues(fieldName , nameSpace=null)
And validate that value exist.

alpha/lib/model/entry.php
```diff
@@ -2794,7 +2794,40 @@ public function postUpdate(PropelPDO $con = null)
...
+			if ( $parentEntry->getEntitledPusersEdit() != $this->getEntitledPusersEdit() ||
+				$parentEntry->getEntitledPusersPublish() != $this->getEntitledPusersPublish() )
+			{
+				$parentEntry->setEntitledPusersEdit($this->getEntitledPusersEdit());
+				$parentEntry->setEntitledPusersPublish($this->getEntitledPusersPublish());
+				$parentEntry->save();
+			}
+
+			//Change all child entries.
+			$children = entryPeer::retrieveChildEntriesByEntryIdAndPartnerId($this->getId(), $this->getPartnerId());
```
MosheMaorKaltura: rename to childEntries.

alpha/lib/model/entry.php
```diff
@@ -2794,7 +2794,40 @@ public function postUpdate(PropelPDO $con = null)
 				$redirectEntry->save();
 			}
 		}
-		
+
+		//Has entitledUserPuserEdit/Publish changed for current entry
+		if (  isset($this->oldCustomDataValues['']) &&
+			($this->oldCustomDataValues['']['entitledUserPuserEdit'] != $this->getEntitledPusersEdit() ||
```
MosheMaorKaltura: Add function getOldCustomDataValues(fieldName,namespace='')
This function should return value or null if not exist

alpha/lib/model/entry.php
```diff
@@ -2794,7 +2794,40 @@ public function postUpdate(PropelPDO $con = null)
...
+			$this->oldCustomDataValues['']['entitledUserPuserPublish'] != $this->getEntitledPusersPublish()) )
+		{
+			//Change for parent entry (and thus changing all brother entries.
+			$parentEntry = $this->getParentEntry();
+			if ( $parentEntry->getEntitledPusersEdit() != $this->getEntitledPusersEdit() ||
```
MosheMaorKaltura: when if no parent entry?

PR #5091, opened 2016-12-26
exclude temporary entries created during live from entryReady email n…

inbal-ben-david:
…otification

SUP-9870


MosheMaorKaltura:
Reviewed.

PR #5097, opened 2016-12-28
when getting updated at it shouldnt matter that the actual asset is already deleted, this is an edge case that can happen sometimes

noam-arad:


plugins/content/caption/search/lib/model/CaptionAssetItem.php
```diff
@@ -155,7 +167,7 @@ public function getDefault()
 	 */
 	public function getUpdatedAt($format = 'Y-m-d H:i:s')
 	{
-		return $this->getAsset()->getUpdatedAt($format);
+		return $this->getAssetNoFilter()->getUpdatedAt($format);
```
MosheMaorKaltura: What about the case that $this->aAsset is null and there is no getCaptionAssetId()?
Please check that getAssetNoFilter has returned a value.

PR #5131, opened 2017-01-10
fix log message.

jessp01:


MosheMaorKaltura:
Please fix indentation

jessp01:
Fixed. Though the problem was the two lines I DID NOT touch were improperly indented...

PR #5151, opened 2017-01-15
looping on partner with id instead of offset

david-winder-kaltura:
PLAT-6712 #time 1h

### DIFF
--- a/alpha/apps/kaltura/lib/batch/myBatchPartnerUsage.class.php
+++ b/alpha/apps/kaltura/lib/batch/myBatchPartnerUsage.class.php
@@ -11,8 +11,8 @@ class myBatchPartnerUsage extends myBatchBase
 	{
 		self::initDb();
 		$partners_exists = true;
-		$start_pos = 0;
 		$bulk_size = 500;
+		$highest_partner_id = -100;
 		while($partners_exists)
 		{
 			$c = new Criteria();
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/lib/batch/myBatchPartnerUsage.class.php
```diff
@@ -11,8 +11,8 @@ public function __construct($partnerId = null, $partnerPackage = 1)
 	{
 		self::initDb();
 		$partners_exists = true;
-		$start_pos = 0;
 		$bulk_size = 500;
+		$highest_partner_id = -100;
```
MosheMaorKaltura: No need to run on partner lower then 100

alpha/apps/kaltura/lib/batch/myBatchPartnerUsage.class.php
```diff
@@ -45,11 +47,11 @@ public function __construct($partnerId = null, $partnerPackage = 1)
 					{
 						myPartnerUtils::doMonthlyPartnerUsage($partner);
 					}
+					$highest_partner_id = max($highest_partner_id, $partner->getId());
```
MosheMaorKaltura: No need  to check every iterate , we can rely on foreach order

alpha/apps/kaltura/lib/batch/myBatchPartnerUsage.class.php
```diff
@@ -45,11 +47,11 @@ public function __construct($partnerId = null, $partnerPackage = 1)
...
 				}
 			}
```
MosheMaorKaltura: Take the last items of the array partners using end() to get the last ID.

alpha/apps/kaltura/lib/batch/myBatchPartnerUsage.class.php
```diff
@@ -45,11 +47,11 @@ public function __construct($partnerId = null, $partnerPackage = 1)
 					{
```
MosheMaorKaltura: Use const from PartnerPackages to check package type

PR #5157, opened 2017-01-17
no need to create new file_sync or increment version of metadata if actual data hasn't changed

noam-arad:
…ctual data hasn't changed

### DIFF
--- a/plugins/metadata/MetadataPlugin.php
+++ b/plugins/metadata/MetadataPlugin.php
@@ -33,6 +33,21 @@ class MetadataPlugin extends KalturaPlugin implements IKalturaVersion, IKalturaP
 	
 	const BULK_UPLOAD_DATE_FORMAT = '%Y-%m-%dT%H:%i:%s';
 
+	/**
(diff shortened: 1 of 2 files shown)


plugins/metadata/MetadataPlugin.php
```diff
@@ -679,9 +684,14 @@ protected static function addMetadataWithProfilesSystemNames (BaseObject $object
     		KalturaLog::info("Metadata [" . $dbMetadata->getId() . "] saved [$xmlData]");
     		
     		$key = $dbMetadata->getSyncKey(Metadata::FILE_SYNC_METADATA_DATA);
-    		kFileSyncUtils::file_put_contents($key, $xmlData);
-    		
-		    kEventsManager::raiseEvent(new kObjectDataChangedEvent($dbMetadata));
+    		if (!kFileSyncUtils::compareContent($key, $xmlData))
```
MosheMaorKaltura: Bettern to move this section into a new function - private saveMetadataFile( $dbMetadata,$xmlData) and use it in the 2 locations

PR #5160, opened 2017-01-18
whenever an entry is cloned it's whole family will be cloned, parent …

noam-arad:
…and all the children of the parent


kaltura-hooks:
Hi @noam-arad,

PR #5170, opened 2017-01-22
Create new recored entry after 7 days

yossipapi:


api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -14,6 +14,9 @@ class KalturaLiveEntryService extends KalturaEntryService
 
 	//amount of time for holding kLock
 	const KLOCK_CREATE_RECORDED_ENTRY_HOLD_TIMEOUT = 3;
+	
+	//Max time from recording created time before creating new recorded entry
+	const SEVEN_DAYS_IN_SECONDS = 604800;
```
kalturaguy: better = 7*24*60*60

yossipapi: The enum name is SEVEN_DAYS_IN_SECONDS so it explains exactly what the value is

PR #5189, opened 2017-01-25
When creating an entry from a template copy also pending categories

NadavHarnik:
PLAT-6832


NadavHarnik:
reviewed by @MosheMaorKaltura

PR #5209, opened 2017-02-01
Watermarking is not working with entry replacement

yossipapi:


plugins/content/caption/base/lib/kMultiCaptionFlowManager.php
```diff
@@ -39,4 +39,49 @@ private function updatedParseMulti($dbBatchJob,$data)
...
+		KalturaLog::debug("Original entry id $originalEntryId");
+		KalturaLog::debug("Replacing entry id [{$entry->getId()}]");
+		
+		$captions = assetPeer::retrieveByEntryId($originalEntryId, array(CaptionPlugin::getAssetTypeCoreValue(CaptionAssetType::CAPTION)));
+		foreach($captions as $caption)
+		{
+			/* @var asset $caption */
+			$caption->copyToEntry($entry->getId());
+		}
+	}
```
MosheMaorKaltura: Should return something here, or remove the rerun true if original entry not found

plugins/transcoding/watermark/lib/kWatermarkFlowManager.php
```diff
@@ -0,0 +1,55 @@
...
+		KalturaLog::debug("Replacing entry id [{$entry->getId()}]");
+		
+		$watermarkMetadata = kWatermarkManager::getWatermarkMetadata($originalEntry);
+		if(!$watermarkMetadata)
+		{
+			KalturaLog::debug("Watermark data not found for entry [$originalEntryId]");
+			return true;
+		}
+		
+		kWatermarkManager::copyWatermarkData($watermarkMetadata, $originalEntry, $entry);
```
MosheMaorKaltura: Same here , must return value.

yossipapi:
@MosheMaorKaltura Done.

PR #5230, opened 2017-02-05
fix code so that we actually verify new partner creation happens only…

noam-arad:
… when authorized

### DIFF
--- a/api_v3/services/PartnerService.php
+++ b/api_v3/services/PartnerService.php
@@ -76,13 +76,12 @@ class PartnerService extends KalturaBaseService
 			$partner_registration = new myPartnerRegistration ( $parentPartnerId );
 			
 			$ignorePassword = false;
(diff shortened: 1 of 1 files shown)


api_v3/services/PartnerService.php
```diff
@@ -76,13 +76,12 @@ public function registerAction( KalturaPartner $partner , $cmsPassword = "" , $t
 			$partner_registration = new myPartnerRegistration ( $parentPartnerId );
 			
 			$ignorePassword = false;
-			if ($existingUser && ($this->getKs()->partner_id == Partner::ADMIN_CONSOLE_PARTNER_ID || $this->getKs()->partner_id == $parentPartnerId)){
```
noam-arad: had to be removed since this was always true

tan-tan-kanarek: It was true because it was always zero? this part of the condition used for VAR console, then it won't necessarily will be true.

noam-arad: $parentPartnerId = $this->getKs()->partner_id;
so there is no need to compare that to itself

PR #5246, opened 2017-02-08
push notification redesign + enable register caching

yossipapi:


plugins/event_notification/providers/push/lib/model/PushNotificationTemplate.php
```diff
@@ -85,9 +105,16 @@ public function getQueueKey($contentParameters, $partnerId = null, kScope $scope
         // sort array according to created keys
         ksort($contentParametersValues);
         
-        $contentParamsHash = md5($partnerId . '_' . implode( '_' , array_values($contentParametersValues) ) );
-        // prepare queue key to return
-        return 'pn_' . $templateId . '_' . $contentParamsHash;
+        $queueConentParams = $partnerId . '_' . implode( '_' , array_values($contentParametersValues));
```
etameran: Probably meant $queueContentParams

plugins/event_notification/providers/push/services/PushNotificationTemplateService.php
```diff
@@ -7,96 +7,126 @@
...
-	    return $result;
+		/* @var $kPushNotificationParams kPushNotificationParams */
+		$missingParams = array();
+		$userContnetParams = array();
+		$userQueueKeyParams = array();
+		$userParamsArrayKeys = array();
+		$userParamsArray = $pushNotificationParams->toObject()->getUserParams();
+		
+		// get template configured params
+		$contnetParams = $dbEventNotificationTemplate->getContentParametersKeyValueArray();
```
etameran: content instead of contnet

api_v3/services/ResponseProfileService.php
```diff
@@ -26,6 +26,10 @@ protected function partnerGroup($peer = null)
 		{
 			case 'get':
 				return $this->partnerGroup . ',0';
+			//When requesting response profiles allow defualt once in case requesting partner is internal
```
etameran: default

plugins/event_notification/providers/push/lib/model/PushNotificationTemplate.php
```diff
@@ -85,9 +105,16 @@ public function getQueueKey($contentParameters, $partnerId = null, kScope $scope
         // sort array according to created keys
         ksort($contentParametersValues);
         
-        $contentParamsHash = md5($partnerId . '_' . implode( '_' , array_values($contentParametersValues) ) );
-        // prepare queue key to return
-        return 'pn_' . $templateId . '_' . $contentParamsHash;
+        $queueContentParams = $partnerId . '_' . implode( '_' , array_values($contentParametersValues));
+        if($returnRaw)
+        	return 'pn_' . $templateId . '_md5s_' . $queueContentParams . "_md5e";
```
tan-tan-kanarek: I don't think that "md5" should be part of the returned string.

plugins/event_notification/providers/push/services/PushNotificationTemplateService.php
```diff
@@ -7,96 +7,109 @@
...
-	    // build the url to return
-	    $protocol = infraRequestUtils::getProtocol();
-	    $host = kConf::get("push_server_host");
-	    $secret = kConf::get("push_server_secret");
-	    $ip = kCurrentContext::$user_ip;
-	    $token = base64_encode($partnerId . ":" . $this->encode($secret . ":" . $ip . ":" . $hash . ":" . uniqid() ) );
-	    $result->url = $protocol . "://" . $host ."/?p=" . $partnerId ."&x=" . urlencode($token);
-	    
-	    return $result;
+		/* @var $kPushNotificationParams kPushNotificationParams */
```
tan-tan-kanarek: $kPushNotificationParams is not defined.

plugins/event_notification/providers/push/services/PushNotificationTemplateService.php
```diff
@@ -7,96 +7,109 @@
...
+		$missingParams = array();
+		$userContentParams = array();
+		$userQueueKeyParams = array();
+		$userParamsArrayKeys = array();
+		$userParamsArray = $pushNotificationParams->toObject()->getUserParams();
+		
+		// get template configured params
+		$contentParams = $dbEventNotificationTemplate->getContentParametersKeyValueArray();
+		$queueKeyParams = $dbEventNotificationTemplate->getQueueKeyParametersKeyValueArray();
```
tan-tan-kanarek: If you just merge them? why did you create the new QueueKeyParameters?

plugins/event_notification/providers/push/services/PushNotificationTemplateService.php
```diff
@@ -7,96 +7,109 @@
...
+			}
+		}
+		
+		foreach ($templateParams as $templateParamKey => $templateParamValue)
+		{
+			if (!in_array($templateParamKey, $userParamsArrayKeys))
+				array_push($missingParams, $templateParamKey);
+		}
+		
+		if ($missingParams != null)
```
tan-tan-kanarek: $missingParams is never null, it's either empty or non-empty array.

plugins/event_notification/providers/push/services/PushNotificationTemplateService.php
```diff
@@ -7,96 +7,109 @@
...
+	function sendCommandAction($queueName, KalturaPushNotificationCommandType $command)
+	{
+		$time = time();
+		$msg = json_encode(array(
+				"data" 		=> null,
+				"queueKey" 	=> null,
+				"queueName"	=> $queueName,
+				"msgId"		=> md5("$queueName $time"),
+				"msgTime"	=> $time,
+				"command"	=> $command
```
tan-tan-kanarek: I think that the command should be an object and not an enum, additionally yo the type you might need additional data in the future.

plugins/event_notification/providers/push/services/PushNotificationTemplateService.php
```diff
@@ -7,96 +7,109 @@
...
+	/**
+	 * Clear queue messages 
+	 *
+	 * @action sendCommand
+	 * @actionAlias eventNotification_eventNotificationTemplate.sendComman
+	 * @param string $queueName QueueNAme to clear messages for
+	 * @param KalturaPushNotificationCommandType $command Command to be sent to push server
+	 * @return bool Allwyas true
+	 */
+	function sendCommandAction($queueName, KalturaPushNotificationCommandType $command)
```
tan-tan-kanarek: You should use id or system-name of the template, using the plain queue name by-passes any security and partners default criterias.

plugins/event_notification/lib/kEventNotificationFlowManager.php
```diff
@@ -151,15 +151,7 @@ public function shouldConsumeEvent(KalturaEvent $event)
 			
 			$scope->resetDynamicValues();
 			
-			$notificationParameters = $notificationTemplate->getContentParameters();
-			foreach($notificationParameters as $notificationParameter)
-			{
-				/* @var $notificationParameter kEventNotificationParameter */
-				if(!is_null($notificationParameter->getValue()))
-					$scope->addDynamicValue($notificationParameter->getKey(), $notificationParameter->getValue());
```
yossipapi: I am not skipping the setting of the dynamic values I am simply getting both arrays (getContentParameters & getUserParameters) together and than executing the foreach on both them instead of one by one.

Do you mean that the order is important i.e I first should add run on contentParameters and only than on userParameters?

plugins/event_notification/providers/push/lib/api/KalturaPushNotificationParams.php
```diff
@@ -0,0 +1,36 @@
+<?php
+/**
+ * Object which contains contextual entry-related data.
+ * @package plugins.pushNotification
+ * @subpackage api.objects
+ */
+class KalturaPushNotificationParams extends KalturaObject
```
yossipapi: @tan-tan-kanarek Before this was added in the testMe this value was represented as a string wrapping it with an  object makes the test me more friendly work with.

plugins/event_notification/providers/push/lib/model/PushNotificationTemplate.php
yossipapi: @tan-tan-kanarek  Changed it to use prefix & postfix similar to what we do when indexing to sphinx

plugins/event_notification/lib/kEventNotificationFlowManager.php
tan-tan-kanarek: I'm not sure if that feature ever used, but it could be used.

plugins/event_notification/providers/push/lib/api/KalturaPushNotificationParams.php
tan-tan-kanarek: Testme is not a good reason to modify our API.

plugins/event_notification/lib/kEventNotificationFlowManager.php
yossipapi: Changed so that each template type will apply its own logic for when applying dynamic values to scope

PR #5255, opened 2017-02-13
throw more specific excepction

noam-arad:


noam-arad:
original https://github.com/kaltura/server/pull/5248

PR #5263, opened 2017-02-16
Drop from name group flavors that are not included in the transcode.xml

gadyaari:


plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -152,15 +152,16 @@ public function serveAction($streamName, $hostname = null, $extraParams = null)
...
-			$this->appendLiveParams($entry, $mediaServer, $encodes, $liveParamsItem, $streamParametersArray);
-			$tags = array("all");
-			foreach($tags as $tag)
-			{
-				if(!isset($groups[$tag]))
-					$groups[$tag] = array();
-
-				$systemName = $liveParamsItem->getSystemName() ? $liveParamsItem->getSystemName() : $liveParamsItem->getId();
-				$groups[$tag][] = $systemName;
+			$flavorExist = $this->appendLiveParams($entry, $mediaServer, $encodes, $liveParamsItem, $streamParametersArray);
```
kalturaguy: rename to shouldAddFlavor

plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -152,15 +152,16 @@ public function serveAction($streamName, $hostname = null, $extraParams = null)
...
-			foreach($tags as $tag)
-			{
-				if(!isset($groups[$tag]))
-					$groups[$tag] = array();
-
-				$systemName = $liveParamsItem->getSystemName() ? $liveParamsItem->getSystemName() : $liveParamsItem->getId();
-				$groups[$tag][] = $systemName;
+			$shouldAddFlavor = $this->appendLiveParams($entry, $mediaServer, $encodes, $liveParamsItem, $streamParametersArray);
+			if ($shouldAddFlavor) {
+				$tags = array("all");
```
MosheMaorKaltura: The 'all' value should be either a const value for this class , or a config value if there is an option for future change.

PR #5271, opened 2017-02-20
push notification redesign + enable register caching

yossipapi:


plugins/event_notification/providers/push/lib/model/PushNotificationTemplate.php
```diff
@@ -112,18 +158,32 @@ protected function getMessage(kScope $scope)
...
-        $queueKey = $this->getQueueKey($contentParameters, null, $scope);
-        
+        $queueNameParameters = $this->getQueueNameParameters();
+        $queueKeyParameters = $this->getQueueKeyParameters();
+        $queueName = $this->getQueueName($queueNameParameters, null, $scope);
+        $queueKey = $this->getQueueKey($queueKeyParameters, null, $scope);
+        $message = $this->getMessage($scope);
+        $time = time();
+
+        $msg = json_encode(array(
```
yossipapi: Need to change so that an object will be sent

PR #5308, opened 2017-03-01
enable new batch with * tag and add to batch configuration template

david-winder-kaltura:


MosheMaorKaltura:
Reviewed , 
1. Fix file plugins/drop_folder/lib/kDropFolderAllocator.php , so it will not show as changed in the commit.

PR #5319, opened 2017-03-05
check if https apiProtocol is set for request when setting protocol

inbal-ben-david:
SUP-9836

### DIFF
--- a/alpha/apps/kaltura/lib/myPartnerUtils.class.php
+++ b/alpha/apps/kaltura/lib/myPartnerUtils.class.php
@@ -344,7 +344,7 @@ class myPartnerUtils
 	public static function getCdnHost ( $partner_id, $protocol = null, $hostType = null )
 	{
 		// in case the request came through https, force https url
-		if (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on')
+		if (infraRequestUtils::getProtocol() == 'https')
 			$protocol = 'https';
 
 		// temporary default is http since the system is not aligned to use https in all of its components (e.g. kmc)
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/lib/myPartnerUtils.class.php
```diff
@@ -344,7 +344,7 @@ public static function getHost ( $partner_id )
 	public static function getCdnHost ( $partner_id, $protocol = null, $hostType = null )
 	{
 		// in case the request came through https, force https url
-		if (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on')
+		if (infraRequestUtils::getProtocol() == 'https')
 			$protocol = 'https';
```
MosheMaorKaltura: Why not use the direct value coming out of infraRequestUtils::getProtocol()?

inbal-ben-david: I think that if you do so you can sometimes miss the protocol value given as arg for the function.
What if you will get as argument to the call $protocol=https and inside the function you will get infraRequestUtils::getProtocol() == 'http', if this case is possible you will override the given value for the protocol.

I also thought we can call to infraRequestUtils::getProtocol() only if the given protocol=null but currently we are doing the first check in getProtocol() even if the value is null and then forcing https so we will lose it in this case.  

There are multiple flows that uses this function so I think it would be less risky to use it that way and basically just add the apiProtocol check.

MosheMaorKaltura: After reviewing all the cases , I can say that - 
1. In most of them value of null is passed , which will be handled as http by infraRequestUtils::getProtocol()
2. In the cases where protocol is being sent, it is either http or https. In first place the argument protocol should have been enum of 2 values.
Bottom line - lets do the full change , review again all the code that sends none null value in protocol argument to getCdnHost and we will watch this case in compat check.

PR #5346, opened 2017-03-15
Added retry on couchbase upsert and get command for response profile …

gotlieb:
…caching in case of a couchbase exception.

SUP-9991

@MosheMaorKaltura

### DIFF
--- a/infra/cache/kCouchbaseCacheWrapper.php
+++ b/infra/cache/kCouchbaseCacheWrapper.php
@@ -479,35 +479,29 @@ class kCouchbaseCacheWrapper extends kBaseCacheWrapper
 	/* (non-PHPdoc)
 	 * @see kBaseCacheWrapper::doGet()
 	 */
-	protected function doGet($key, $retries = 1)
+	protected function doGet($key, $attempts = 2)
 	{
-		try
+		$couchbaseError = null;
+		while ($attempts > 0)
 		{
-			$meta = $this->bucket->get($key);
-			
-			if($this->debug)
-				KalturaLog::debug("key [$key], meta [" . print_r($meta, true) . "]");
-				
-			return $meta->value;
-		}
-		catch(CouchbaseException $e)
-		{
-			if($e->getCode() == self::ERROR_CODE_THE_KEY_DOES_NOT_EXIST_IN_THE_SERVER)
-				return false;
-
-			else
+			try
 			{
-				if ($retries > 0 )
-				{
-					KalturaLog::debug("Retrying Couchbase get operation.");
-					return $this->doGet($key, $retries - 1);
-				}
-				else{
-					KalturaLog::debug("No retries left for Couchbase get operation.");
-					throw $e;
-				}
+				$meta = $this->bucket->get($key);
+
+				if ($this->debug)
+					KalturaLog::debug("key [$key], meta [" . print_r($meta, true) . "]");
+
+				return $meta->value;
+			} catch (CouchbaseException $e)
+			{
+				$couchbaseError = $e;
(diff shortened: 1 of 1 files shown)


infra/cache/kCouchbaseCacheWrapper.php
```diff
@@ -479,24 +479,29 @@ protected function doInit($config)
...
+				return $meta->value;
+			} catch (CouchbaseException $e)
+			{
+				$couchbaseError = $e;
+				if ($e->getCode() == self::ERROR_CODE_THE_KEY_DOES_NOT_EXIST_IN_THE_SERVER)
+					return false;
+			}
+			$attempts--;
 		}
+		KalturaLog::err("No retries left for Couchbase get operation.");
```
MosheMaorKaltura: In case of failure print the key the we wanted to get and we failed to.

infra/cache/kCouchbaseCacheWrapper.php
```diff
@@ -479,24 +479,29 @@ protected function doInit($config)
...
+
+				if ($this->debug)
+					KalturaLog::debug("key [$key], meta [" . print_r($meta, true) . "]");
+
+				return $meta->value;
+			} catch (CouchbaseException $e)
+			{
+				$couchbaseError = $e;
+				if ($e->getCode() == self::ERROR_CODE_THE_KEY_DOES_NOT_EXIST_IN_THE_SERVER)
+					return false;
```
MosheMaorKaltura: KalturaLog::debug about retry, so we can learn about system behavior.

infra/cache/kCouchbaseCacheWrapper.php
```diff
@@ -479,24 +479,29 @@ protected function doInit($config)
...
-			$meta = $this->bucket->get($key);
-			
-			if($this->debug)
-				KalturaLog::debug("key [$key], meta [" . print_r($meta, true) . "]");
-				
-			return $meta->value;
-		}
-		catch(CouchbaseException $e)
+		$couchbaseError = null;
+		while ($attempts > 0)
```
MosheMaorKaltura: Change to do-while, since $attempts can be called with value Zero.

infra/cache/kCouchbaseCacheWrapper.php
```diff
@@ -548,16 +553,29 @@ protected function doIncrement($key, $delta = 1)
...
 			KalturaLog::debug("Bucket name [$this->name] key [$key] var [" . print_r($var, true) . "]");
-			
-		$meta = $this->bucket->upsert($key, $var, array(
-			'expiry' => $expiry
-		));
-		
-		return is_null($meta->error);
+
+		while ($attempts > 0)
+		{
```
MosheMaorKaltura: Change to do-while, since $attempts can be called with value Zero.

infra/cache/kCouchbaseCacheWrapper.php
```diff
@@ -548,16 +553,29 @@ protected function doIncrement($key, $delta = 1)
...
+			try
+			{
+				$meta = $this->bucket->upsert($key, $var, array(
+					'expiry' => $expiry
+				));
+
+				return is_null($meta->error);
+			} catch (CouchbaseException $e)
+			{
+				KalturaLog::err($e);
```
MosheMaorKaltura: In this case we should only retry in case of timeout exception.

infra/cache/kCouchbaseCacheWrapper.php
```diff
@@ -548,16 +553,29 @@ protected function doIncrement($key, $delta = 1)
...
+			}
+			$attempts--;
+		}
+
+		KalturaLog::debug("No retries left for Couchbase upsert operation.");
+		return false;
```
MosheMaorKaltura: Until today , upsert has thrown unhandled exception ,  we need to keep this behavior here.

PR #5357, opened 2017-03-19
Lynx 12.13.0 plat 7167

MosheMaorKaltura:


api_v3/lib/types/partner/KalturaPartner.php
```diff
@@ -293,7 +293,7 @@ class KalturaPartner extends KalturaObject implements IFilterable
 	
 	/**
 	 * @var int
-	 * @readonly
+	 * @insertonly
```
tan-tan-kanarek: Both insertonly and readonly are not enforced in partner.register.

PR #5367, opened 2017-03-21
set duration of live recording while recording is still being done

noam-arad:
Should be merged to next branch

### DIFF
--- a/api_v3/lib/types/entry/KalturaPlayableEntry.php
+++ b/api_v3/lib/types/entry/KalturaPlayableEntry.php
@@ -106,4 +106,18 @@ class KalturaPlayableEntry extends KalturaBaseEntry
 			
 		return $dbObject;
 	}
+
+	public function doFromObject($sourceObject, KalturaDetachedResponseProfile $responseProfile = null)
+	{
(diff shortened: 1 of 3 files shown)


api_v3/lib/types/entry/KalturaPlayableEntry.php
```diff
@@ -106,4 +106,18 @@ public function toObject($dbObject = null, $skip = array())
 			
 		return $dbObject;
 	}
+
+	public function doFromObject($sourceObject, KalturaDetachedResponseProfile $responseProfile = null)
+	{
+		parent::doFromObject($sourceObject, $responseProfile);
+		$recordedLengthInMs = $sourceObject->getRecordedLengthInMsecs();
+		if ($sourceObject->getIsRecordedEntry() && $recordedLengthInMs > 0)
```
yossipapi: I think this condition should also include && if ! !$sourceObject->getLengthInMsecs()
I see that every time setLengthInMsecs is called this field value is reseted to 0 and once the entry replacement will happen and the entry will get it duration wont the API return duration of 0 in the response?

PR #5371, opened 2017-03-22
Rename 09.MetadataProfile.Polls.ini to 09.MetadataProfilePolls.ini

DBezemer:
Rename is needed for regex to match the file in $APP_DIR/deployment/base/scripts/insertDefaults.php


yossipapi:
Looks right. will merge to next server version.

yossipapi:
On second thought this wont work the format must translate to list($order, $objectType, $fileExtension) = explode('.', $fileName, 3);

PR #5392, opened 2017-03-30
PLAT-7167:allow only admin_console to insert and update parent-partner-id and p…

noam-arad:
…artner-package


MosheMaorKaltura:
Reviewed

PR #5437, opened 2017-04-12
SUP-10845: fix delivery profile ids will be set to none if during sto…

noam-arad:
…rage profile update they are not given


MosheMaorKaltura:
Reviewed.

PR #5460, opened 2017-04-23
PS-2993: syndication-feeds - allow relative dynamic filter values

shpiNi:


api_v3/services/SyndicationFeedService.php
```diff
@@ -134,6 +134,8 @@ public function getAction($id)
 	 * @param KalturaBaseSyndicationFeed $syndicationFeed
 	 * @return KalturaBaseSyndicationFeed
 	 * @throws KalturaErrors::INVALID_FEED_ID
+	 * 
+	 * @disableRelativeTime $syndicationFeed
```
rkreich: What about `add` action?

api_v3/lib/syndication/KalturaSyndicationFeedRenderer.php
```diff
@@ -167,6 +169,16 @@ public function __construct($feedId, $feedProcessingKey = null, $ks = null, $sta
 	
 		if($this->addLinkForNextIteration)
 		{
+			$entryFilter = $this->syndicationFeed->entryFilter;
+			foreach(self::$relativeTimeFieldNames as $relativeTimeFieldName)
```
rkreich: Can't we do this in `entryFilter` so we could re-use `entryFilter::$relative_time_fields`?

PR #5481, opened 2017-04-26
add metadataProfileId for MR

david-winder-kaltura:


plugins/scheduled_task/admin/MediaRepurposingUtils.php
```diff
@@ -162,16 +162,19 @@ private static function getMrMetadataProfile()
 		$filter = new Kaltura_Client_Metadata_Type_MetadataProfileFilter();
 		$filter->systemNameEqual = 'MRP';
 		$metadataPlugin = self::getPluginByName('Kaltura_Client_Metadata_Plugin');
-		return $metadataPlugin->metadataProfile->listAction($filter, null);
+		$profiles = $metadataPlugin->metadataProfile->listAction($filter, null);
+		return $profiles->objects[0];
```
noam-arad: why is only the first one enough, we might have many with system name MRP?
what if none are returned?

PR #5495, opened 2017-04-27
prevent creation of duplicate kuser when the puser is long

inbal-ben-david:
SUP-10928

### DIFF
--- a/alpha/lib/model/kuserPeer.php
+++ b/alpha/lib/model/kuserPeer.php
@@ -45,6 +45,9 @@ class kuserPeer extends BasekuserPeer implements IRelatedObjectPeer
 	 */
 	public static function getKuserByPartnerAndUid($partnerId, $puserId, $ignorePuserKuser = false)
 	{
+		if(strlen($puserId) > 100)
+			$puserId = substr($puserId, 0, 100);
+
 		if(!$ignorePuserKuser && !kCurrentContext::isApiV3Context())
 		{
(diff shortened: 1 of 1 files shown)


alpha/lib/model/kuserPeer.php
```diff
@@ -45,6 +45,9 @@ public static function getKuserByScreenName( $screen_name  )
 	 */
 	public static function getKuserByPartnerAndUid($partnerId, $puserId, $ignorePuserKuser = false)
 	{
+		if(strlen($puserId) > 100)
```
MosheMaorKaltura: 100 should be a const value defined in this class and not hard coded.

alpha/lib/model/kuserPeer.php
```diff
@@ -89,6 +92,8 @@ public static function getActiveKuserByPartnerAndUid($partner_id , $puser_id)
 	
 	public static function createKuserForPartner($partner_id, $puser_id, $is_admin = false)
 	{
+		if(strlen($puser_id) > 100)
+			$puser_id = substr($puser_id, 0, 100);
```
MosheMaorKaltura: Move the 2 line logic to a static private function - 
getvalidPuserStr() 
Later, we can use this function to add more validation rules ( if we will be required) but in one place.

MosheMaorKaltura:
Description of the pull request needs to include the JIRA ID

PR #5509, opened 2017-05-05
Apply URANDOM to PlayReady form as it slow down the backend add/configure

kimdelmote:
Very slow form loading due to the mcrypt_create_iv function.


MosheMaorKaltura:
Thank you @kimdelmote  for this code change , we have merged it into a newer branch.
Will be glad to see more from you.

Moshe.

PR #5535, opened 2017-05-10
SUP-10970: filter temprorary entries from event notifications so they…

noam-arad:
… don't get distributed since these will be deleted in the near future

### DIFF
--- a/plugins/content_distribution/lib/model/EntryDistribution.php
+++ b/plugins/content_distribution/lib/model/EntryDistribution.php
@@ -479,7 +479,7 @@ class EntryDistribution extends BaseEntryDistribution implements IIndexable, ISy
 
 			$outFlavors[] = $asset->getId();
 		}
-		return $outFlavors;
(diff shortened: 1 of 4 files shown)


plugins/content_distribution/lib/model/EntryDistribution.php
```diff
@@ -435,5 +435,51 @@ public function getCacheInvalidationKeys()
...
+		{
+			/**
+			 * @var flavorAsset $asset
+			 */
+			if ($asset && $asset->getIsOriginal() && $asset->getStatus() == flavorAsset::ASSET_STATUS_TEMP)
+				continue;
+			if (in_array($asset->getFlavorParamsId(), $tempFlavorsParamsIds))
+				continue;
+
+			$outFlavors[] = $asset->getId();
```
MosheMaorKaltura: You need to verify here also that $asset is not empty as in 475, try to consolidate it.

PR #5554, opened 2017-05-14
SUP-10928: add a check for null puser_id

inbal-ben-david:


moshemery:
Reviewed - please merge

PR #5628, opened 2017-06-01
Default parameter assignment is not supported in IE

DBezemer:
Default assignment will be supported from ECMA2015 and onwards, which IE does not (yet) use
http://www.ecma-international.org/ecma-262/6.0/#sec-functiondeclarationinstantiation

This currently breaks the Admin Console functions in IE

### DIFF
--- a/admin_console/views/scripts/partner/list.phtml
+++ b/admin_console/views/scripts/partner/list.phtml
@@ -462,10 +462,10 @@ function assignLiveDeliveryProfile(format, currentDps)
 	);
 }
 
-function assignDeliveryProfile(format, currentDps, type = null)
+function assignDeliveryProfile(format, currentDps, type)
 {
+	var name = name === undefined? null: name;
 	var partnerId = $("#id")[0].value;
-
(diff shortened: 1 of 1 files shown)


admin_console/views/scripts/partner/list.phtml
```diff
@@ -462,10 +462,10 @@ function assignLiveDeliveryProfile(format, currentDps)
 	);
 }
 
-function assignDeliveryProfile(format, currentDps, type = null)
+function assignDeliveryProfile(format, currentDps, type)
 {
+	var name = name === undefined? null: name;
```
yossipapi: This var is not used so why was it added?
Does name = type?

DBezemer:
@yossipapi brain fart, obviously name has to be type, changed it

admin_console/views/scripts/partner/list.phtml
DBezemer: Yes, changed the commit

DBezemer:
@yossipapi can you please confirm / merge? Without this Admin Console does not work in IE. As it's now branch 12.18 let me know if you want me to reopen a pull against that branch

PR #5633, opened 2017-06-02
Lynx 12.17.0 ps 3079

hilak:


hilak:
@assafb I've amended the code, adding the enforcement of the kuserId on the KalturaViewHistoryUserEntryFilter.

PR #5673, opened 2017-06-11
Lynx 12.18.0 plat 7531

MosheMaorKaltura:


plugins/poll/lib/PollActions.php
```diff
@@ -25,134 +24,197 @@ class PollActions
...
+		return $validUserId;
+	}
+
+	public function setVote($pollId, $userId, $ksUserId, $ansIds)
+	{
+		if ($this->isValidPollIdStructure($pollId))
+		{
+			$pollType = $this->getPollType($pollId);
+			//validate User ID
+			$str = $userId." ".$ksUserId;
```
yossipapi: Remove $str not needed

plugins/poll/lib/PollActions.php
```diff
@@ -25,134 +24,197 @@ class PollActions
...
+				$this->pollsCacheHandler->incrementPollVotersCount($pollId);
+			$this->pollsCacheHandler->incrementAnswersCounter($pollId, $answers);
+			return "Successfully voted";
+		}
+		return "Failed to vote due to bad poll id structure";
+	}
+
+
+
+	private static function getUserIdFromKs($params)
```
yossipapi: This is not longer used

plugins/poll/services/PollService.php
```diff
@@ -98,7 +98,9 @@ public function voteAction($pollId, $userId, $answerIds)
 		KalturaResponseCacher::disableCache();
 		try
 		{
-			return PollActions::setVote($pollId, $userId, $answerIds);
+			$pollActions = new PollActions();
+			$ksUserId = kCurrentContext::getCurrentKsKuserId();
```
yossipapi: This will return the kuserid and not the puser id 
Need to use kCurrentContext::$uid

plugins/poll/services/PollService.php
```diff
@@ -116,7 +118,8 @@ public function voteAction($pollId, $userId, $answerIds)
 	public function getVoteAction($pollId, $userId)
 	{
 		KalturaResponseCacher::disableCache();
-		return PollActions::getVote($pollId, $userId);
+		$pollActions = new PollActions();
+		return $pollActions->doGetVote($pollId, $userId);
```
yossipapi: doGetVote get params and not attributes

PR #5682, opened 2017-06-13
SUP-11306: handle ampersand chars in XML document

MosheMaorKaltura:


infra/general/KDOMDocument.php
```diff
@@ -16,7 +16,15 @@ public function load ( $filename , $options = 0 )
 		if(!file_exists($filename) || !filesize($filename))
 			throw new Exception('Empty file supplied as input');
 			
-		return parent::loadXML(file_get_contents($filename), $options);
+		return $this->loadXML(file_get_contents($filename), $options);
+	}
+
+	public function loadXML ( $source, $options = null)
+	{
+		$regex = '/(\s\&\s)/i';
```
tan-tan-kanarek: It's too specific, it won't catch R&D, && and other use cases.

PR #5692, opened 2017-06-15
SUP-11263: Don't try to generate source flavor since it can only be ingested

yossipapi:


alpha/apps/kaltura/lib/batch2/bcdl/kBusinessPreConvertDL.php
```diff
@@ -423,6 +423,13 @@ public static function decideAddEntryFlavor(BatchJob $parentJob = null, $entryId
 			KalturaLog::err("Flavor Params Id [$flavorParamsId] not found");
 			return null;
 		}
+		
+		//Check if the flavor we are trying to generate has source flavor, source can only be ingested, not generated.
+		if(in_array(assetParams::TAG_SOURCE, $flavorParams->getTagsArray()))
+		{
+			KalturaLog::err("Cannot generate flavor params Id [$flavorParamsId], flavor has source tag");
```
MosheMaorKaltura: Using kalturaLog::err is wrong here, since it generate exception traceback message , and it is not required.
I would change it to -  warning or notice.

PR #5756, opened 2017-06-27
Override filter limit from advanced filter

hilak:


plugins/content/view_history/lib/model/filters/kViewHistoryUserEntryAdvancedFilter.php
```diff
@@ -68,11 +68,21 @@ public function applyCondition(IKalturaDbQuery $query)
 		$limit = $query->getLimit() ? $query->getLimit() : ($this->filterLimit ? $this->filterLimit : self::ENTRIES_COUNT);
 		
 		$query->setLimit(self::ENTRIES_COUNT);
+		if ($this->filterLimit)
+		{
+			KalturaLog::info ("Overriding filter limit to " . self::ENTRIES_COUNT);
+			$this->overrideFilterLimit = self::ENTRIES_COUNT;
```
MosheMaorKaltura: $this->overrideFilterLimit   will be override in line 84 , what was the purpose of this line?

shpiNi: @MosheMaorKaltura the intention here is to use a constant hard-coded value for the first doSelect query and then re-set it to the actual limit for the next doSelect done internally in UserEntryPeer::getEntryIdsByFilter on line 96. it's done the same way for $query->setLimit() in lines 70, 81.

PR #5814, opened 2017-07-12
PLAT-7669: add addContent action to Data service

inbal-ben-david:


api_v3/services/DataService.php
```diff
@@ -187,4 +187,52 @@ function serveAction($entryId, $version = -1, $forceProxy = false)
...
+	* @return string
+	* @throws KalturaErrors::ENTRY_ID_NOT_FOUND
+	* @validateUser entry entryId edit
+	*/
+	function addContentAction($entryId, KalturaDataEntryAllowedContentResource $resource)
+	{
+		$dbEntry = entryPeer::retrieveByPK($entryId);
+
+		if (!$dbEntry || $dbEntry->getType() != KalturaEntryType::DATA)
+			throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $entryId);
```
MosheMaorKaltura: Change error to INVALID_ENTRY_TYPE in case of wrong type

api_v3/lib/types/resource/KalturaUploadedFileTokenResource.php
```diff
@@ -7,7 +7,7 @@
  * @see api/services/UploadService#uploadAction()
  * @see api/services/UploadTokenService#addAction()
  */
-class KalturaUploadedFileTokenResource extends KalturaDataCenterContentResource
+class KalturaUploadedFileTokenResource extends KalturaDataEntryAllowedContentResource
```
MosheMaorKaltura: Rename KalturaDataEntryAllowedContentResource to KalturaGenericDataCenterContentResource

api_v3/services/DataService.php
```diff
@@ -187,4 +187,52 @@ function serveAction($entryId, $version = -1, $forceProxy = false)
...
+
+		if($resource) {
```
MosheMaorKaltura: NO need to check if resource , since it is from dedicated type and cannot be null

api_v3/services/DataService.php
```diff
@@ -187,4 +187,52 @@ function serveAction($entryId, $version = -1, $forceProxy = false)
...
+			$kResource = $resource->toObject();
+			$resource->validateEntry($dbEntry);
```
MosheMaorKaltura: Validate first before toObject.

PR #5825, opened 2017-07-17
Add info to mail task

david-winder-kaltura:


plugins/scheduled_task/lib/object_task_engine/KObjectTaskMailNotificationEngine.php
```diff
@@ -0,0 +1,87 @@
...
+		foreach ($toArray as $to)
+			$mailer->AddAddress($to);
+		$mailer->Subject = $subject;
+		$mailer->Body = $body;
+
+		$mailer->From = '';
+		$mailer->FromName = $sender;
+
+		KalturaLog::info("sending mail to " . implode(",",$toArray) . " with body: $body");
+		return $mailer->Send();
```
MosheMaorKaltura: Catch the mailer exception and print the ->message

PR #5837, opened 2017-07-18
PLAT-7663: shouldn't use recorderLengthInMsecs if entry is no longer …

noam-arad:
…served from live


MosheMaorKaltura:
reviewed

PR #5862, opened 2017-07-25
SUP-11537:adding support for mo files in file type restriction mode

ZurKaltura:


alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -162,14 +162,33 @@ protected function checkIfFileIsAllowed()
...
 
 		$fileTypes = kConf::get('file_type');
 		return in_array($fileType, $fileTypes['allowed']);
 	}
 
 	/**
+	* Try to find the file type by running the file cmd and match the output to a pattern
+	 * It will return empty string if no pattern was matched
+	*/
+	private function findFileTypeByFileCmd($filePath)
```
yossipapi: Better pass this to kFile rather than having the function in this class.

alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -162,14 +162,33 @@ protected function checkIfFileIsAllowed()
...
+	{
+		$fileType = '';
+		$fileBrief = shell_exec('file -b ' . $filePath);
+		$moPattern = "GNU message catalog";
```
yossipapi: Define strong as const

etameran:
Please use realpath to avoid malicious use of the exec


$filePath = "aa ; echo 123";
echo shell_exec('file -b ' . $filePath);

vs

$realPath = realpath($filePath);
if ($realPath)
    echo shell_exec('file -b ' . realpath($filePath));


________________________________

PR #5866, opened 2017-07-25
No Plat: indexIdGreaterThan should refer to the entry int_id

yossipapi:


yossipapi:
Opened: https://github.com/kaltura/server/pull/5869 instead.

PR #5867, opened 2017-07-26
PLAT-7821: When fetching multiple entrries need to reset deliveryAttributes between entry iterations

yossipapi:


MosheMaorKaltura:
Reviewed

PR #5869, opened 2017-07-26
PLAT-7822: indexIdGreaterThan should refer to the entry int_id in case of entry

yossipapi:


yossipapi:
Created instead of: https://github.com/kaltura/server/pull/5866

MosheMaorKaltura:
Reviewed.

PR #5870, opened 2017-07-26
PLAT-7823 add params validation and partner filters to userEntry filter

gotlieb:
@MosheMaorKaltura 
@erankor

### DIFF
--- a/api_v3/lib/KalturaErrors.php
+++ b/api_v3/lib/KalturaErrors.php
@@ -655,6 +655,7 @@ class KalturaErrors extends APIErrors
 	const MUST_FILTER_ON_ENTRY_OR_USER = 'MUST_FILTER_ON_ENTRY_OR_USER;;Must filter on entry ID or user ID';
 	const USER_ENTRY_FILTER_FORBIDDEN_FIELDS_USED = 'USER_ENTRY_FILTER_FORBIDDEN_FIELDS_USED;;UserEntry filter object forbidden fields used';
 	const USER_ENTRY_ALREADY_EXISTS = 'USER_ENTRY_ALREADY_EXISTS;;UserEntry for this type already exists';
+	const USER_ENTRY_FILTER_MISSSING_MANDATORY_PARAMS = 'USER_ENTRY_FILTER_MISSSING_MANDATORY_PARAMS;;One of the following parameters must be supplied: userIdEqual, userIdIn, entryIdEqual, entryIdIn ';
 	
 	/*
 	 * serverNode service
--- a/api_v3/services/UserEntryService.php
+++ b/api_v3/services/UserEntryService.php
@@ -94,10 +94,6 @@ class UserEntryService extends KalturaBaseService {
 	 */
 	public function listAction(KalturaUserEntryFilter $filter, KalturaFilterPager $pager = null)
 	{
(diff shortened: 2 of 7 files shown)


api_v3/lib/KalturaErrors.php
```diff
@@ -655,6 +655,7 @@ class KalturaErrors extends APIErrors
 	const MUST_FILTER_ON_ENTRY_OR_USER = 'MUST_FILTER_ON_ENTRY_OR_USER;;Must filter on entry ID or user ID';
 	const USER_ENTRY_FILTER_FORBIDDEN_FIELDS_USED = 'USER_ENTRY_FILTER_FORBIDDEN_FIELDS_USED;;UserEntry filter object forbidden fields used';
 	const USER_ENTRY_ALREADY_EXISTS = 'USER_ENTRY_ALREADY_EXISTS;;UserEntry for this type already exists';
+	const USER_ENTRY_FILTER_MISSSING_MANDATORY_PARAMS = 'USER_ENTRY_FILTER_MISSSING_MANDATORY_PARAMS;;One of the following parameters must be supplied: userIdEqual, userIdIn, entryIdEqual, entryIdIn ';
```
yossipapi: There is already a genetic error for missing manadatory parameter no need to add anew one

api_v3/services/UserEntryService.php
```diff
@@ -94,10 +95,6 @@ public function deleteAction($id)
 	 */
 	public function listAction(KalturaUserEntryFilter $filter, KalturaFilterPager $pager = null)
```
yossipapi: Why force the passing of the filter?
We should stick to the standards where we allow null in the function signature and create an empty filter if one was not passed.

I suggest reverting the if statement deletion and adding null defaul value tot the filter.

MosheMaorKaltura: Filter must be forced since the whole idea was to avoid cases where no userId or entryid is specified.

api_v3/lib/KalturaErrors.php
MosheMaorKaltura: But using it the user will not know the rules that just on param of the following list is required.

api_v3/services/UserEntryService.php
yossipapi: But you are already forcing this in the filter itself so if they are not provided you throw an error

MosheMaorKaltura: Question - can there be a case that filter not supplied to the listAction? if not why allow it

etameran:
We don't have specific response for other similar lists actions with missing filter params . Additionally did you try it and actually got your specific exception?

Sent from my iPhone

On 27 Jul 2017, at 7:41, Moshe Maor <notifications@github.com<mailto:notifications@github.com>> wrote:


@MosheMaorKaltura commented on this pull request.

________________________________

In api_v3/services/UserEntryService.php<https://github.com/kaltura/server/pull/5870#discussion_r129752426>:

> @@ -94,10 +95,6 @@ public function deleteAction($id)
         */
        public function listAction(KalturaUserEntryFilter $filter, KalturaFilterPager $pager = null)


Question - can there be a case that filter not supplied to the listAction? if not why allow it

PR #5872, opened 2017-07-28
Update LiveStreamService.php

hilak:


api_v3/services/LiveStreamService.php
```diff
@@ -102,13 +115,29 @@ protected function prepareEntryForInsert(KalturaBaseEntry $entry, entry $dbEntry
...
 					$dbEntry->setConversionProfileId($partner->getDefaultLiveConversionProfileId());
 			}
-				
-			$dbEntry->save();
 		}
 		
 		return $dbEntry;
 	}
 	
+	protected function getTemplateEntry($conversionProfileId, $templateEntryId)
```
MosheMaorKaltura: Why not reuse getTemplateEntry from KalturaEntryService its a 95% same code?

hilak: @MosheMaorKaltura  the getTemplateEntry on KalturaEntryService tries to retrieve the default vod conversion profile from the partner in case $conversionProfileId is null. This is not the desired behavior here since I want to use the template entry ID on the live conversion profile. Furthermore, I only need to access the live conversion profile when the entry is of specific source type. Please see @erankor  code review on PLAT-7836

PR #5885, opened 2017-08-01
Mercury 13.1.0 elastic

NadavHarnik:


alpha/lib/model/KuserKgroup.php
```diff
@@ -56,4 +56,29 @@ public function getCacheInvalidationKeys()
...
+	{
+		parent::postUpdate($con);
+
+		if (!$this->alreadyInSave)
+			$this->updateKuserIndex();
+	}
+
+	protected function updateKuserIndex()
+	{
+		$kuserId = $this->getKuserId();
```
MosheMaorKaltura: You can call $this->getkuserRelatedByKuserId();

PR #5896, opened 2017-08-03
PLAT-7827: recorded entry should be created for backup as well as pri…

noam-arad:
…mary but only one per live entry so it has to be locked


noam-arad:
should not be merged without relevant changes in liveController

PR #5929, opened 2017-08-14
Update unlockJobsByScheduler.php

gotlieb:
modified unlock batch jobs script - remove getting id from outside. retrieving id from db according to host name

@MosheMaorKaltura  @etameran  @elizil  please review

### DIFF
--- a/alpha/lib/model/SchedulerPeer.php
+++ b/alpha/lib/model/SchedulerPeer.php
@@ -29,13 +29,13 @@ class SchedulerPeer extends BaseSchedulerPeer
 		$c->add(SchedulerPeer::HOST, $hostname);
 		$result = SchedulerPeer::doSelect( $c);
 		if (!$result)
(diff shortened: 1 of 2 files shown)


alpha/lib/model/SchedulerPeer.php
```diff
@@ -22,4 +22,21 @@ public static function deleteBySchedulerConfigId($schedulerId)
...
+		$c->add(SchedulerPeer::HOST, $hostname);
+		$result = SchedulerPeer::doSelect( $c);
+		if (!$result)
+			throw new kCoreException("Could not find scheduler for host $hostname");
+
+		if ( count($result)> 1 )
+			throw new kCoreException("More than one result for host $hostname");
+
+		if (!$result[0]->getConfiguredId())
+			throw new kCoreException("Could not find scheduler id for scheduler.");
```
MosheMaorKaltura: The exception is not meaningful , add hostname and id for the error message.

PR #5937, opened 2017-08-15
PLAT-7906: Add php7 support for accessing sftp urls

yossipapi:
PHP bug details can be seen here: https://bugs.php.net/bug.php?id=73597


MosheMaorKaltura:
Reviewed

PR #5943, opened 2017-08-15
Plat 7489 live dashboard integration

atarsh:


atarsh:
need to add relevant version (live_dashboard_version) in base.ini , where liveanalytics_version is set.

PR #5950, opened 2017-08-17
PLAT-7883: in cielo24 flow manager when no remote content - set error…

shpiNi:
… status on all transcripts


MosheMaorKaltura:
reviewed

PR #5970, opened 2017-08-26
TR-1682

hilak:


plugins/integration/providers/voicebase/lib/KVoicebaseIntegrationEngine.class.php
```diff
@@ -26,6 +26,7 @@ public function close(KalturaBatchJob $job, KalturaIntegrationJobData &$data)
 	
 	protected function doDispatch(KalturaBatchJob $job, KalturaIntegrationJobData &$data, KalturaVoicebaseJobProviderData $providerData)
 	{
+		KalturaLog::info ();
```
rkreich: Can be removed

plugins/integration/providers/voicebase/lib/KVoicebaseIntegrationEngine.class.php
```diff
@@ -69,6 +70,16 @@ protected function doDispatch(KalturaBatchJob $job, KalturaIntegrationJobData &$
 	
 	protected function doClose(KalturaBatchJob $job, KalturaIntegrationJobData &$data, KalturaVoicebaseJobProviderData $providerData)
 	{
+		$entryId = $providerData->entryId;
+		$this->clientHelper = VoicebasePlugin::getClientHelper($providerData->apiKey, $providerData->apiPassword);
+		$remoteProcess = $this->clientHelper->retrieveRemoteProcess($entryId);
+		
+		//false result means that something has gone wrong - the VB job is either in status error or missing altogether
+		if(!$remoteProcess || $remoteProcess->requestStatus == VoicebaseClientHelper::VOICEBASE_FAILURE_MEESAGE || !isset($remoteProcess->fileStatus) || $curlResult->fileStatus == VoicebaseClientHelper::VOICE_MACHINE_FAILURE_MESSAGE)
```
rkreich: Do we know for sure that `$remoteProcess->fileStatus` will always be set?
Because it currently checks for `!isset($remoteProcess->fileStatus)` and then assumes that there's an error

plugins/integration/providers/voicebase/lib/KVoicebaseIntegrationEngine.class.php
```diff
@@ -69,6 +70,16 @@ protected function doDispatch(KalturaBatchJob $job, KalturaIntegrationJobData &$
...
+		{
+			throw new Exception("VoiceBase transcription failed.");
+		}
+		
 		return false;
```
rkreich: We should also support returning true in case processing finished (for cases where notification is not received from VoiceBase)

plugins/integration/providers/voicebase/VoicebaseClientHelper.php
```diff
@@ -94,7 +104,7 @@ private function sendAPICall($url, $options = null, $noDecoding = false)
 		if(!$noDecoding)
 		{
 			$stringResult = $result;
-			$result = json_decode($result);
+			$result = json_decode($result, true);
```
rkreich: Why changing to assoc arrays?

hilak: @rkreich if the json object has depth (or changes to contain depth) the json_decode will fail.

plugins/integration/providers/voicebase/lib/KVoicebaseIntegrationEngine.class.php
hilak: @rkreich it has not happened in my testing. After calling mediaUpload for the entry ID, you have a file in "processing" status. During testing, I never had a job fail after upload (unless the video format was impossible for VB to handle (i.e .arf)

hilak:
@rkreich  please review changes.

PR #5974, opened 2017-08-27
Add uiConfId for LiveDashboard on kmc.vars

gadyaari:


alpha/apps/kaltura/modules/kmc/lib/kmcUtils.class.php
```diff
@@ -114,10 +114,10 @@ public static function getPartnersUiconfs($partner_id, $tag)
 		return $confs;
 	}
 
-	public static function getLiveAUiconf()
+	public static function getLiveUiconf($uiConfTag)
 	{
 		$c = new Criteria();
-		$tag = 'livea_player';
+		$tag = $uiConfTag;
```
yossipapi: Use $uiConfTag as is why do you define another variable for it.

alpha/apps/kaltura/modules/kmc/lib/kmcUtils.class.php
```diff
@@ -114,10 +114,10 @@ public static function getPartnersUiconfs($partner_id, $tag)
 		return $confs;
 	}
 
-	public static function getLiveAUiconf()
+	public static function getLiveUiconf($uiConfTag)
```
yossipapi: I would change the name from getLiveUiconf to getLiveUiconfByTag

gadyaari: ok

alpha/apps/kaltura/modules/kmc/lib/kmcUtils.class.php
gadyaari: I don't really know. didn't actually think about it

PR #5975, opened 2017-08-27
PLAT-7939: Prevent adding user entry without user id + prevent changing entryId on existing user entry object

yossipapi:


api_v3/lib/KalturaErrors.php
```diff
@@ -655,6 +655,8 @@ class KalturaErrors extends APIErrors
 	const MUST_FILTER_ON_ENTRY_OR_USER = 'MUST_FILTER_ON_ENTRY_OR_USER;;Must filter on entry ID or user ID';
 	const USER_ENTRY_FILTER_FORBIDDEN_FIELDS_USED = 'USER_ENTRY_FILTER_FORBIDDEN_FIELDS_USED;;UserEntry filter object forbidden fields used';
 	const USER_ENTRY_ALREADY_EXISTS = 'USER_ENTRY_ALREADY_EXISTS;;UserEntry for this type already exists';
+	const USER_ID_NOT_PROVIDED = 'USER_ID_NOT_PROVIDED;;User ID not found neither on the object or KS';
+	const USER_ENTRY_DOES_NOT_MATCH_USER_ID = 'USER_ENTRY_DOES_NOT_MATCH_USER_ID;USER_ENTRY_ID;The userID on the user-entry [@USER_ENTRY_ID@] does not match the entry-id given';
```
hilak: @yossipapi please check the error code. From the use of this error on KalturaUserEntry::validateEntryId, it seems the wording is off.

api_v3/lib/types/KalturaUserEntry.php
```diff
@@ -159,4 +159,68 @@ protected function doFromObject($srcObj, KalturaDetachedResponseProfile $respons
...
 
+	public function validateForInsert($propertiesToSkip = array())
+	{
+		parent::validateForInsert($propertiesToSkip);
+		$this->validatePropertyNotNull("entryId");
+		$this->validateEntryId();
+		$this->validateUserID();
+	}
+	
+	public function validateForUpdate($sourceObject, $propertiesToSkip = array())
```
hilak: No need for this method to be public. Used internally.

api_v3/lib/types/KalturaUserEntry.php
```diff
@@ -159,4 +159,68 @@ protected function doFromObject($srcObj, KalturaDetachedResponseProfile $respons
...
+
+		$this->validateUserID($sourceObject->getId());
+		return parent::validateForUpdate($sourceObject, $propertiesToSkip);
+	}
+	
+	/*
+	 * @param string $userEntryID
+	 * @throw KalturaAPIException
+	 */
+	private function validateEntryId($userEntryID = null)
```
hilak: No need for this method to be private - we may want to override on KalturaUserEntry child classes, better make if protected.

api_v3/lib/types/KalturaUserEntry.php
```diff
@@ -159,4 +159,68 @@ protected function doFromObject($srcObj, KalturaDetachedResponseProfile $respons
...
+			if($this->entryId !== null && $this->entryId != $dbUserEntry->getEntryId())
+				throw new KalturaAPIException(KalturaErrors::USER_ENTRY_DOES_NOT_MATCH_ENTRY_ID, $userEntryID);
+		}
+	}
+	
+	/*
+	 * @param string $userEntryID
+	 * @throw KalturaAPIException
+	 */
+	private function validateUserId($userEntryID = null)
```
hilak: No need for this method to be private - we may want to override on KalturaUserEntry child classes, better make if protected.

api_v3/lib/types/KalturaUserEntry.php
```diff
@@ -159,4 +159,68 @@ protected function doFromObject($srcObj, KalturaDetachedResponseProfile $respons
 		parent::doFromObject($srcObj, $responseProfile);
 	}
 
+	public function validateForInsert($propertiesToSkip = array())
+	{
+		parent::validateForInsert($propertiesToSkip);
+		$this->validatePropertyNotNull("entryId");
+		$this->validateEntryId();
```
hilak: @yossipapi I don't see why you need all this. There is an "@insertonly" on both the entryId and userId properties of the KalturaUserEntry object, which should make both these properties non-updatable.

api_v3/lib/types/KalturaUserEntry.php
```diff
@@ -159,4 +159,68 @@ protected function doFromObject($srcObj, KalturaDetachedResponseProfile $respons
...
+		$this->validateUserID();
+	}
+	
+	public function validateForUpdate($sourceObject, $propertiesToSkip = array())
+	{
+		if($this->entryId !== null)
+			$this->validateEntryId($sourceObject->getId());
```
hilak: @yossipapi  I don't see why you need to validate entryId or userId, there is an "@insertOnly" tag in the KalturaUserEntry object on both of these properties which prevents their update.

api_v3/lib/types/KalturaUserEntry.php
```diff
@@ -159,4 +159,68 @@ protected function doFromObject($srcObj, KalturaDetachedResponseProfile $respons
...
+	 */
+	private function validateUserId($userEntryID = null)
+	{
+		if($userEntryID === null)
+		{
+			$userId = $this->userId ? $this->userId : kCurrentContext::getCurrentKsKuserId();
+			if(!$userId)
+				throw new KalturaAPIException(KalturaErrors::USER_ID_NOT_PROVIDED);
+		}
+		else
```
hilak: @yossipapi don't think you need this else block. userId property on the KalturaUserEntry object is "@insertonly" and cannot be updated.

api_v3/lib/types/KalturaUserEntry.php
```diff
@@ -159,4 +159,68 @@ protected function doFromObject($srcObj, KalturaDetachedResponseProfile $respons
...
+		{
+			// update
+			$dbUserEntry = UserEntryPeer::retrieveByPK($userEntryID);
```
hilak: @yossipapi don't think you need this else block. entryId property on the KalturaUserEntry object is "@insertonly" and cannot be updated.

api_v3/lib/types/KalturaUserEntry.php
yossipapi: I did not notice the part they are insert only I will remove the update validation in this case.
This comment applies on the other related comments as well.

PR #5994, opened 2017-09-03
PLAT-7827: add entry server node recording status

noam-arad:


alpha/lib/enums/EntryServerNodeRecordingStatus.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * @package Core
+ * @subpackage model.enum
+ */ 
+interface EntryServerNodeRecordingStatus extends BaseEnum
+{
+	const STOPPED = 0;
```
yossipapi: I don't see the use case of using the STOPPED enum value.
If there is please explain.

PR #6020, opened 2017-09-07
PLAT-7952: Add user login by ks action allow change account operation

yossipapi:


api_v3/services/UserService.php
```diff
@@ -595,6 +595,17 @@ function indexAction($id, $shouldUpdate = true)
...
-
+	
+	/**
+	 * Loges a user to the destination account as long the ks user id exists in the desc acount and the loginData id match for both accounts
+	 *
+	 * @action loginByKs
+	 * @param int $destPartnerId
+	 * @throws APIErrors::PARTNER_CHANGE_ACCOUNT_DISABLED
+	 */
+	public function loginByKsAction($destPartnerId)
```
tan-tan-kanarek: You didn't specify the returned type but I understand it's string, I think we would like to avoid new actions that return primitive types if possible.

PR #6037, opened 2017-09-11
PLAT-7988: add caption support for webvtt format on clip/trim

inbal-ben-david:
also contains fixes for PLAT-7958 and PLAT-7967


plugins/content/caption/base/lib/kCaptionsContentManager.php
```diff
@@ -21,7 +68,26 @@ public function __construct()
...
+	 */
+	public abstract function buildFile($content, $clipStartTime, $clipEndTime);
+
+	/**
+	 * @param array $matches
+	 * @param int $clipStartTime
+	 * @param int $clipEndTime
+	 * @return string
+	 */
+	public function createAdjustedTimeLine($matches, $clipStartTime, $clipEndTime)
```
MosheMaorKaltura: Change function to pure abstract and protected since no access from the outside.

plugins/content/caption/base/lib/kCaptionsContentManager.php
```diff
@@ -89,4 +155,110 @@ public static function addParseMultiLanguageCaptionAssetJob($captionAsset, $file
...
+			return $ret;
+		if(!is_numeric($parts[1]))
+			return null;
+
+		$ret += floatval($parts[1]) * 1000;  // seconds im milliseconds
+		return round($ret);
+	}
+
+
+	public function createCaptionsFile($content, $clipStartTime, $clipEndTime, $timeCode)
```
MosheMaorKaltura: This function is being called as if it is static , since it access the instance, change all the callers to use it as non-static

plugins/content/caption/base/lib/kCaptionsContentManager.php
```diff
@@ -5,11 +5,58 @@
...
+		);
+		$contentArray = explode(self::UNIX_LINE_ENDING, $textContent); // Create array from text content
+		return $contentArray;
+	}
+
+	/**
+	 * @param array $array
+	 * @return mixed
+	 */
+	public static function getNextValueFromArray(array &$array)
```
MosheMaorKaltura: No use to access from the outside - make it protected.

PR #6057, opened 2017-09-14
Mercury 13.2.0 plat 7951

hilak:
@MosheMaorKaltura  please review.


alpha/apps/kaltura/lib/kEntitlementUtils.php
```diff
@@ -132,9 +132,9 @@ public static function isEntryEntitled(entry $entry, $kuserId = null)
 			}
 
 			// kuser is set on the entry entitled users edit or publish
```
MosheMaorKaltura: Update remark

alpha/lib/model/entry.php
```diff
@@ -1951,11 +1951,53 @@ public function getEntitledKusersEditArray()
...
 	{
 		return implode(',', $this->getEntitledUserPuserEditArray());
 	}
 	
+	public function setEntitledPusersView($v)
+	{
+		$entitledUserPuserView = array();
+		
+		$v = trim($v);
+		if($v == '')
```
MosheMaorKaltura: This block is not really needed , since a case of empty array will result the same in the foreach loop below.

alpha/lib/model/entry.php
```diff
@@ -1951,11 +1951,53 @@ public function getEntitledKusersEditArray()
...
+				
+		foreach ($entitledPusersView as $puserId)
+		{
+			$puserId = trim($puserId);
+			if ( $puserId === '' )
+			{
+				continue;
+			}
+
+			$partnerId = kCurrentContext::$partner_id ? kCurrentContext::$partner_id : kCurrentContext::$ks_partner_id;
```
MosheMaorKaltura: Better to access $partnerId only once outside the loop.

alpha/lib/model/entry.php
```diff
@@ -1951,11 +1951,53 @@ public function getEntitledKusersEditArray()
...
+			$kuser = kuserPeer::getActiveKuserByPartnerAndUid($partnerId, $puserId);
+			if (!$kuser)
```
MosheMaorKaltura: Do we really want to break the chain if one user is not defined?

alpha/lib/model/entry.php
```diff
@@ -1979,6 +2021,20 @@ public function isEntitledKuserEdit($kuserId)
 
 		return $this->isOwnerActionsAllowed($kuserId);
 	}
+	
+	public function isEntitledKuserView($kuserId)
+	{
+		$entitledKuserArray = array_keys($this->getEntitledPusersViewArray());
```
MosheMaorKaltura: Consolidate the common code of this function and isEntitledKuserEdit since most of it is the same.

alpha/lib/model/entry.php
hilak: @MosheMaorKaltura  this block deals with the case of setting the string to ''. It is needed.

alpha/lib/model/entry.php
hilak: @MosheMaorKaltura  this behavior is consistent with the way we set co-editors and co-publishers. I don't see a reason this should behave differently.

MosheMaorKaltura:
Reviewed , please make sure you merge it to 13.4.0.

PR #6106, opened 2017-09-26
PLAT-7961:Unite getTempFileWithContent function to kFile

david-winder-kaltura:


infra/storage/kFile.class.php
```diff
@@ -722,6 +722,26 @@ public static function safeFilePutContents($filePath, $var, $mode=null)
...
+
+	/**
+	 * Lazy saving of file content to a temporary path, the file will exist in this location until the temp files are purged
+	 * @param string $fileContent
+	 * @param string $prefix
+	 * @param integer $permission
+	 * @param boolean $strict - check if to write file without content
+	 * @return string path to temporary file location
+	 */
+	public static function getTempFileWithContent($fileContent, $prefix = '' , $permission = null, $strict = false )
```
yossipapi: From what I can see from the changes it would be mush better to set strict to true and pass false only when you don't want to validate the file content.

infra/storage/kFile.class.php
```diff
@@ -722,6 +722,23 @@ public static function safeFilePutContents($filePath, $var, $mode=null)
...
+	 * @param string $fileContent
+	 * @param string $prefix
+	 * @param integer $permission
+	 * @return string path to temporary file location
+	 */
+	public static function createTempFile($fileContent, $prefix = '' , $permission = null)
+	{
+		$tempDirectory = sys_get_temp_dir();
+		$fileLocation = tempnam($tempDirectory, $prefix);
+		file_put_contents($fileLocation, $fileContent);
```
MosheMaorKaltura: kFile::file_put_contents

infra/storage/kFile.class.php
```diff
@@ -722,6 +722,23 @@ public static function safeFilePutContents($filePath, $var, $mode=null)
...
+		if ($permission)
+			chmod($fileLocation, $permission);
```
MosheMaorKaltura: All file system access function that are touch , only via kFile.
in this case kFile::chmod

PR #6116, opened 2017-09-27
PLAT7988-return only the time and not an array when parsing caption t…

inbal-ben-david:
…imes

### DIFF
--- a/plugins/content/caption/base/lib/kCaptionsContentManager.php
+++ b/plugins/content/caption/base/lib/kCaptionsContentManager.php
@@ -213,10 +213,8 @@ abstract class kCaptionsContentManager
 	 */
 	public function parseCaptionTime($time)
 	{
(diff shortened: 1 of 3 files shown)


plugins/content/caption/base/lib/kCaptionsContentManager.php
```diff
@@ -207,6 +207,18 @@ public static function parseStrTTTime($timeStr)
...
 	}
 
+	/**
+	 * @param $time
+	 * @return string
+	 */
+	public function parseCaptionTime($time)
+	{
+		if($this instanceof srtCaptionsContentManager)
+			$time = str_replace(',','.',$time);
```
MosheMaorKaltura: Either fix the time string before calling this function , or override this entire function in srt

MosheMaorKaltura:
approved

PR #6117, opened 2017-09-27
SUP-12069 - add url domain to uploadToken->add in order to explicitly…

gotlieb:
… determine where to send the uploadToken->upload call

@erankor 
@MosheMaorKaltura


MosheMaorKaltura:
reviewed.

PR #6125, opened 2017-09-28
Mercury 13.4.0 PLAT 6597 apply 6 sec default chunk duration

lilachmaliniak:


alpha/lib/model/Partner.php
```diff
@@ -1806,7 +1806,11 @@ public function setHtmlPurifierBaseListUsage($v)
 	
 	public function getDefaultLiveStreamSegmentDuration()
 	{
-		return $this->getFromCustomData( "default_live_stream_segment_duration", null , null );
+		if (PermissionPeer::isValidForPartner(PermissionName::FEATURE_DYNAMIC_SEGMENT_DURATION, $this->getId()))
+		{
+			return $this->getFromCustomData("default_live_stream_segment_duration", null, null);
```
MosheMaorKaltura: What if value is not set already, better supply default to the get function

MosheMaorKaltura:
Approved

alpha/lib/model/Partner.php
lilachmaliniak: done added default value to getFromCustomData()

PR #6130, opened 2017-10-01
Mercury 13.4.0 plat 7986

inbal-ben-david:


api_v3/services/MediaService.php
```diff
@@ -1182,6 +1182,67 @@ private function isClipTrimFlow($resource)
...
+		if(!$supportedFlavor)
+			throw new KalturaAPIException(KalturaErrors::SUPPORTED_FLAVOR_NOT_EXIST, $entryId);
+		$supportedFlavorId = $supportedFlavor->getId();
+
+		$packagerRetries = 3;
+		$content = null;
+		while ($packagerRetries && !$content)
+		{
+			$content = $this->retrieveLocalVolumeMapFromPackager($supportedFlavor);
+			if(!$content)
```
MosheMaorKaltura: This check is not needed , only need to decrease counter

api_v3/services/MediaService.php
```diff
@@ -1182,6 +1182,67 @@ private function isClipTrimFlow($resource)
...
+			throw new KalturaAPIException(KalturaErrors::RETRIEVE_VOLUME_MAP_FAILED, $entryId);
+
+		header("Content-Disposition: attachment; filename=".$entryId.'_'.$supportedFlavorId."_volumeMap.csv");
+		return new kRendererString($content, 'text/csv');
+	}
+
+	private function retrieveLocalVolumeMapFromPackager($flavorAsset)
+	{
+		$packagerVolumeMapUrlPattern = kConf::get('packager_local_volume_map_url', 'local', null);
+		if (!$packagerVolumeMapUrlPattern)
```
MosheMaorKaltura: If the parameter is not defined, better throw a relevant exception then return null , because the result will be an exception saying - RETRIEVE_VOLUME_MAP_FAILED

alpha/lib/model/assetPeer.php
```diff
@@ -590,6 +604,42 @@ public static function retrieveHighestBitrateByEntryId($entryId, $tag = null, $e
 	}
 
 	/**
+	 * if $retrieveHighestBitrate is set to true we will retrieve true if flavotAsset bitrate is higher then ret bitrate
+	 * if $retrieveHighestBitrate is set to false we will retrieve true if flavotAsset bitrate is lower then ret bitrate
+	 *
+	 */
+	public static function replaceSavedFlavorWithBetterBitrateFlavor($flavorAsset, $ret, $retrieveHighestBitrate)
```
MosheMaorKaltura: 1. Better rename replaceSavedFlavorWithBetterBitrateFlavor to replaceFlavorWithBetterBitrateFlavor 
2. What is $ret? is it a type of flavorAsset? please rename to a more meaningful name 
3. If you assume that $flavorAsset and$ret are type of flavorAsset , better state the type in the function definition to protect against null input
4. Is this function really needs to be public?

PR #6152, opened 2017-10-10
PLAT-7961:call fullMkfileDir in the sftp distribution engine and add …

david-winder-kaltura:
…documentation

### DIFF
--- a/plugins/content_distribution/batch/Distribute/engines/SftpDistributionEngine.php
+++ b/plugins/content_distribution/batch/Distribute/engines/SftpDistributionEngine.php
@@ -30,7 +30,7 @@ abstract class SftpDistributionEngine extends DistributionEngine
 	{
 		$tempFilePath = $this->tempDirectory . '/' . $this->getTempDirectory() . '/' . $distributionProfileId . '/';
 		if (!file_exists($tempFilePath))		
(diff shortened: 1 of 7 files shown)


plugins/content_distribution/batch/Distribute/engines/SftpDistributionEngine.php
```diff
@@ -30,7 +30,7 @@ protected function getTempDirectoryForProfile($distributionProfileId)
 	{
 		$tempFilePath = $this->tempDirectory . '/' . $this->getTempDirectory() . '/' . $distributionProfileId . '/';
 		if (!file_exists($tempFilePath))		
-			mkdir($tempFilePath, 0777, true);
+			kFile::fullMkfileDir($tempFilePath);
```
yossipapi: I reverted this change last weeek since it caused issues. Please note that the function you are calling requires the full file path to create the directory for. In your case you are passing only the directory path.

MosheMaorKaltura: kFile::fullMkfileDir expects to get directory as input.

MosheMaorKaltura:
reviewed

PR #6153, opened 2017-10-10
Plat 8089- facebook place is deprecated and updating tags

ZurKaltura:
Ignoring place when we do create and update.
change the way we handle tags to support tagging pepole

### DIFF
--- a/plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
+++ b/plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
@@ -32,14 +32,28 @@ class FacebookDistributionEngine extends DistributionEngine implements
 	public function submit(KalturaDistributionSubmitJobData $data)
 	{
 		$this->validate($data);
-		if ($data->entryDistribution->remoteId) {
+		if ($data->entryDistribution->remoteId)
+		{
 			$data->remoteId = $data->entryDistribution->remoteId;
-		} else {
+		}
+		else
+		{
 			$this->doSubmit($data);
 		}
 		return true;
 	}
 
+	/** on submission of a video we can take care of the following :
+	 * 1. video content
+	 * 2. thumbnail
+	 * 3. call to action
+	 * 4. name + description
+	 * 5. place
+	 * 6. tags
+	 * 7. targeting
+	 * 8. scheduled_publishing_time
+	 * 9. feed targeting
+	 */
 	protected function doSubmit(KalturaDistributionSubmitJobData $data)
 	{
 		$videoPath = $data->providerData->videoAssetFilePath;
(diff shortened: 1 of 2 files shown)


plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
```diff
@@ -113,22 +123,52 @@ protected function doUpdate(KalturaDistributionUpdateJobData $data)
...
+	{
+		try
 		{
-			foreach ($data->providerData->captionsInfo as $captionInfo)
+			// first delete the captions that were already distributed
+			while ($mediaFile = array_pop($data->mediaFiles))
 			{
-				/* @var $captionInfo KalturaFacebookCaptionDistributionInfo */
-				$data->mediaFiles[] = $this->submitCaption($data->distributionProfile, $captionInfo, $data->entryDistribution->remoteId);
+				$this->deleteCaption($data->distributionProfile, $mediaFile->remoteId, $data->entryDistribution->remoteId);
```
MosheMaorKaltura: Calling this function may throw exception.
Please handle it.

plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
```diff
@@ -72,18 +76,24 @@ protected function doSubmit(KalturaDistributionSubmitJobData $data)
...
 				$facebookMetadata);
-		} catch (Exception $e) {
+		}
+		catch (Exception $e)
+		{
 			throw new Exception("Failed to submit facebook video , reason:".$e->getMessage());
 		}
 
-		if ($data->providerData->captionsInfo)
+		$tags = FacebookDistributionEngine::helperGetTags($data->providerData->fieldValues);
```
MosheMaorKaltura: Accessing static function in the same class should be made with self:: operator.

plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
```diff
@@ -113,22 +123,52 @@ protected function doUpdate(KalturaDistributionUpdateJobData $data)
 			throw new Exception("Failed to update facebook video , reason:".$e->getMessage());
 		}
 
-		// first delete the captions that were already distributed
-		while ($mediaFile = array_pop($data->mediaFiles))
+		$tags = FacebookDistributionEngine::helperGetTags($data->providerData->fieldValues);
```
MosheMaorKaltura: Same here , use self::

plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
```diff
@@ -113,22 +123,52 @@ protected function doUpdate(KalturaDistributionUpdateJobData $data)
...
-		if(isset($data->providerData->captionsInfo))
+	}
+
+	private function handleCaptions(KalturaDistributionUpdateJobData $data)
+	{
+		try
 		{
-			foreach ($data->providerData->captionsInfo as $captionInfo)
+			// first delete the captions that were already distributed
+			while ($mediaFile = array_pop($data->mediaFiles))
```
MosheMaorKaltura: Why not use the common way of foreach on mediaFiles

plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
ZurKaltura: Its a try catch block

plugins/content_distribution/providers/facebook/lib/FacebookDistributionEngine.php
ZurKaltura: Not my code I just moved it around but I think he used that way in case mediaFiles is null so this way he dont need to check it before calling the foreach

PR #6154, opened 2017-10-10
No Plat: Don't verify SSL host when using KCurlWrapper

yossipapi:


infra/general/KCurlWrapper.class.php
```diff
@@ -225,7 +225,7 @@ public function __construct($params = null) {
 		if(!$params || !isset($params->curlVerifySSL) || !$params->curlVerifySSL)
 		{
 			curl_setopt($this->ch, CURLOPT_SSL_VERIFYPEER, false);
-			curl_setopt($this->ch, CURLOPT_SSL_VERIFYHOST, 1);
```
MosheMaorKaltura: This reduces security. What was the motivation?

1 to check the existence of a common name in the SSL peer certificate. 
2 (default) to check the existence of a common name and also verify that it matches the hostname provided. 
0 to not check the names. 
--

yossipapi:
Already handled.

PR #6168, opened 2017-10-16
PLAT-7224: clone DRM profile script

david-winder-kaltura:


alpha/scripts/utils/drmProfileCopyBetweenAccounts.php
```diff
@@ -0,0 +1,16 @@
+<?php
+if($argc != 3)
+	die ("Usage: drmProfileCopyBetweenAccounts <srcDrmProfileId> <targetPartnerId>" . PHP_EOL);
```
MosheMaorKaltura: change - Usage: drmProfileCopyBetweenAccounts
to : Usage: $argv[0] 
So we can rename script if needed, but the code will remain authentic.

MosheMaorKaltura:
Reviewed

PR #6171, opened 2017-10-16
set plugins in player config object from uiConfig data

tanyaLibatter:


alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -179,6 +179,7 @@ private function getAutoEmbedCode($targetId = null)
 
 		$config["partnerId"] = $this->partnerId;		
 		$config["uiConfId"] = $this->uiconfId;
+		$config["plugins"] = $this->playerPlugins;
```
OrenMe: This needs to be added also to dynamic embed

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -272,6 +273,9 @@ private function initMembers()
 		$uiConf = uiConfPeer::retrieveByPK($this->uiconfId);
 		if (!$uiConf)
 			KExternalErrors::dieError(KExternalErrors::UI_CONF_NOT_FOUND);
+
+		$playerConfig = json_decode($uiConf->getConfig(), true);
```
OrenMe: the uiconf config can be any arbitrary json config, not only plugins config.
Basically you need to assume that getConfig will return the entire config object of the player.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
tanyaLibatter: done

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
tanyaLibatter: done

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -177,8 +188,11 @@ private function getAutoEmbedCode($targetId = null)
 			$config[$key] = json_decode($val);
 		}
 
+
 		$config["partnerId"] = $this->partnerId;		
 		$config["uiConfId"] = $this->uiconfId;
+		$config["uiConf"] = $this->playerConfig;
```
yossipapi: playerConfig is missing in the class params definition.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -272,6 +286,8 @@ private function initMembers()
 		$uiConf = uiConfPeer::retrieveByPK($this->uiconfId);
 		if (!$uiConf)
 			KExternalErrors::dieError(KExternalErrors::UI_CONF_NOT_FOUND);
+
+		$this->playerConfig = json_decode($uiConf->getConfig(), true);
```
yossipapi: playerConfig is missing in the class params definition.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
OrenMe: @tanyaLibatter there's no uiConf in the player code, you need to pass the config as is.
The config option would be same format as player expects it to be, e.g. 
```json
{
   "playback": {...},
   "plugins": {
      "pluginA": {...},
      "pluginB": {...}
   }
}
```

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -108,6 +108,17 @@ private function formatBundleContent($bundleContent)
 		{
 			$bundleContent = $this->getIfarmEmbedCode($bundleContent);
 		}
+		else
+		{
+			$config["uiConf"] = $this->playerConfig;
+			$config = json_encode($config);	
+
+			$kalturaPlayerConfig = "
+			var kalturaPlayerConfig = $config;
```
OrenMe: how do you expect the player to get the default config? we will need to add some relevant code in player that takes this into consideration when it's loaded.
Currently this kalturaPlayerConfig is just lying there and the app needs to refer to it when calling `KalturaPlayer.setup(..., config);`

OrenMe: Also, need to make this config take into consideration a situation where multiple players are on page and we don't want them to run over another

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
OrenMe: Also, need to handle situation where config is passed from client side and the we need to merge accordingly.

PR #6176, opened 2017-10-17
plat 8040: updating google api to latest google API version

ZurKaltura:
Using the google-api-php-client-1.1.2


MosheMaorKaltura:
Reviewed

PR #6186, opened 2017-10-18
PLAT-8140: Support parsing unix regex of FtpRawListOutput where user or group include the . char

yossipapi:


MosheMaorKaltura:
Review

PR #6259, opened 2017-10-30
PSVAMB-459 - Email template to email current session user

rkreich:


rkreich:
@MosheMaorKaltura @erankor - this email notification template can be used to send an email to the user of the current session. It will be used as manual dispatch.

PR #6261, opened 2017-10-31
use /sbin/service rather than /etc/init.d

jessp01:


MosheMaorKaltura:
On my ce install ubu16 this path will not work.
It can do with /usr/sbin/service

jessp01:
When packaging the deb package, this is replaced with /usr/sbin/service as all others here:
https://github.com/kaltura/server/tree/Mercury-13.8.0/configurations/monit/monit.d

See:
https://github.com/kaltura/platform-install-packages/blob/Mercury-13.8.0/deb/kaltura-base/debian/rules#L77

PR #6267, opened 2017-11-01
Mercury 13.5.0 plat 7977

noam-arad:


alpha/apps/kaltura/lib/batch2/kFlowManager.php
```diff
@@ -586,7 +586,12 @@ public function shouldConsumeChangedEvent(BaseObject $object, array $modifiedCol
 			{
 				return true;
 			}
-			
+
+		if ($object instanceof LiveEntry)
```
yossipapi: Why not add this condition:
$object->isCustomDataModified(LiveEntry::CUSTOM_DATA_RECORDING_STATUS)
here as well you are doing it only in "objectChanged function"

alpha/lib/model/LiveEntry.php
```diff
@@ -448,8 +452,16 @@ protected function getMediaServersHostnames()
 	/**
 	 * @return boolean
 	 */
-	public function hasMediaServer($currentDcOnly = false)
+	public function canPlayLive($currentDcOnly = false)
 	{
+		if ($this->getExplicitLive())
+		{
+			$isAdmin = kCurrentContext::$ks_object && kCurrentContext::$ks_object->isAdmin();
+			$isUserAllowedPreview = $this->isEntitledKuserEdit(kCurrentContext::getCurrentKsKuserId());
```
yossipapi: What if the user is the owner don't you wna't to allow this as well.

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2884,4 +2884,17 @@ private static function activateConvertProfileJob($entryId, $localFilePath)
 			return false;
 	}
 
+	public static function handleLiveEntryChanged(LiveStreamEntry $liveEntry)
```
yossipapi: Can you explain the why was this addition needed? 
You are not checking if these values were changed but, only that they exist.

alpha/lib/model/LiveEntry.php
```diff
@@ -448,8 +452,17 @@ protected function getMediaServersHostnames()
...
 	 * @return boolean
 	 */
-	public function hasMediaServer($currentDcOnly = false)
+	public function isCurrentlyLive($currentDcOnly = false)
 	{
+		if ($this->getExplicitLive())
+		{
+			$isAdmin = kCurrentContext::$ks_object && kCurrentContext::$ks_object->isAdmin();
+			$userIsOwner = kCurrentContext::getCurrentKsKuserId() == $this->getKuserId();
+			$isUserAllowedPreview = $this->isEntitledKuserEdit(kCurrentContext::getCurrentKsKuserId());
```
yossipapi: Why not use the field "$entitledUsersView" do we really want to couple between isEntitledKuserEdit to this feature?

alpha/lib/model/LiveEntry.php
```diff
@@ -448,8 +452,17 @@ protected function getMediaServersHostnames()
 	/**
 	 * @return boolean
 	 */
-	public function hasMediaServer($currentDcOnly = false)
+	public function isCurrentlyLive($currentDcOnly = false)
 	{
+		if ($this->getExplicitLive())
```
yossipapi: If $this->getViewMode() == ViewMode:: ALLOW_ALL we should not check this flow but continue directly to checking if the stream is live.

api_v3/lib/KalturaLiveEntryService.php
```diff
@@ -215,9 +216,10 @@ function registerMediaServerAction($entryId, $hostname, $mediaServerIndex, $appl
 		$this->setMediaServerWrapper($dbLiveEntry, $mediaServerIndex, $hostname, $liveEntryStatus, $applicationName);
 
 		// setRedirectEntryId to null in all cases, even for broadcasting...
-		$dbLiveEntry->setRedirectEntryId(null);
-		$dbLiveEntry->save();
+		if ($dbLiveEntry->getViewMode() == ViewMode::ALLOW_ALL)
```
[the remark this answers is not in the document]
noam-arad: This is not the only place we do this, it' s just a validation for every registerMediaServer

PR #6281, opened 2017-11-06
Mercury 13.7.0 playKitJs

tanyaLibatter:
Add env config to be passed to player embed code(service and API urls and stat services URLs)

### DIFF
--- a/alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
+++ b/alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
@@ -69,7 +69,7 @@ class embedPlaykitJsAction extends sfAction
 		$config = str_replace("\"", "'", json_encode($context->bundleConfig));
 		if(!$config)
 		{
-			KExternalErrors::dieError(KExternalErrors::BUNDLE_CREATION_FAILED, $config);
+			KExternalErrors::dieError(KExternalErrors::BUNDLE_CREATION_FAILED, $config . " wrong config object");
 		}
 
 		$url = $context->bundlerUrl . "/build?config=" . base64_encode($config) . "&name=" . $context->bundle_name . "&source=" . base64_encode($context->sourcesPath);
@@ -77,13 +77,13 @@ class embedPlaykitJsAction extends sfAction
 
 		if (!$content) 
 		{
-			KExternalErrors::dieError(KExternalErrors::BUNDLE_CREATION_FAILED, $config);
+			KExternalErrors::dieError(KExternalErrors::BUNDLE_CREATION_FAILED, $config . " failed to get content from bundle builder");
 		}
 
 		$content = json_decode($content, true);
 		if(!$content || !$content['bundle'])
 		{
-			KExternalErrors::dieError(KExternalErrors::BUNDLE_CREATION_FAILED, $config);
+			KExternalErrors::dieError(KExternalErrors::BUNDLE_CREATION_FAILED, $config . " bundle created with wrong content");
 		}
 
 		$sourceMapContent = base64_decode($content['sourceMap']);
--- a/alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
+++ b/alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
@@ -89,6 +89,7 @@ class embedPlaykitJsAction extends sfAction
 		$sourceMapContent = base64_decode($content['sourceMap']);
 		$bundleContent = time() . "," . base64_decode($content['bundle']);
 		$bundleSaved =  $context->bundleCache->set($context->bundle_name, $bundleContent);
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -143,6 +145,66 @@ private function appendUiConfToContent($content)
...
+	{
+
+		$protocol = (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on') ? "https" : "http";
+
+		// The default Kaltura service url:
+		$serviceUrl = requestUtils::getApiCdnHost().'/api_v3';
+		// Default Kaltura CDN url:
+		$cdnUrl = requestUtils::getCdnHost($protocol);
+		// Default Stats URL
+		$statsServiceUrl = ($protocol == "https") ? $this->buildUrl($protocol,"stats_host_https") : $this->buildUrl($protocol,"stats_host");
```
OrenMe: @yossipapi can we also get these URLs(stats services) from an infra util?

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -143,6 +145,66 @@ private function appendUiConfToContent($content)
...
 		return $content;
 	}
+
+	private function appendEnvConfigToContent($content)
+	{
+
+		$protocol = (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on') ? "https" : "http";
+
+		// The default Kaltura service url:
+		$serviceUrl = requestUtils::getApiCdnHost().'/api_v3';
```
OrenMe: @yossipapi the "api_v3" end point, is it configured somewhere so we can also get it from server config, or we need to use it like this as hardcoded string?

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -143,6 +145,66 @@ private function appendUiConfToContent($content)
...
+		// Default Kaltura CDN url:
+		$cdnUrl = requestUtils::getCdnHost($protocol);
```
OrenMe: @yossipapi do you know why `requestUtils::getCdnHost` requires passing protocol while in 
 `requestUtils::getApiCdnHost` the protocol is resolved inside the utils method?

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -143,6 +145,66 @@ private function appendUiConfToContent($content)
 
 		return $content;
 	}
+
+	private function appendEnvConfigToContent($content)
+	{
+
+		$protocol = (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on') ? "https" : "http";
```
yossipapi: Use requestUtils::getRequestProtocol() it does the same thing.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
yossipapi: No obvious reason.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
yossipapi: There no function implanted that does this, but it may be useful if you want to add one.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -143,6 +145,66 @@ private function appendUiConfToContent($content)
...
+	{
+
+		$protocol = infraRequestUtils::getProtocol();
+
+		// The default Kaltura service url:
+		$serviceUrl = requestUtils::getApiCdnHost().'/api_v3';
+		// Default Kaltura CDN url:
+		$cdnUrl = requestUtils::getCdnHost($protocol);
+		// Default Stats URL
+		$statsServiceUrl = ($protocol == "https") ? $this->buildUrl($protocol,"stats_host_https") : $this->buildUrl($protocol,"stats_host");
```
yossipapi: BTWits probably worth to move this logic of if-else to the build urL function instead of repeating it multiple times.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
OrenMe: @yossipapi agree, but I think this should be also abstracted to infraUtils or statUtils so any service will be able to resolve these urls the same way.
@erankor - do you have plans to expose the stats end points in some infra and via INI config? maybe you already do?

PR #6288, opened 2017-11-07
Mercury 13.5.0 plat 7961 encryption infra

david-winder-kaltura:


configurations/base.ini
```diff
@@ -123,6 +123,8 @@ template_partner_id = 99
 DropFolderListTTL = 300
 ResourceReservationDuration = 5
 
+encryption_iv = @DEFAULT_IV_16B@
```
MosheMaorKaltura: Please add it to local.template.ini and also set this default.

MosheMaorKaltura:
Generic remark - 
1. We should add : maxFileSize for encryption configuration

alpha/lib/model/FileSync.php
```diff
@@ -43,15 +43,104 @@ public static function createForFileSyncKey ( FileSyncKey $key )
...
+		}
+
+		$key = $this->getEncryptionKey();
+		KalturaLog::debug("Decrypting content of fileSync " . $this->id . ". key is: [$key]");
+		$plainData = kEncryptFileUtils::decryptData($fileData, $key, $this->getIv());
+		return $plainData;
+	}
+
+	public function shouldEncryptFile()
+	{
```
MosheMaorKaltura: If file size is too big than return false, take the max file size from base.ini.

PR #6306, opened 2017-11-08
PLAT-8230:User add should call toInsertable for htmlPurify validation

MosheMaorKaltura:


gonenradai:
I can't review this...
I have no idea whether using toInsertableObject indeed goes through purifier or not, so my review doesn't really provide any benefit here.

PR #6309, opened 2017-11-08
KMS-14857: add language support to name and description in partial search

NadavHarnik:
add language support to name and description in partial search + optimize partial query for captions


MosheMaorKaltura:
reviewed

PR #6311, opened 2017-11-09
PLAT-8207: fix group_ids indexing on kuser

NadavHarnik:


alpha/lib/model/KuserKgroupPeer.php
```diff
@@ -99,7 +99,36 @@ public static function retrieveKgroupIdsByKuserId($kuserId){
...
+	 */
+	public static function retrieveKgroupIdsByKuserIdAndPartnerId($kuserId, $partnerId)
+	{
+		if (isset(self::$kgroupIdsByKuserId[$kuserId])){
+			return self::$kgroupIdsByKuserId[$kuserId];
+		}
+
+		$c = new Criteria();
+		$c->add(KuserKgroupPeer::KUSER_ID, array($kuserId), Criteria::IN);
+		$c->addAnd ( KuserKgroupPeer::STATUS, array(KuserKgroupStatus::DELETED), Criteria::NOT_IN);
```
MosheMaorKaltura: These 2 rules are not needed, you already get them by the default criteria , (Status != delete and partner ID).
And also remove the call to  setUseCriteriaFilter(false)

NadavHarnik:
when indexing from populate script we don't have partnerId from kCurrentContext::getCurrentPartnerId(), so we need to add it from the kuser to the criteria

MosheMaorKaltura:
reviewed

PR #6377, opened 2017-11-20
PLAT-7977: add a parameter signifying when the last time view_mode was …

noam-arad:
…changed


alpha/lib/model/LiveEntry.php
```diff
@@ -954,6 +955,7 @@ public function getViewMode()
 	public function setViewMode($v)
 	{
 		$this->putInCustomData(self::CUSTOM_DATA_VIEW_MODE, $v);
+		$this->setLastChangedViewMode(time());
```
kalturaguy: i think you need to call setLastChangedViewMode only when it actually changed and not the same..

PR #6394, opened 2017-11-21
PLAT-8173: remove creator kuser id from entitlement condition to alig…

NadavHarnik:
…n code to kEntitlementUtils::isEntryEntitled


MosheMaorKaltura:
reviewed

PR #6462, opened 2017-12-04
Mercury 13.8.0 plat 8313

noam-arad:


alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -324,8 +324,12 @@ protected function getLivePackagerUrl($serverNode, $streamFormat = null)
 		
 		$livePackagerUrl = "$livePackagerUrl/p/$partnerID/e/$entryId/sd/$segmentDuration/";
 		$secureToken = $this->generateLiveSecuredPackagerToken($livePackagerUrl);
-		$livePackagerUrl .= "t/$secureToken/"; 
-		
+		$livePackagerUrl .= "t/$secureToken/";
+
+		$entry = $this->getDynamicAttributes()->getEntry();
+		if ($entry->getExplicitLive() &&  (!$entry->canViewExplicitLive()))
```
yossipapi: @kalturaguy will this require changes on the packager configuration file as well?

As to passing "type/user" it would be better to always pass type if explicitLive is turned on rather than passing it only if its type user. So if user pass user and if admin pass admin.

Aslo maske "user" & "admin" const's.

deployment/updates/scripts/xml/notifications/explicit_live_notification.template.xml
```diff
@@ -25,7 +25,7 @@
 				<eventConditions objectType="array">
 					<item objectType="KalturaEventObjectChangedCondition">
 						<description>Entry view mode or recording status changes</description>
-						<modifiedColumns>{view_mode,recording_status}</modifiedColumns>
+						<modifiedColumns>view_mode,recording_status</modifiedColumns>
```
yossipapi: Did you meant to merge this change along with the url modification? 
It would be better to separate to 2 different pull requests.

noam-arad: you are correct, will do

alpha/lib/model/DeliveryProfileLive.php
kalturaguy: @yossipapi  yes, we are modifying the packager config too...

noam-arad: I looked at the liveController code, since we only set a clipTo value if explicit live is enabled then even if we pass /type/user when it is off it won't have any effect.
I will make the requested change.

PR #6471, opened 2017-12-04
PLAT-8313: minor fix

noam-arad:


yossipapi:
Noam please add a new script for the deployment and do not re use the old one.

PR #6485, opened 2017-12-05
Plat-8366 alert when asking for unsupported format in storage delivery profile

ZurKaltura:


ZurKaltura:
playManifest CompatCheck was done successfully

PR #6494, opened 2017-12-06
add html5 version to the studio/studioV3 config

yairans:


atarsh:
@yairans does this change means you no longer need the uiconfID and the config fields?
if so, please remove them from all relevant places.

OrenMe:
@atarsh we will first run a test and then coordinate this with deployment instructions - if all is well we will clean it up.

yossipapi:
Looks Ok

PR #6528, opened 2017-12-12
Update plugins.template.ini

DBezemer:
Add missing Beacon plugin


yossipapi:
The beacon plugin should be off by default. 
This plugin relies on other plugins which are also not enabled by default.

DBezemer:
@yossipapi can you then add it commented out and also add the other plugins it depends on?

PR #6539, opened 2017-12-17
Update categoryCount.php

MosheMaorKaltura:


alpha/scripts/categoryCount.php
```diff
@@ -3,11 +3,12 @@
 
 require_once(__DIR__ . '/bootstrap.php');
 
-
-if(!count($argv))
-	die("No partner_id passed to script!");
+if(count($argv)<2)
```
hilak: @MosheMaorKaltura  why not just use $argc?

MosheMaorKaltura: Actually , why not?

PR #6545, opened 2017-12-18
Merucry-13.9.0-PLAT-8414 - Moving filter logic from the API to the DB

ZurKaltura:


alpha/lib/model/BatchJob.php
```diff
@@ -436,7 +436,14 @@ public function getChildJobs(Criteria $c = null)
 		BatchJobPeer::setUseCriteriaFilter(true);
 		return array_merge($result1, $result2);
 	}
-	
+
+	public function getOpenStatusChildJobs(Criteria $c = null)
+	{
+		$c ? $c = clone $c : $c = new Criteria();
```
yossipapi: no need to clone the criteria.
And also better do:
$c = $c ? $c : new Criteria();

PR #6559, opened 2017-12-20
PLAT-8354: avoid warning when missing label from caption

inbal-ben-david:


MosheMaorKaltura:
Reviewed - Must run playManifest compat check

PR #6578, opened 2017-12-24
throw error if bundleConfig does not exist

yairans:
FEC-7654

### DIFF
--- a/alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
+++ b/alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
@@ -339,29 +339,27 @@ class embedPlaykitJsAction extends sfAction
 	
 	private function setLatestOrBetaVersionNumber()
 	{
-		if ($this->bundleConfig) {
-			//if latest/beta version required set version number in config obj
-			$isLatestVersionRequired = array_search("{latest}", $this->bundleConfig) !== false;
-			$isBetaVersionRequired = array_search("{beta}", $this->bundleConfig) !== false;
+		//if latest/beta version required set version number in config obj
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -428,6 +426,9 @@ private function initMembers()
 		
 		$this->bundleConfig = json_decode($confVars, true);
 		$this->mergeVersionsParamIntoConfig();
+		if (!$this->bundleConfig) {
```
yossipapi: mergeVersionsParamIntoConfig uses the bundle config ($this->bundleConfig = array_merge($this->bundleConfig, $versionsArr);) so you may want to move this check up. 

If bundleConfig in null and you will call array_merge it will result in a warning.

Another option is to define the bundleConfig to default value of an empty array and change this check accordingly.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -339,29 +339,27 @@ private function mergeVersionsParamIntoConfig()
 	
 	private function setLatestOrBetaVersionNumber()
 	{
-		if ($this->bundleConfig) {
-			//if latest/beta version required set version number in config obj
-			$isLatestVersionRequired = array_search("{latest}", $this->bundleConfig) !== false;
-			$isBetaVersionRequired = array_search("{beta}", $this->bundleConfig) !== false;
+		//if latest/beta version required set version number in config obj
+		$isLatestVersionRequired = array_search("{latest}", $this->bundleConfig) !== false;
```
yossipapi: Move the strings "{latest}" and "{beta}" to be class constants since you use them more than once.

PR #6606, opened 2018-01-01
SUP-11664 - get live default thumbnail in case recorded entry is not …

gotlieb:
…ready yet and has no thumbnail

@MosheMaorKaltura

### DIFF
--- a/alpha/lib/model/entry.php
+++ b/alpha/lib/model/entry.php
@@ -3509,7 +3509,8 @@ public function copyTemplate($copyPartnerId = false, $template)
 		}
 		elseif ($this->getType () == entryType::MEDIA_CLIP || $this->getType() == entryType::PLAYLIST) {
 			try {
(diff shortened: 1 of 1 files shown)


alpha/lib/model/entry.php
```diff
@@ -3886,6 +3888,28 @@ protected function getCategoriesElasticSearchData($categoryEntries)
...
+	 */
+	protected function getDefaultThumbPath($type)
+	{
+		switch ($type)
+		{
+			case entryType::MEDIA_CLIP:
+			case entryType::PLAYLIST:
+			{
+				// in case of a recorded entry from live that doesn't have flavors yet nor thumbs we will use the live default thumb.
+				if ($this->getSourceType() != null && $this->getSourceType() == EntrySourceType::RECORDED_LIVE && !assetPeer::countByEntryId($this->getId(), array(assetType::FLAVOR, assetType::THUMBNAIL)))
```
yossipapi: You have another recorded source type KALTURA_RECORDED_LIVE

PR #6615, opened 2018-01-04
TR-2059

hilak:
When transcript is verified, it should not be overwritten by content from VoiceBase.


MosheMaorKaltura:
Reviewed

PR #6644, opened 2018-01-14
PLAT-8471: fix fatal when there is no form and set condition type to …

david-winder-kaltura:
…read only outside the template


MosheMaorKaltura:
reviewed

PR #6653, opened 2018-01-15
PLAT-8465: if params is not an array - throw API exception

david-winder-kaltura:


api_v3/lib/KalturaRequestDeserializer.php
```diff
@@ -260,8 +260,10 @@ private function validateFile($fileData)
 		}
 	}
 
-	private function buildObject(KalturaTypeReflector $typeReflector, array &$params, $objectName)
+	private function buildObject(KalturaTypeReflector $typeReflector, &$params, $objectName)
```
MosheMaorKaltura: Add objectName to the exception message

PR #6668, opened 2018-01-17
No Plat: When filtering on fields index with prefix use the same prefix when using in criteria + fix enum value 0 not taken into filter

yossipapi:


plugins/search/providers/sphinx_search/lib/SphinxCriteria.php
```diff
@@ -724,7 +724,14 @@ protected function applyFilterFields(baseObjectFilter $filter)
 					if(count($vals))
 					{
 						$vals = array_slice($vals, 0, SphinxCriterion::MAX_IN_VALUES);
-						$vals = array_filter($vals, 'trim');
+						$vals = array_map("trim", $vals);
```
MosheMaorKaltura: Since we already iterate on all the values in the foreach loop above ,why not trim and remove the empty string there?

PR #6675, opened 2018-01-18
PLAT-8493: check if delivery profile ids var is empty instead of is null

inbal-ben-david:
added to handle the case where ids equals to {}


MosheMaorKaltura:
reviewed

PR #6680, opened 2018-01-21
get ovp/ott env from configuration

yairans:


alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -407,21 +436,24 @@ private function initMembers()
 			KExternalErrors::dieError(KExternalErrors::MISSING_PARAMETER, self::UI_CONF_ID_PARAM_NAME);
 		
 		// retrieve uiCong Obj
-		$uiConf = uiConfPeer::retrieveByPK($this->uiconfId);
-		if (!$uiConf)
+		$this->uiConf = uiConfPeer::retrieveByPK($this->uiconfId);
```
yossipapi: I don't see you added this member to the class similar to what we do for uiConfId:
private $uiconfId = null;

PR #6685, opened 2018-01-22
Mercury-13.12.0-PLAT-8503 - media re purposing failures

ZurKaltura:
Adding a try catch for get user when getting the email
Removing impressionist and unimpersonate partner from each engine and only do it from a central location


MosheMaorKaltura:
Reviewed.

PR #6722, opened 2018-01-30
PLAT-8530: check if fileSync need to be sync between DB before create…

david-winder-kaltura:
… remote FileSync


MosheMaorKaltura:
Reviewed

PR #6724, opened 2018-01-30
PLAT-8047: update notification template to send for every status change

noam-arad:


deployment/updates/scripts/xml/2018_01_30_UpdateServerNodeDownTemplate.template.xml
```diff
@@ -0,0 +1,63 @@
...
+        <request service="eventNotificationTemplate" action="listTemplates" plugin="eventNotification">
+            <filter objectType="KalturaEventNotificationTemplateFilter">
+                <systemNameEqual>Server_Node_Down</systemNameEqual>
+            </filter>
+            <pager objectType="KalturaFilterPager">
+                <pageSize>1</pageSize>
+            </pager>
+        </request>
+
+        <request service="eventNotificationTemplate" action="update" plugin="eventNotification" partnerId="0">
```
yossipapi: Missing the following:
1. The update made to the body of the notification ("is down" => "Status has changed ...")
2. You did not remove from the event conditions the condition the checks the new status is NOT_REGISTERED
3. The system name no longer matches what the notification does so either update the system name (if possible) or create a template.

The same comments should probably be valid for the file without the template.

PR #6729, opened 2018-01-31
PLAT-8538: remove duplicate values in user_names field

NadavHarnik:


alpha/lib/model/entry.php
```diff
@@ -3611,7 +3611,11 @@ private function getAllUserNamesAsArray()
 		}
 
 		if($result)
+		{
+			$result = array_unique($result);
```
MosheMaorKaltura: look like the order of these calls should be replaced , in case that same value is used for some of field , no point in taking it together.

PR #6733, opened 2018-01-31
Mercury 13.12.0 plat 8446

inbal-ben-david:


api_v3/services/UserService.php
```diff
@@ -630,4 +630,66 @@ public function loginByKsAction($requestedPartnerId)
...
+	 *
+	 * Will serve a requested csv
+	 * @action serveCsv
+	 *
+	 *
+	 * @param string $id - the requested file id
+	 * @return string
+	 */
+	public function serveCsvAction($id)
+	{
```
MosheMaorKaltura: verify that id is a pure file and does not contain '..\' symbols

api_v3/services/UserService.php
```diff
@@ -630,4 +630,66 @@ public function loginByKsAction($requestedPartnerId)
...
+		// KS verification - we accept either admin session or download privilege of the file
+		$ks = $this->getKs();
+		if(!$ks || !($ks->isAdmin() || $ks->verifyPrivileges(ks::PRIVILEGE_DOWNLOAD, $id)))
```
MosheMaorKaltura: Use permission config to validate that there us at least user KS

PR #6748, opened 2018-02-04
No Plat: change file sync version column to int to allow optimizing the query used to delete old file_sync versions

yossipapi:


alpha/apps/kaltura/lib/storage/kFileSyncUtils.class.php
```diff
@@ -1586,19 +1586,14 @@ private function deleteOldFileSyncVersions(FileSync $newFileSync)
 			$c->add ( FileSyncPeer::OBJECT_TYPE , $newFileSync->getObjectType() );
 			$c->add ( FileSyncPeer::OBJECT_SUB_TYPE , $newFileSync->getObjectSubType() );
 			$c->add ( FileSyncPeer::STATUS, array(FileSync::FILE_SYNC_STATUS_PURGED, FileSync::FILE_SYNC_STATUS_DELETED), Criteria::NOT_IN);
+			$c->add ( FileSyncPeer::VERSION, $intVersion - $keepCount, Criteria::LESS_THAN);
```
etameran: Change the order to DESC since the key doesn't include the status and we rather delete version prior to the int-keep (if everything is working well it won't change much anyway as you wont have many versions to delete).
UNIQUE KEY `unique_index` (`object_id`,`object_type`,`version`,`object_sub_type`,`dc`,`deleted_id`),

jessp01:
@yossipapi, please change https://github.com/kaltura/server/blob/Mercury-13.13.0/deployment/base/sql/01.kaltura_ce_tables.sql accordingly.

yossipapi:
@jessp01, done.

PR #6750, opened 2018-02-04
PLAT-8165: optimize elastic queries

NadavHarnik:
if the operator is AND and the field has no boost factor we can move it to filter context to skip score calculation and reduce query time


MosheMaorKaltura:
Reviewed

PR #6783, opened 2018-02-13
Mercury-13.14.0-SUP-13039 Playing apple_http with non kaltura live

ZurKaltura:


ZurKaltura:
This code change was validated using play manifest compact check

PR #6794, opened 2018-02-15
Shuffle the $deliveryNodes array to randomize the assigned KES

supermasita:
When more than one KES is assigned as primary, you will always get only the first one. This allows to randomize the $deliveryNodes array and eventually get different KES.

In some cases, you want to have more than one KES assigned and you usually need to use the same playback domain for all of them and then create a DNS round robin for it. This change, although its not round robin, will allow some balancing between nodes.

Thanks,

### DIFF
--- a/alpha/lib/model/DeliveryProfile.php
+++ b/alpha/lib/model/DeliveryProfile.php
@@ -477,7 +477,10 @@ abstract class DeliveryProfile extends BaseDeliveryProfile implements IBaseObjec
 			KalturaLog::debug("No active delivery nodes found among the requested edge list: " . print_r($deliveryNodeIds, true));
 			return null;
 		}
-		
+	
+	        /* Shuffle the array to randomize the assigned KES, if more than one in the same rule */
+		shuffle($deliveryNodes);
+	
 		$deliveryNode = null;
 		foreach ($deliveryNodes as $node)
 		{


alpha/lib/model/DeliveryProfile.php
```diff
@@ -477,7 +477,10 @@ public function getDeliveryServerNodeUrl($removeAfterUse = false)
 			KalturaLog::debug("No active delivery nodes found among the requested edge list: " . print_r($deliveryNodeIds, true));
 			return null;
 		}
-		
+	
+	        /* Shuffle the array to randomize the assigned KES, if more than one in the same rule */
+		shuffle($deliveryNodes);
+
```
kalturaguy: this will break use cases for primary/backup:
-  primary and backup CAN be routed different
-  primary can get flavor set X and backup can get flavor set Y (for example lower quality)

If we want other high availablity modes, we need first to configure them on the dashboard and pass the mode with the accesscontrol profile.

supermasita: Hi @kalturaguy 

This should not break the use case for primary/backup: _you can still have **one** server set as primary and **one** as backup - in active/passive way - this should affect **only when more than one server exists in the list**_.

* The customer may choose to use more than one server as primary (or backup), but if you add more than one to the list, you only get the first one; always. Many customer would prefer an active/active setup and, in some case, it might be needed as only one server could not be enough to handle the job (KES servers might have limite BW, for example).
* This would lower the hitratio initially, that is correct, and increase the traffic to CDN if these KES do not have a parent.
* The primary/backup workflow (including which flavours are used) should not be altered by this change.


Extending on what I ve said on my first comment: 
* This idea comes up from my work from one our customers.
* All KESs are limited to 1Gbps IFs (no bonding possible)
* Some locations sum up more than 1Gbps in traffic, so more than one KES is needed.
* We could have sliced their network in smaller parts, and create many delivery rules for a same location, but if the primary KES goes down and all traffic needs to be routed to the backup, the 1Gbps will be reached.
* We had to:
  1. Create a playback domain which is set up as DNS RR among all KES for the location
  1. Add a healthchech to the DNS RR (which is inefficient, as it is only a ICMP)
  1. Assign the same playback domain to all KES
  1. Assign all KES as PRIMARY, so as long as one KES is active, you will get the playback domain.

_Keep in mind that what we did with this customer might not be available to all of them, and it also adds a layer which is managed by the customer and could have problems._

**This very simple change does not want to address HA, but rather just help "distribution/balancing" in the most basic way**. If you want to explore more complex HA possibilities, thats even better (but I am sure they will require more time and coding).

Thanks!

kalturaguy: ok, understood, i though you wanted to mix the backup and primary together... all good, you can merge

supermasita: Thanks @kalturaguy ! Someone else will have to merge, as I don't have write access.

DBezemer:
@yossipapi can you merge?

yossipapi:
Guy wasn't the initial assumption that the order of the KES nodes matter? i.e what is first should be used first. This change does not align with that.

Also note that the playmanifeat calls are cached so this may not have that kind of dramatic affect as needed. 

In my opinion this shouldn't be added and if added than this behavior should be very carefully documented so that the random URL behavior will be known to partners.

supermasita:
Hi @yossipapi 

I cannot answer about what it was expected to do, but I can tell you that ONLY the first KES node will be ever used - that is what we are trying to avoid.

I also cannot confirm about the caching done in SaaS as, of course, I ve tested this change in a local enviroment.

If Guy and you decide this change is not worth, I understand and I will add a ticket for you to analyze how to solve this problem/limitation. If you decide to go ahead with the change, I am happy to help you document it.

Thanks!!

PR #6814, opened 2018-02-26
fix PHP Strict Standards

yairans:
set `playerConfig` as empty object is it doesn't exist


alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -149,6 +149,10 @@ private function appendConfig($content)
 
 	private function mergeEnvConfig($uiConf)
 	{
+		if (!$uiConf) 
+		{
+			return;
```
OrenMe: Now you will never get provider info if uiconf doesn't exist, need to create stdClass if it doesn't

PR #6822, opened 2018-02-28
PLAT-8492: allow users with specific permissions to update or insert …

noam-arad:
…adminTags


api_v3/lib/KalturaEntryService.php
```diff
@@ -1396,7 +1396,12 @@ protected function validateEntitledUsersUpdate(KalturaBaseEntry $entry, entry $d
 	protected function checkAdminOnlyUpdateProperties(KalturaBaseEntry $entry)
 	{
 		if ($entry->adminTags !== null)
-			$this->validateAdminSession("adminTags");
+		{
+			$canUpdateAdminTags = $this->hasPermissionToWriteProperty("admin_tags");
```
yossipapi: Better to first check isAdmin and only than the other option

yossipapi: Why not add the admin_tags edit as a privilege on the KS? and check for that instead of adding a new permission which will be needed to add to the role the user is coming with.

What I mean is something similar to what we do here KalturaEntryService.php::1346:
if (!$ks ||(!$this->getKs()->isAdmin() && !$ks->verifyPrivileges(ks::PRIVILEGE_EDIT_USER, $entry->userId)))

PR #6826, opened 2018-02-28
Use OpenSSL functions if mcrypt is not available

jessp01:
This is the server side follow up of https://github.com/kaltura/clients-generator/pull/278

Starting from PHP 7.1, the Mcrypt extension is deprecated.
There is a good underlying reason for that, to wit: libmcrypt's last update was back in 2007. This is bad for any piece of code but especially troubling for security sensitive ones.

The pulls above will ensure compatibility with KS strings generated using Mcrypt because we set the OPENSSL_ZERO_PADDING bit and do our own padding thus making it compatible with mcrypt's PKCS#5 padding, as opposed to OpenSSL's default which is PKCS#7.

More info about this can be found here:
http://thefsb.tumblr.com/post/110749271235/using-opensslendecrypt-in-php-instead-of
and in particular under the "Encoding, padding and OPENSSL_RAW_DATA vs.
OPENSSL_ZERO_PADDING" section.

### DIFF
--- a/infra/general/KCryptoWrapper.class.php
+++ b/infra/general/KCryptoWrapper.class.php
@@ -2,14 +2,26 @@
 
 class KCryptoWrapper
 {
-    public static function getEncryptor()
-    {
-	if (extension_loaded('mcrypt')){
-	    return new McryptWrapper();
-	}else{
-	    return new OpenSSLWrapper();
-	}
-    }
+        public static function getEncryptorClassName()
+        {
+                if (extension_loaded('mcrypt')){
+		    return 'McryptWrapper';
+                }else{
+		    return 'OpenSSLWrapper';
+                }
+        }
+
+        public static function __callStatic($func, $args)
+        {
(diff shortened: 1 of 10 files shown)


infra/general/KCryptoWrapper.class.php
```diff
@@ -0,0 +1,136 @@
...
+}
+
+class OpenSSLWrapper
+{
+    const AES_BLOCK_SIZE = 16;
+    const DES3_BLOCK_SIZE = 8;
+    const AES_METHOD = 'AES-128-CBC';
+    const DES3_METHOD = 'DES-EDE3';
+
+    public static function random_pseudo_bytes($bytes)
```
MosheMaorKaltura: rename bytes to length or size, since it may be confusing.

jessp01: I don't think it's confusing but renamed it to $length anyhow since it's not worth arguing about.

PR #6862, opened 2018-03-13
Mercury 13.15.0 sup 13373

noam-arad:


alpha/lib/model/flavorAsset.php
```diff
@@ -324,4 +324,41 @@ public function shouldEncrypt()
...
+		if (isset($multiStream))
+		{
+			$multiStreamObj = json_decode($multiStream);
+			if (isset($multiStreamObj))
+			{
+				if (isset($multiStreamObj->audio) && isset($multiStreamObj->audio->languages) && count($multiStreamObj->audio->languages) > 0)
+				{
+					$flavorLang = $multiStreamObj->audio->languages[0];
+					$this->setLanguage($flavorLang);
+					$entry = $this->getentry();
```
MosheMaorKaltura: Validating that entry exists better be done in first place.

alpha/lib/model/flavorAsset.php
```diff
@@ -324,4 +324,41 @@ public function shouldEncrypt()
...
+					if (!$entry)
+						throw new kCoreException("Invalid entry id [".$this->getEntryId()."]", APIErrors::INVALID_ENTRY_ID);
+					$conversionProfile = $entry->getconversionProfile2();
+					if ($conversionProfile->getDefaultAudioLang() == $flavorLang)
+					{
+						$this->setDefault(true);
```
MosheMaorKaltura: Since it is possible to change conversionProfile after setting flavorasset->isDefault , then this info might not be authentic.
Why do we need to set this audio flavor as default while we always can use a function that tells if this language is the default for this conversiobProfile?

PR #6867, opened 2018-03-13
CURLOPT_SSL_VERIFYHOST no longer accepts the value 1

jessp01:
As documented here http://php.net/manual/en/function.curl-setopt.php:
1 used to mean check the existence of a common name in the SSL peer certificate.
2 will check the existence of a common name and also verify that it matches
the hostname provided. 0 will not check the names.

Support for value 1 was removed in cURL 7.28.1.
That means that all machines that have libcurl of version 7.28.1 or
higher will effectively behave as if this value was set to 2.

7.28.1 is very old and no current Linux distro version includes it.
Just by way of reference, RHEL/CentOS 7 has 7.29.0, Ubuntu 14.04 has
7.35.0 and Ubuntu 16.04 has 7.47.0.

### DIFF
--- a/infra/general/KCurlWrapper.class.php
+++ b/infra/general/KCurlWrapper.class.php
@@ -225,7 +225,7 @@ class KCurlWrapper
 		if(!$params || !isset($params->curlVerifySSL) || !$params->curlVerifySSL)
 		{
 			curl_setopt($this->ch, CURLOPT_SSL_VERIFYPEER, false);
-			curl_setopt($this->ch, CURLOPT_SSL_VERIFYHOST, 2);
+			curl_setopt($this->ch, CURLOPT_SSL_VERIFYHOST, self::getSslVerifyHostValue());
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


infra/general/KCurlWrapper.class.php
```diff
@@ -225,7 +225,7 @@ public function __construct($params = null) {
 		if(!$params || !isset($params->curlVerifySSL) || !$params->curlVerifySSL)
 		{
 			curl_setopt($this->ch, CURLOPT_SSL_VERIFYPEER, false);
-			curl_setopt($this->ch, CURLOPT_SSL_VERIFYHOST, 1);
+			curl_setopt($this->ch, CURLOPT_SSL_VERIFYHOST, 2);
```
MosheMaorKaltura: Since our server in production run older version of curl - 7.19.4,
and to maintain backward compatibility, lets add static function getSslVerifyHostValue() that will decide its value based on curl_version().

MosheMaorKaltura:
Reviewed

PR #6891, opened 2018-03-19
PLAT-8581: add metadata service object access validations

yossipapi:


rkreich:
@yossipapi will it handle dynamic objects correctly?

yossipapi:
Yes, currently I left that part empty (return true) but I will add validations there as well.

plugins/metadata/lib/model/MetadataCategoryPeer.php
```diff
@@ -14,4 +14,33 @@ public static function getEntry($objectId)
...
+    		KalturaLog::debug("Metadata object id with id [$objectId] not found");
+    		return false;
+    	}
+    	
+    	if (kEntitlementUtils::getEntitlementEnforcement())
+    	{
+    		$currentKuserCategoryKuser = categoryKuserPeer::retrievePermittedKuserInCategory($categoryDb->getId(), kCurrentContext::getCurrentKsKuserId(), array(PermissionName::CATEGORY_EDIT));
+    		if(!$currentKuserCategoryKuser || $currentKuserCategoryKuser->getPermissionLevel() != CategoryKuserPermissionLevel::MANAGER)
+    		{
+    			KalturaLog::debug("Current user is not permitted to access categoru with id [$objectId]");
```
jessp01: Small typo here "categoru" instead of "category". Also, IMHO, should be KalturaLog::err() in this case, not KalturaLog::debug().

PR #6915, opened 2018-03-26
PLAT-8733: fix potential fatal

noam-arad:


noam-arad:
should be committed to Mercury-13.17.0

PR #6922, opened 2018-03-26
Mercury 13.15.0 rtc

noam-arad:


alpha/apps/kaltura/lib/live/kBroadcastUrlManager.php
```diff
@@ -211,4 +211,17 @@ public function getSecondaryBroadcastUrl(LiveStreamEntry $entry, $protocol)
...
 	}
+
+	public function getRTCBroadcastingUrl($protocol, $hostname, $externalPort)
+	{
+		if (!$hostname)
+		{
+			return '';
+		}
+		$url = "$protocol://$hostname";
+		if ($externalPort)
```
kalturaguy: 1.i think you ought to rename the $externalport, since we probably need to pass entire URI (for loadbalancers and such)

plugins/konference/lib/model/ConferenceEntryServerNode.php
```diff
@@ -0,0 +1,75 @@
...
+		 * @var ConferenceServerNode $conferenceServerNode
+		 */
+		if ($this->getConfRoomStatus() != ConferenceRoomStatus::READY)
+		{
+			throw new kCoreException(KalturaKonferenceErrors::ROOM_NOT_READY, $this->getId());
+		}
+
+		$hostname = $conferenceServerNode->getHostName();
+		$manager = kBroadcastUrlManager::getInstance($partnerId);
+		$url = $manager->getRTCBroadcastingUrl('https', $hostname, $conferenceServerNode->getExternalPort());
```
kalturaguy: using the $hostname is not enough, since we probably use load balancer in front,  maybe use something similar to $playbackDomain like we do in KES

plugins/konference/lib/model/ConferenceServerNode.php
```diff
@@ -0,0 +1,49 @@
...
+ 	*/
+	public function postUpdate(PropelPDO $con = null)
+	{
+		if($this->isColumnModified(ServerNodePeer::STATUS) && $this->getStatus() === ServerNodeStatus::NOT_REGISTERED)
+		{
+			$this->removeAttachedEntryServerNodes();
+		}
+	}
+
+	public function getExternalPort()
```
kalturaguy: maybe let's rename the external Port to entryPoint Uri or something similar

plugins/konference/lib/model/ConferenceEntryServerNode.php
kalturaguy: should this be http or wss?

PR #6956, opened 2018-04-04
PLAT-8765 - credit fixes

gotlieb:
@yossipapi

### DIFF
--- a/plugins/reach/lib/api/data/KalturaBaseVendorCredit.php
+++ b/plugins/reach/lib/api/data/KalturaBaseVendorCredit.php
@@ -39,4 +39,12 @@ abstract class KalturaBaseVendorCredit extends KalturaObject implements IApiObje
 
 		return $credit;
 	}
+
+		/* (non-PHPdoc)
+		* @see KalturaObject::validateForInsert()
+		*/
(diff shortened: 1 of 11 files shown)


plugins/reach/lib/api/data/KalturaBaseVendorCredit.php
```diff
@@ -39,4 +39,12 @@ public static function getInstance($sourceObject, KalturaDetachedResponseProfile
 
 		return $credit;
 	}
-}
\ No newline at end of file
+
+		/* (non-PHPdoc)
+		* @see KalturaObject::validateForInsert()
+		*/
+	  public function validateForInsert($propertiesToSkip = array())
```
yossipapi: Why do you need this if this only calls it's parent?

plugins/reach/services/ReachProfileService.php
```diff
@@ -30,6 +30,7 @@ public function initService($serviceId, $serviceName, $actionName)
 	 */
 	public function addAction(KalturaReachProfile $reachProfile)
 	{
+		KalturaLog::debug("EREZ1");
```
yossipapi: Remove the debug?

PR #6967, opened 2018-04-09
PLAT-8723

gotlieb:
fix missing strtolower

@MosheMaorKaltura 
@erankor


hilak:
@gotlieb I would advise against this. In this manner you are allowing an application to create 2 versions of a view history. For instance, user user.name logged into an app as User.Name would have 2 ViewHistory objects for the same entry which would cause inconsistency. Is this edge case handled prevented anywhere?

PR #6977, opened 2018-04-11
PLAT-7832: enable API server side of the live Clipping

david-winder-kaltura:


api_v3/lib/types/EntryServerNode/base/KalturaClippingTaskEntryServerNode.php
```diff
@@ -0,0 +1,38 @@
...
+	public $clipAttributes;
+
+	/**
+	 * @var string
+	 */
+	public $clippedEntryId;
+
+	private static $map_between_objects = array
+    (
+        "clipAttributes",
```
noam-arad: fix indentation

api_v3/lib/KalturaEntryService.php
```diff
@@ -700,6 +700,13 @@ protected function attachOperationResource(kOperationResource $resource, entry $
 				return $this->attachLiveEntryResource($internalResource, $dbEntry, $dbAsset, $operationAttributes);
 			}
 
+			$srcEntry = self::getEntryFromContentResource($internalResource);
```
noam-arad: should be checked when we check for multi clip, if someone tries to multi-clip on a recorded entry we should notify them this operation is not permitted.

PR #6985, opened 2018-04-13
PSVAMB-2895

hilak:
Fix issue - when access control profile specifies Delivery Profile which is not associated with the partner (even if it's a global profile, or belongs to the partner), Delivery Profile lookup will fail.


hilak:
@erankor please review. I think it's not necessary to check existence of the count($deliveryAttributes->getIsDeliveryProfilesBlockedList ()) on line 361 since it is being checked later on 388. In fact, DeliveryProfilePeer::filterDeliveryProfilesCriteria behaves as if the 2 arrays are mutually exclusive.

hilak:
approved by @erankor

PR #7037, opened 2018-04-25
SUP-14151 - fix retry on widevine packager failure

gotlieb:
@MosheMaorKaltura 
@david-winder-kaltura 
@noam-arad 

since we have few failures per month in avg we can retry on all failures.

### DIFF
--- a/plugins/drm/providers/widevine/batch/Convert/OperationEngines/KWidevineOperationEngine.php
+++ b/plugins/drm/providers/widevine/batch/Convert/OperationEngines/KWidevineOperationEngine.php
@@ -158,14 +158,9 @@ class KWidevineOperationEngine extends KOperationEngine
 		KBatchBase::unimpersonate();
 		$errorMessage = '';
 		$errorMessage = KWidevineBatchHelper::getEncryptPackageErrorMessage($returnValue);
-		$logMessage = 'Package encryption failed, asset name: '.$this->packageName.' error: '.$errorMessage;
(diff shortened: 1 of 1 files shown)


plugins/drm/providers/widevine/batch/Convert/OperationEngines/KWidevineOperationEngine.php
```diff
@@ -158,14 +158,9 @@ private function handleEncryptPackageError($returnValue)
...
-		$logMessage = 'Package encryption failed, asset name: '.$this->packageName.' error: '.$errorMessage;
+		$logMessage = 'Package encryption failed, asset name: '.$this->packageName.' error: '.$errorMessage . ' error code: ' .$returnValue;
 		KalturaLog::err($logMessage);
-					
-		// in some cases this specific Widevine error needs a simple job retry in order to convert successfully  
-		if ($returnValue == KWidevineBatchHelper::SYNC_FRAME_OFFSET_MATCH_ERROR)
-			throw new kTemporaryException ($logMessage);
-			
-		throw new KOperationEngineException($logMessage);		
+		throw new kTemporaryException ($logMessage);
```
MosheMaorKaltura: Since you changed functionality please validate the following -
1. From the logs - what are the failure types that we encounter?
2. Can we have cases that the failure will persistent thus the retry not be required.
3. How many retries will that yield at max? we need no more than 1.

PR #7046, opened 2018-04-27
Mercury 13.19.0 notifications

DBezemer:
Adding missing notification templates and renaming files with incorrect formatted dates


DBezemer:
@jessp01 FYI

DBezemer:
@MosheMaorKaltura please review and merge for 13.20, and if possible add a review step for new notifications so this problem does not occur again where notifications are only created in the /tests/standAloneClient folder and do not have an associated install/update script

PR #7088, opened 2018-05-08
SUP-14325:Validate the size of the local file when downloading the file from th…

ravitshalem:
…e remote drop folder

### DIFF
--- a/infra/storage/kFileTransferMgr.class.php
+++ b/infra/storage/kFileTransferMgr.class.php
@@ -428,7 +428,7 @@ abstract class kFileTransferMgr
 	 *
 	 * @return FILETRANSFERMGR_RES_OK / FILETRANSFERMGR_RES_ERR
 	 */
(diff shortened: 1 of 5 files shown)


infra/storage/kFileTransferMgr.class.php
```diff
@@ -445,21 +445,35 @@ public function getFile ( $remote_file, $local_file = null)
...
 				
 			return self::FILETRANSFERMGR_RES_OK;
 		}
 	}
 
+	private function validateFileSize( $expectedSize ,$localFile)
+	{
+		clearstatcache();
+		$actualFileSize = kFile::fileSize($localFile);
+		if($expectedSize && $actualFileSize < $expectedSize)
```
MosheMaorKaltura: Lets check $expectedSize in outer scope , since if the value is 0 , no since in checking actual file size and clearcache.

MosheMaorKaltura:
Reviewed

PR #7099, opened 2018-05-09
PSVAMB-3216

hilak:
Fix issue - When an attachment asset's content is updated from a KalturaStringResource, the asset's file extension disappears.


hilak:
reviewed by @yossipapi

PR #7140, opened 2018-05-16
PLAT-8918: strip utf8 invalid chars if search term is longer then maximum length

NadavHarnik:


plugins/search/providers/elastic_search/lib/api/types/items/KalturaESearchItemImpl.php
```diff
@@ -12,7 +12,8 @@ public static function eSearchItemToObjectImpl(&$eSearchItem, $dynamicEnumMap, $
 	{
 		if(strlen($eSearchItem->searchTerm) > self::MAX_SEARCH_TERM_LENGTH)
 		{
-			$eSearchItem->searchTerm = substr($eSearchItem->searchTerm, 0, self::MAX_SEARCH_TERM_LENGTH);
+			$searchTerm = substr($eSearchItem->searchTerm, 0, self::MAX_SEARCH_TERM_LENGTH);
```
MosheMaorKaltura: Consider the opposite , first removing invalid chars and then trim it.

NadavHarnik: it is invalid because of the trimming (asian languages)

PR #7154, opened 2018-05-17
Update base.ini

hilak:
Update number of partner copy ACPs to 20


hilak:
Merging with approval from @erankor

PR #7186, opened 2018-05-25
Replace {deliveryCode}

hilak:
Replace delivery code in redirect headers of manifest


hilak:
Merging with approval from @etameran

PR #7204, opened 2018-05-29
PLAT-8786: Limit the concurrent rtc streams allowed to partner, and m…

gadyaari:
…ake it configurable from admin_console


plugins/conference/services/ConferenceService.php
```diff
@@ -58,7 +58,14 @@ public function allocateConferenceRoomImpl($entryId)
 		$existingConfRoom = $this->findExistingConferenceRoom($entryId);
 		if ($existingConfRoom)
 			return $existingConfRoom;
-
+		
+		$partner = $this->getPartner();
+		$numOfConcurrentRtcStreams = EntryServerNodePeer::retrieveByPartnerIdAndServerType($this->getPartnerId(), ConferencePlugin::getCoreValue('EntryServerNodeType', ConferenceEntryServerNodeType::CONFERENCE_ENTRY_SERVER ));
+		if ($numOfConcurrentRtcStreams >= $partner->getMaxLiveRtcStreamInputs())
```
noam-arad: if getMaxLiveRtcStreamInputs returns null because it was never set we should have a default
have a look at LiveStreamService.php:212

plugins/conference/services/ConferenceService.php
```diff
@@ -58,7 +58,15 @@ public function allocateConferenceRoomImpl($entryId)
...
+		
+		$partner = $this->getPartner();
+		$numOfConcurrentRtcStreams = EntryServerNodePeer::retrieveByPartnerIdAndServerType($this->getPartnerId(), ConferencePlugin::getCoreValue('EntryServerNodeType', ConferenceEntryServerNodeType::CONFERENCE_ENTRY_SERVER ));
+		// $partner->getMaxLiveRtcStreamInputs() will return the number configured for the user in the admin console, otherwise the default value - 2
+		if ($numOfConcurrentRtcStreams >= $partner->getMaxLiveRtcStreamInputs())
+		{
+			throw new KalturaAPIException(KalturaErrors::LIVE_STREAM_EXCEEDED_MAX_RTC_STREAMS, $this->getPartnerId());
+		}
+		
 		$liveStreamEntry = entryPeer::retrieveByPK($entryId);
```
yossipapi: You fetch the entry in line 30 why not pass it to this function?

api_v3/lib/KalturaErrors.php
```diff
@@ -581,6 +581,8 @@ class KalturaErrors extends APIErrors
 	
 	const LIVE_STREAM_EXCEEDED_MAX_TRANSCODED = "LIVE_STREAM_EXCEEDED_MAX_TRANSCODED;ENTRY_ID;Partner exceeded max concurrent transcoded live streams in entry[@ENTRY_ID@]";
 	
+	const LIVE_STREAM_EXCEEDED_MAX_RTC_STREAMS = "LIVE_STREAM_EXCEEDED_MAX_RTC_STREAMS;PARTNER_ID;Partner [@PARTNER_ID@] exceeded max concurrent rtc streams allowed";
```
yossipapi: MAybe its worth to add the numbers for currentCount & allowed streams

alpha/lib/model/EntryServerNodePeer.php
```diff
@@ -66,7 +66,25 @@ public static function retrieveByEntryIdAndServerType($entryId, $serverType, Pro
...
+	 * @param      string $partnerId .
+	 * @param      EntryServerNodeType $serverType .
+	 * @param      PropelPDO $con the connection to use
+	 * @return 	   int of EntryServerNodes under partnerId
+	 * @throws     kCoreException
+	 */
+	public static function retrieveByPartnerIdAndServerType($partnerId, $serverType, PropelPDO $con = null)
+	{
+		$criteria = new Criteria();
+		$criteria->add(EntryServerNodePeer::PARTNER_ID, $partnerId);
```
yossipapi: Maybe it worth adding an index for this fetch.

PR #7221, opened 2018-06-04
Added UrlTokenizer WowzaSecureToken to support token based security f…

jbrule:
…rom Wowza origin. Supports WowzaSecureToken v2 across all HTTP delivery formats


kaltura-hooks:
Hi @jbrule,

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,157 @@
...
+	
+	/**
+	 * @var string
+	 */
+	protected $paramPrefix;
+	
+	/**
+	 * @var bool
+	 */
+	protected $shouldIncludeClientIp;
```
erankor: better use the existing kUrlTokenizer/KalturaUrlTokenizer::limitIpAddress field

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,157 @@
...
+     * 
+     * @return array
+     */
+	protected function prefixParams($prefix,$params){
+		
+		if(!empty($prefix))
+        {
+            foreach($params as $key => $param)
+            {
+                if(strpos($key, $prefix) === false)
```
erankor: Better use a different array for the result, will remove the need for strpos (simply $result[$key] = $prefix . $param)

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,157 @@
...
+	}
+	
+	/**
+     * Get prefixed parameters
+     * 
+     * @return array
+     */
+	protected function prefixParams($prefix,$params){
+		
+		if(!empty($prefix))
```
erankor: Reverse this if => if (empty($prefix)) return $params

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,157 @@
...
+     * @return string
+     */
+    public function getHash($params,$url){
+	
+		$tokenKey = $this->getKey();
+		$shouldIncludeClientIp = $this->getShouldIncludeClientIp();
+				
+		$params[$tokenKey] = "";
+		
+		if($shouldIncludeClientIp == true){
```
erankor: can remove the '== true'

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,157 @@
...
+		
+        $query = trim($query, '&');
+		
+		$urlInfo = parse_url($url);
+		
+        $path = ltrim($urlInfo['path'], '/');
+        $pathItems = explode('/', $path);
+
+        $path = "";
+        foreach ($pathItems as $k => $pathItem) {
```
erankor: Can't we just remove the file name part of the URL instead of this loop?

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,157 @@
...
+            if(1 === preg_match('/(^Manifest|\.m3u8|\.f4m|\.mpd)/',$pathItem)){
+                break;
+            }
+            $path .= $pathItem;
+            if(count($pathItems)-1 != $k) {
```
erankor: Isn't this redundant considering the if that strips it after the loop?

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,157 @@
...
+                $path .= '/';
+            }
+        }
+        if(strrpos($path, '/') === strlen($path)-1) {
```
erankor: Can be replaced with rtrim, or better - remove the ltrim before the loop and just do trim here to remove both

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,157 @@
...
+		$expiryTimeFrame = $this->getWindow();
+		
+		$params = array();
+		
+		$params["starttime"] = time();
+		$params["endtime"] = time() + $expiryTimeFrame;
+		
+		$prefixedParams = $this->prefixParams($tokenPrefix,$params);
+		
+		return $url."?".http_build_query($prefixedParams,'','&').'&'.$tokenPrefix.'hash='.$this->getHash($prefixedParams,$url);
```
erankor: IMHO it would be a bit cleaner to do $prefixedParams[$tokenPrefix.'hash'] = $this->getHash($prefixedParams,$url); before the call to http_build_query

jbrule:
Eran,

All great points. I'll work on putting together a test harness too, unless those already exist for UrlTokenizers. I couldn't find anything sifting through the repos.

Thank you,
Josh

erankor:
I don't think we have tests for those, they hardly change, and I think the most important test for tokenizers is to make sure they work against the real CDN (or in this case Wowza).
Btw, unless what I'm seeing in GitHub is somehow not up to date, you didn't sign the CLA, note that we can't merge without it.

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
jbrule: I can do that but from what I can tell KalturaUrlTokenizer::limitIpAddress is intended to enforce IP based ACLs, via Access Control Profile I assume, which is not what this option is doing. This option includes the Client IP address in the hash generation like [Vnpt tokenzier](https://github.com/jbrule/server/blob/85031093e3e3f591b956d641781ffbafae47f5d5/alpha/apps/kaltura/lib/storage/urlTokenizers/kVnptUrlTokenizer.php)

--Josh

erankor: No, it has the exact same meaning as you have here - it determines whether the token is signed with the client ip address or not. You can see for example - kCloudFrontUrlTokenizer::generateToken

admin_console/configs/lang/en.php
```diff
@@ -1343,6 +1343,7 @@
 	'Kaltura_Client_Type_UrlTokenizerVnpt'        => 'Vnpt Tokenizer',
 	'Kaltura_Client_Type_UrlTokenizerCht'		=> 'CHT Tokenizer',
 	'Kaltura_Client_Type_UrlTokenizerChinaCache'		=> 'ChinaCache Tokenizer',
+    'Kaltura_Client_Type_UrlTokenizerWowzaSecureToken'		=> 'Wowza SecureToken v2',
```
erankor: Please align the indentation - adjacent lines use tabs

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,132 @@
...
+	/**
+	 * Get prefixed parameters
+	 * 
+	 * @return array
+	 */
+	private function prefixParams($prefix, $rawParameters)
+	{
+		$prefixedParameters = array();
+		
+		if(empty($rawParameters))
```
erankor: Looking at this now, we can remove this if, it will never by empty, and even if it will, the result will be correct

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,132 @@
...
+			{
+				$query .= '=' . $v;
+			}
+		}
+		
+		$query = trim($query,'&');
+		
+		$path = '';
+		
+		if($parsed_url = parse_url($url))
```
erankor: Since you care only about the path, better pass PHP_URL_PATH to parse_url. Also, the rest of the code here uses lower camel convention while here you use underscore (parsed_url)

alpha/apps/kaltura/lib/storage/urlTokenizers/kWowzaSecureTokenUrlTokenizer.php
```diff
@@ -0,0 +1,132 @@
...
+		{
+			$path = $parsed_url["path"];
+		}
+		
+		$finalSlash = strrpos($path, '/');
+		
+		$filepath = $path = substr($path, 0, $finalSlash);
```
erankor: filepath is unused

jbrule:
Requested changes made. Updated to current active development branch.

erankor:
Merged, thanks

PR #7257, opened 2018-06-10
PLAT-8976  phase 1 - NFR identify local url calls

gotlieb:
@MosheMaorKaltura 
@erankor


infra/general/KCurlWrapper.class.php
```diff
@@ -498,6 +498,9 @@ public function getHeader($sourceUrl, $noBody = false)
 	 */
 	public function exec($sourceUrl, $destFile = null,$progressCallBack = null)
 	{
+		if ($this->isInternalUrl($sourceUrl))
```
MosheMaorKaltura: Need to extract the target host from the URL before checking if it is internal , same as done in setSourceUrlAndprotocol.

PR #7258, opened 2018-06-10
PLAT-8952: checking times before copping cue point in live Clipping e…

david-winder-kaltura:
…ngine


MosheMaorKaltura:
Reviewed

PR #7280, opened 2018-06-13
PLAT-9005: enable delete of temporary recording 

david-winder-kaltura:


api_v3/lib/KalturaEntryService.php
```diff
@@ -745,7 +745,10 @@ protected function attachOperationResource(kOperationResource $resource, entry $
 	protected function handleLiveClippingFlow($recordedEntry, $clippedEntry, $operationAttributes)
 	{
 		if (($recordedEntry->getId() == $clippedEntry->getId()) || ($recordedEntry->getId() == $clippedEntry->getReplacedEntryId()))
+		{
+			$this->cancelReplace($recordedEntry->getId(), $recordedEntry->getType());
```
yossipapi: IMHO this is not right to interfere with the basic flow here.
If we are not doing this in other flows lets not do it here.

PR #7282, opened 2018-06-13
PLAT-6872: remove Exclude source from admin_console

david-winder-kaltura:


admin_console/forms/PartnerConfiguration.php
```diff
@@ -112,11 +112,6 @@ public function init()
 				'label'	  => 'Default Entitlement Enforcement',
 				'decorators' => array('ViewHelper', array('Label', array('placement' => 'append')), array('HtmlTag',  array('tag' => 'dt', 'class' => 'partner_configuration_checkbox_field')))
 			));
-
-		$this->addElement('checkbox', 'time_aligned_renditions', array(
```
yossipapi: This flag is being used here: alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php:989

So why remove this option from the admin console?

kalturaguy: @yossipapi  the player doesn't send this property playerType  anymore
the code is commented out in the mwembed since 2015...

PR #7312, opened 2018-06-21
PLAT-9035: fix fatal

noam-arad:


plugins/conference/lib/model/ConferenceEntryServerNode.php
```diff
@@ -87,4 +87,9 @@ public function setLastAllocationTime($v)
 		$this->putInCustomData(self::CUSTOM_DATA_LAST_ALLOCATE_TIME, $v);
 	}
 
+	public function isValid()
+	{
+		return (time() - kConf::get('conf_not_finished_timeout', 'local', 0)) <= $this->getLastAllocationTime();
```
david-winder-kaltura: Make sure to add conf_not_finished_timeout default in conf files

MosheMaorKaltura: I suggest adding default value 0 to getLastAllocationTime , otherwise you will be comparing int with null.

PR #7319, opened 2018-06-24
No Plat: take matchType value when updating metadata profile as well use int value instead of passed string value

yossipapi:


MosheMaorKaltura:
reviewed

PR #7343, opened 2018-06-28
This alias will never match because of the one defined in line 26

jessp01:
Alias /apps "/opt/kaltura/apps"

All it does is trigger the following warning:

> The Alias directive in
> /opt/kaltura/app/configurations/apache/conf.d/enabled.apps.conf at line
> 38 will probably never match because it overlaps an earlier Alias.


jessp01:
Closed in favour of https://github.com/kaltura/server/pull/7957/files

PR #7372, opened 2018-07-04
PLAT-9076 use kmc_secured_login flag to determine if should enforce https

esakal:


MosheMaorKaltura:
Reviewed - please notice that this pull is for 14.2.0 , while coming deployment is 14.3.0.

esakal:
not needed anymore

PR #7378, opened 2018-07-05
Fix Curl Wrapper responce for Http notification

RoieBeck:


infra/general/KCurlWrapper.class.php
```diff
@@ -304,12 +304,15 @@ public function getError()
 
 		return $err;
 	}
-	
+
 	/**
+	 * @param $opt
 	 * @return string
```
MosheMaorKaltura: Fix annotation

PR #7393, opened 2018-07-09
Revert "SUPPS-1476"

hilak:
Reverts kaltura/server#7392


hilak:
Reverting - will defer the fix of the issue until the resolution of PSVAMB-3678.

PR #7406, opened 2018-07-12
PLAT-8914: fix undefined offset for values returned from CB in order …

inbal-ben-david:
…to enter to the check that they are valid

### DIFF
--- a/api_v3/lib/KalturaResponseProfileCacher.php
+++ b/api_v3/lib/KalturaResponseProfileCacher.php
@@ -139,18 +139,20 @@ class KalturaResponseProfileCacher extends kResponseProfileCacher
 		$key = self::getObjectSpecificCacheKey($object, $responseProfileKey);
 		$responseProfileCacheKey = self::getResponseProfileCacheKey($responseProfileKey, $object->getPartnerId());
 
-		$results = self::get(array($key, $responseProfileCacheKey));
-		$value = array_slice($results, 0, 1);
-		$responseProfileCache = array_slice($results, 1, 1);
+		$cachedResults = self::get(array($key, $responseProfileCacheKey));
+		$values = array_values($cachedResults);
+		$objectCache = $values[0];
+		$responseProfileCache = $values[1];
+
 		
 		$invalidationKeys = array(
 			self::getObjectKey($object),
 			self::getRelatedObjectKey($object, $responseProfileKey),
 		);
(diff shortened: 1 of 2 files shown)


api_v3/lib/KalturaResponseProfileCacher.php
```diff
@@ -138,17 +138,32 @@ public static function start(KalturaObject $apiObject, IRelatedObject $object, K
...
+			if(sizeof($cachedResults) < 2)
+				KalturaLog::debug("Missing values. Only the following values returned: " . print_r($cachedResults));
+			else
+			{
+				$objectCache = $values[0];
+				$responseProfileCache = $values[1];
+			}
+		}
+		else
+			KalturaLog::debug("Missing records for given keys");
```
MosheMaorKaltura: Specify the keys for later debug

api_v3/lib/KalturaResponseProfileCacher.php
```diff
@@ -138,17 +138,32 @@ public static function start(KalturaObject $apiObject, IRelatedObject $object, K
...
+		$responseProfileCache = null;
+		$cachedResults = self::get(array($key, $responseProfileCacheKey));
+		if($cachedResults)
+		{
+			$values = array_values($cachedResults);
+			if(sizeof($cachedResults) < 2)
+				KalturaLog::debug("Missing values. Only the following values returned: " . print_r($cachedResults));
+			else
+			{
+				$objectCache = $values[0];
```
MosheMaorKaltura: Use const for 0 & 1 , fixed location inside the response.

api_v3/lib/KalturaResponseProfileCacher.php
```diff
@@ -138,17 +138,32 @@ public static function start(KalturaObject $apiObject, IRelatedObject $object, K
...
-		list($value, $responseProfileCache) = self::get(array($key, $responseProfileCacheKey));
-		
+
+		$objectCache = null;
+		$responseProfileCache = null;
+		$cachedResults = self::get(array($key, $responseProfileCacheKey));
+		if($cachedResults)
+		{
+			$values = array_values($cachedResults);
+			if(sizeof($cachedResults) < 2)
```
MosheMaorKaltura: For a better performance check the most common flow first (in this case it is the positive flow).

PR #7411, opened 2018-07-15
Naos 14.3.0 plat 8248

RoieBeck:


alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
```diff
@@ -761,4 +760,21 @@ protected function getBestMatchFlavor($groupedFlavors, $referenceFlavor)
...
+	 * @param string $path
+	 * @param int $flavorId
+	 * @return array
+	 */
+	private function getClipData($path, $flavorId)
+	{
+		$mediaInfo = mediaInfoPeer::retrieveByFlavorAssetId($flavorId);
+		$clipDesc = array('type' => 'source', 'path' => $path);
+		if ($mediaInfo && !$mediaInfo->getAudioBitRate())
+		{
```
MosheMaorKaltura: Please add a debug info here, so we can cover bizarre cases where there is audio track but not bit rate is set

PR #7428, opened 2018-07-17
PLAT-9057: add disable populate elastic indexing by partner id

NadavHarnik:


plugins/search/lib/model/SphinxLogPeer.php
```diff
@@ -52,9 +52,18 @@ public static function retrieveByLastId(array $servers, $gap = 0, $limit = 1000,
 		
 		if($criterions)
 			$criteria->addAnd($criterions);
-			
-		$disabledPartnerIds = kConf::get('disable_sphinx_indexing_partners', 'local', array());
-		if ($disabledPartnerIds)
+
+		$disabledPartnerIds = array();
+		if (in_array($type ,array(SphinxLogType::SPHINX)))
```
MosheMaorKaltura: Why do need in_array?  since its only one value , you can just check $type==

PR #7464, opened 2018-07-25
PLAT-9121: add priority logic to the entry-server node choose

david-winder-kaltura:


alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -111,65 +111,55 @@ protected function initLiveStreamConfig()
...
-				$this->liveStreamConfig->setBackupStreamInfo($liveEntryServerNode->getStreams());
-			}
+	}
+
+	private function getIsMaintenanceFromCache($entryId)
+	{
+		$cache = kCacheManager::getSingleLayerCache(kCacheManager::CACHE_TYPE_PLAY_MANIFEST);
+		if ($cache) {
+			$result = json_decode($cache->get("MaintenanceDataCacheKey"), true);
+			KalturaLog::debug("Got maintenance data from cache: " . print_r($result, true));
```
kalturaguy: 1.will add a check if this key even exists...
2.rename MaintenanceDataCacheKey to **Live-**MaintenanceData

alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -111,65 +111,55 @@ protected function initLiveStreamConfig()
...
+			if (key_exists("MaintenanceDC", $result))
```
kalturaguy: should start with lower case (JSON is camel case)

alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -111,65 +111,55 @@ protected function initLiveStreamConfig()
...
+				return $result["MaintenanceDC"];
+			if (key_exists($entryId, $result))
```
kalturaguy: should be per dc

PR #7467, opened 2018-07-25
SUP-14981:due to Itunes limitations removing unneccesry fields

ravitshalem:


MosheMaorKaltura:
Tested on staging server - play out from itunes.

PR #7473, opened 2018-07-26
PLAT-9122: allow returning indexed array with CB extracted values + f…

inbal-ben-david:
…ix debug prints


MosheMaorKaltura:
Reviewed

PR #7495, opened 2018-08-01
PLAT-8944: Add gpu support for wowza machine

david-winder-kaltura:


plugins/media/wowza/services/LiveConversionProfileService.php
```diff
@@ -210,6 +219,11 @@ public function serveAction($streamName, $hostname = null, $extraParams = null)
 		
 		return new kRendererString($dom->saveXML(), 'text/xml');
 	}
+
+	private function isGpuSupported($streamParametersArray)
+	{
+		return $streamParametersArray['gpu'] == 'true';
```
yossipapi: check if $streamParametersArray['gpu'] is defined to avoid getting notice message in the log.

PR #7496, opened 2018-08-05
Update local.template.ini

esakal:
remove unused keys and add keys for KEA and Reach.

@gonenradai do you want to include the keys for KEA and Reach or to remove them from the template as?

### DIFF
--- a/configurations/local.template.ini
+++ b/configurations/local.template.ini
@@ -220,8 +220,8 @@ play_kit_js_cache_version = 1
 1 = _KDP_RE_PL
 
 [kmcng]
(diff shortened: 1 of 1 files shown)


DBezemer:
@esakal please keep kmcng_version in this file as we replace it based on the correct version when installing. For the kea/reach you should probably add these to base.ini like kmcng_version and then add
```
; kmcng_kea_version = @KMCNG_KEA_VERSION@
;kmcng_reach_version = @KMCNG_REACH_VERSION@
```
to local.template.ini

esakal:
closing, not needed

PR #7513, opened 2018-08-13
PLAT-9012 - multiple admin secrets support

RoieBeck:


alpha/lib/model/Partner.php
```diff
@@ -685,7 +687,46 @@ public function setDefaultEmbedCodeType( $v ) { $this->putInCustomData("defaultE
...
+
+	/**
+	 * @param kAdditionalAdminSecrets $v
+	 */
+	public function setAdditionalAdminSecrets($v) {
+		$this->putInCustomData( "additionalAdminSecret",serialize($v) );
+	}
+
+	/**
+	 * @return array|null
```
MosheMaorKaltura: Change return type to array  only (empty array instead of null).

PR #7533, opened 2018-08-22
PLAT-9163 handle ingested flavors with language different than kaltur…

gotlieb:
…a language name

@MosheMaorKaltura


MosheMaorKaltura:
reviewed

PR #7534, opened 2018-08-23
 PLAT-7981: add search history plugin

NadavHarnik:


configurations/logstash/kaltura_search_history.template.conf
```diff
@@ -0,0 +1,91 @@
...
+output
+{
+        file
+        {
+                path => "/var/log/kaltura_search_history_output.log"
+                flush_interval => 5
+                file_mode => 0644
+        }
+
+        if [@metadata][_id]
```
MosheMaorKaltura: No need to handle [id]

PR #7550, opened 2018-08-30
Naos-14.6.0-SUP-15117 Change account from multi user management console

ZurKaltura:
listPartnersForUserAction doesn't work with ks with a master partner id different then its partner id


MosheMaorKaltura:
reviewed

PR #7558, opened 2018-09-03
add Access-Control-Max-Age header to API calls

muli:
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age.

This change should reduce the number of preflight `OPTIONS` requests to the server significantly, especially with a client-side application, e.g., KMC.


muli:
@MosheMaorKaltura - ping.

PR #7562, opened 2018-09-03
PLAT-9220: ecdn super-redundancy initial POC

noam-arad:


alpha/lib/model/DeliveryProfileVodPackagerHls.php
```diff
@@ -34,4 +34,35 @@ public function getAllowFairplayOffline()
 		return $this->getFromCustomData("allowFairplayOffline", null, false);
 	}
 
+	/**
+	 * @param bool $oneOnly
+	 * @return array
+	 */
+	protected function buildHttpFlavorsArray()
```
yossipapi: This code adds the feature support only for HLS shouldn't this be added for all formats?
If so better apply the logic in: alpha/lib/model/DeliveryProfile.php:466

PR #7575, opened 2018-09-06
PLAT-9212: add ability to get full path of nodeserver, also to set th…

noam-arad:
…e status during reportStatus


yossipapi:
Add release notes

PR #7580, opened 2018-09-12
Add check for local file exist in media add and update content action

david-winder-kaltura:


api_v3/services/MediaService.php
```diff
@@ -112,10 +112,10 @@ function addContentAction($entryId, KalturaResource $resource = null)
 
 		if ($resource)
 		{
-			$resource->validateEntry($dbEntry);
-			$kResource = $resource->toObject();
 			try
 			{
+				$resource->validateEntry($dbEntry, true);
```
MosheMaorKaltura: In case of exception EntryIdNotFound , there is no sense in canceling replace that is done inside handleErrorDuringSetResource please handle.

api_v3/lib/types/resource/KalturaEntryResource.php
```diff
@@ -36,7 +36,41 @@ public function validateForUsage($sourceObject, $propertiesToSkip = array())
...
+	 * @see KalturaResource::validateEntry()
+	 */
+	public function validateEntry(entry $dbEntry, $validateLocal = false)
+	{
+		parent::validateEntry($dbEntry, $validateLocal);
 		
+		$srcEntry = entryPeer::retrieveByPK($this->entryId);
+		if(!$srcEntry)
+			throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $this->entryId);
+		if ($validateLocal && !$this->checkIfFileExist(true))
```
MosheMaorKaltura: Since you are in API , no sense in raising kCoreException.
try - throw new KalturaAPIException(APIErrors::SOURCE_FILE_REMOTE); ......The source file is remote, no local file found.....

api_v3/services/MediaService.php
```diff
@@ -1227,9 +1227,14 @@ function getVolumeMapAction($entryId)
...
 	private function handleErrorDuringSetResource($entryId, Exception $e)
 	{
+		if ($e->getCode() == KalturaErrors::ENTRY_ID_NOT_FOUND)
+			throw $e; //if no entry found then no need to do anything
+
 		KalturaLog::info("Exception was thrown during setContent on entry [$entryId] with error: " . $e->getMessage());
 		$this->cancelReplaceAction($entryId);
-		if (($e->getCode() == kCoreException::SOURCE_FILE_NOT_FOUND) && (kDataCenterMgr::dcExists(1 - kDataCenterMgr::getCurrentDcId())))
+
+		$errorArr = array(kCoreException::SOURCE_FILE_NOT_FOUND, KalturaErrors::SOURCE_FILE_NOT_FOUND);
```
noam-arad: KalturaErrors::SOURCE_FILE_NOT_FOUND is not the error code but has the whole string, only the first string before the ';' is considered the error code

PR #7586, opened 2018-09-16
PLAT-9248: Use custom sort function to sort kFile::list Dir result since it returns array of array's

yossipapi:


MosheMaorKaltura:
reviewed

PR #7587, opened 2018-09-16
Naos 14.6.0 plat 9190

RoieBeck:


plugins/vendor/services/ZoomVendorService.php
```diff
@@ -0,0 +1,437 @@
...
+				$zoomIntegration->setStatus(VendorStatus::ACTIVE);
+			$this->saveNewTokenData($tokens, $accountId, $zoomIntegration);
+			$permissions = $permissions['permissions'];
+			$isAdmin = $this->canConfigureEventSubscription($permissions);
+		}
+		if ($isAdmin)
+		{
+			$this->loadLoginPage($tokens);
+		}
+		return false;
```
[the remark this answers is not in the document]
MosheMaorKaltura: Throw API exception - only admin

deployment/updates/sql/2018_08_28_create_vendor_integration_table.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE  IF NOT EXISTS `vendor_integration` (
+	`id` int  NOT NULL AUTO_INCREMENT,
+	`account_id` varchar(64) NOT NULL,
+	`partner_id` int NOT NULL,
+	`vendor_Type` smallint(6) NOT NULL,
+	`custom_data` text,
+	`status` smallint(6) NOT NULL,
```
yossipapi: Why not define status as tinyint?

plugins/vendor/lib/model/VendorIntegration.php
```diff
@@ -0,0 +1,62 @@
...
+ * You should add additional methods to this class to meet the
+ * application requirements.  This class will only be generated as
+ * long as it does not already exist in the output directory.
+ *
+ * @package plugins.vendor
+ * @subpackage model
+ */
+class VendorIntegration extends BaseVendorIntegration {
+
+	public function setAccessToken ($v)	{ $this->putInCustomData ( "accessToken" , $v);	}
```
yossipapi: Move strings such as "accessToken" & "refreshToken" & etc ... to be const on the class

plugins/vendor/lib/model/VendorIntegration.php
```diff
@@ -0,0 +1,62 @@
...
+	public function getZoomCategory( )	{ return $this->getFromCustomData( "zoomCategory" );	}
+
+
+	/**
+	 * returns all tokens as array
+	 * @return array
+	 */
+	public function getTokens()
+	{
+		return array(kZoomOauth::ACCESS_TOKEN => $this->getAccessToken(), kZoomOauth::REFRESH_TOKEN => $this->getRefreshToken(),
```
yossipapi: Any reason why not to create an extended class called ZoomVendorIntegration.php instead of referring zoom related info in the generic class? applies to saveNewTokenData as well

plugins/vendor/lib/model/VendorIntegrationPeer.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * You should add additional methods to this class to meet the
+ * application requirements.  This class will only be generated as
+ * long as it does not already exist in the output directory.
+ *
+ * @package plugins.vendor
+ * @subpackage model
+ */
+class VendorIntegrationPeer extends BaseVendorIntegrationPeer {
+
```
yossipapi: I am missing the setDefaultCriteriaFilter method which will filter out any objects that are in statuses which you do not want to get back in the result.

plugins/vendor/lib/model/VendorIntegrationPeer.php
```diff
@@ -0,0 +1,36 @@
...
+ * 
+ *
+ * You should add additional methods to this class to meet the
+ * application requirements.  This class will only be generated as
+ * long as it does not already exist in the output directory.
+ *
+ * @package plugins.vendor
+ * @subpackage model
+ */
+class VendorIntegrationPeer extends BaseVendorIntegrationPeer {
```
yossipapi: In my opinion, this class should be abstract.

plugins/vendor/lib/model/zoom/RetrieveDataFromZoom.php
```diff
@@ -0,0 +1,101 @@
+<?php
+/**
+ * @package plugins.venodr
+ * @subpackage model.zoom
+ */
+class RetrieveDataFromZoom
```
yossipapi: In my opinion, a better name would be zoomWrapper and this class should contain any methods that interact with the zoom API.

plugins/vendor/services/ZoomVendorService.php
```diff
@@ -0,0 +1,431 @@
+<?php
+/**
+ * @service zoomVendor
+ * @package plugins.vendor
+ * @subpackage api.services
+ */
+class ZoomVendorService extends KalturaBaseService
```
yossipapi: Shouldn't this service be the entry point for other vendor integrations in the future as well?
If implementing it this way we will need to create a new service class for every vendor integration we will do in the future.

yossipapi: I don't see you are applying the partner filter on the criteria this is why you are always doing:
$partnerId = kCurrentContext::getCurrentPartnerId();
		$zoomIntegration = VendorIntegrationPeer::retrieveSingleVendorPerPartner($accountId,
			VendorTypeEnum::ZOOM_ACCOUNT, $partnerId);

Is there a reason for this?

plugins/vendor/services/ZoomVendorService.php
```diff
@@ -0,0 +1,431 @@
...
+
+	/**
+	 * @param kuser $dbUser
+	 * @param string $zoomCategory
+	 * @param $emails
+	 * @param $meetingId
+	 * @return string
+	 * @throws Exception
+	 */
+	private function createEntryForZoom($dbUser, $zoomCategory, $emails, $meetingId)
```
yossipapi: In my opinion, it would be better to move all these private functions to a helper class if possible.
Generally, the service class contains only the API calls that are getting exposed and the logic is moved to a class that exposes static functions which you can call.

plugins/vendor/lib/model/VendorIntegrationPeer.php
RoieBeck: generated automatically via propel

plugins/vendor/lib/model/VendorIntegrationPeer.php
RoieBeck: generated automatically via propel

plugins/vendor/services/ZoomVendorService.php
RoieBeck: After talking about it -> agreed to keep like I suggested

RoieBeck: done

PR #7588, opened 2018-09-16
PLAT-9235 make userGroup sync go via api calls and via batch(if neede…

gotlieb:
…d in case of removing from group or creating new groups)

@MosheMaorKaltura

### DIFF
--- a/plugins/bulk_upload/base/services/BulkService.php
+++ b/plugins/bulk_upload/base/services/BulkService.php
@@ -11,6 +11,7 @@ class BulkService extends KalturaBaseService
 	const PARTNER_DEFAULT_CONVERSION_PROFILE_ID = -1;
 	
 	const SERVICE_NAME = "bulkUpload";
+	const USER_GROUP_SYNC_THRESHOLD_DEFUALT = "50";
 
 	/**
 	 * Add new bulk upload batch job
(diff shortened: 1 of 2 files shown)


plugins/bulk_upload/base/services/BulkService.php
```diff
@@ -445,21 +445,72 @@ function abortAction($id)
...
+		$bulkUpload = null;
+
+		$kUser = kuserPeer::getKuserByPartnerAndUid($this->getPartnerId(), $userId);
+		if (!$kUser || $kUser->getType() != KuserType::USER)
+		{
 			throw new KalturaAPIException(KalturaErrors::INVALID_USER_ID, $userId);
+		}
 
-		if(!$groupIds)
+		if (!$groupIds)
```
MosheMaorKaltura: No need to validate that value exist since this parameter does not have default value, user cannot call this action without sending it.

plugins/bulk_upload/base/services/BulkService.php
```diff
@@ -445,21 +445,72 @@ function abortAction($id)
 	 */
 	public function syncGroupUsersAction($userId, $groupIds, $removeFromExistingGroups = true, $createNewGroups = true)
 	{
-		$kuser = kuserPeer::getKuserByPartnerAndUid($this->getPartnerId(), $userId);
-		if (!$kuser || $kuser->getType() != KuserType::USER)
+		$groupLimit = kConf::get('user_groups_sync_threshold','local',50);
```
MosheMaorKaltura: Move the 50 to a defined const value.

PR #7591, opened 2018-09-16
Support for image entries to be exported

abrararshad:
Currently only entries of type video are being exported to an external storage. This adds support for image entries to be exported as well. 

Basically any straight/non-flavor entry type could be exported by adding its type to `$entryTypes` array in `kFlowManager.php` under `shouldConsumeAddedEvent` function but in this commit only image type is added and tested


jessp01:
Hi @abrararshad,

Thank you for your pull. Can you please stick with the same tabstop as the original code in order to reduce the diff?

abrararshad:
Hi @jessp01 ,

PHPStorm did that automatically before. I used different IDE this time. However, unchanged code under `objectAdded` is wrapped by an `IF` statement but not indented for easy review. I will do the indentation after the review, if you like

Thanks

abrararshad:
Sorry, pushed again. Just added some spacing between lines

jessp01:
Hi @abrararshad,

I'm not sure I understand why you added code to `kFlowManager.php`.
This should be handled in `kStorageExporter::exportEntry()` in alpha/apps/kaltura/lib/storage/kStorageExporter.php

Also, 

```php
$entry = entryPeer::retrieveByPK($object->getObjectId());
            
            if (!$entry instanceof entry)
                return true;
```
not sure I understand the purpose of this check? What object can entryPeer::retrieveByPK() possibly return other than entry?

Thanks,

abrararshad:
Hi @jessp01 ,

Ohh!! I overlooked and didn't see this function `kStorageExporter::exportEntry()`. Definitely this function needs to be called for exporting entries and i am refactoring the code and will push again.

As per my understanding of Kaltura so far, the first encounter of starting/creating batch jobs for media entries is in `kFlowManager.php` such as `kJobsManager::addConvertProfileJob` is the first job created for video entry.  So I tried to keep the similar logic around. Thats the reason I have put the logic for image/non-flavor entries in the same place where video entry's logic is. I may be wrong, please advise.

About this condition `if (!$entry instanceof entry)` i just meant to check if the entry is NOT null. I believe it could return `null` from `BaseentryPeer::retrieveByPK`. Otherwise you are totally right

Thanks

abrararshad:
Code is now a lot less and changes are tested. Awaiting if there are more improvements.

Thanks

abrararshad:
Hi @jessp01 ,

I am sure it has been quite busy for you as well. I am sorry but is there any update on this? This is a blocker in our project and we have more issues mostly related to image media entries, I am planning to share with you once i get the exact cause of those.

Thanks

jessp01:
Hi @abrararshad,

Sorry but I still don't understand why you've changed the code in alpha/apps/kaltura/lib/batch2/kFlowManager.php. The change should be made in https://github.com/kaltura/server/blob/Naos-14.6.0/alpha/apps/kaltura/lib/storage/kStorageExporter.php

You'll notice that `exportEntry()` calls:
`exportFlavorAsset()` [twice, once for the results returned from `assetPeer::retrieveFlavorsByEntryIdAndStatus()` and once for these returned from `assetPeer::retrieveReadyThumbnailsByEntryId()`]
 `exportAdditionalEntryFiles()`

You just need to add code that exports the image file itself, in the event the Kaltura Entry is of type `IMAGE` since naturally, such an entry will not have flavours as it isn't a video.

abrararshad:
New pull request: https://github.com/kaltura/server/pull/7605

PR #7604, opened 2018-09-23
PLAT-9256: make sure thumbnail actions are cached for KES rules

noam-arad:


alpha/lib/model/accessControl.php
```diff
@@ -156,7 +156,20 @@ public function hasRules($contextType = null)
...
-	
+
+	public function hasKESRule()
+	{
+		$rules = $this->getRulesArray();
+
+		foreach($rules as $rule)
+		{
+			/* @var $rule kRule */
+			if (json_decode($rule->getRuleData())->type == 'ServeFromKESRule')
```
kalturaguy: i think it's better to check if one of the action is KalturaAccessControlServeRemoteEdgeServerAction

+ return false at the end of the method

kalturaguy: also for performance issue, let's save the result in the custom_data when you save the ACP, and just check the custom_data in the action

alpha/lib/model/accessControl.php
```diff
@@ -309,4 +330,37 @@ public function getCacheInvalidationKeys()
...
+		foreach ($rules as $rule)
+		{
+			/* @var $rule kRule */
+			$actions = $rule->getActions();
+			foreach ($actions as $action)
+			{
+				/* @var kRuleAction $action */
+				if ($action->getType() == RuleActionType::SERVE_FROM_REMOTE_SERVER)
+				{
+					$hasKESRule = true;
```
yossipapi: You can probably break from the loop after hasKESRule was assigned with true.

alpha/lib/model/accessControl.php
```diff
@@ -309,4 +330,37 @@ public function getCacheInvalidationKeys()
...
+
+	public function setHasKES($v)
+	{
+		$this->putInCustomData(self::CUSTOM_DATA_HAS_KES_RULE, $v);
+	}
+
+	/**
+	 * @return bool
+	 */
+	protected function calculateHasKESRule(): bool
```
yossipapi: Remove the ": bool" from the function signature

alpha/apps/kaltura/modules/extwidget/actions/thumbnailAction.class.php
```diff
@@ -453,8 +453,8 @@ public function execute()
 		}
 		
 		$nocache = false;
-		if ($secureEntryHelper->shouldDisableCache() || kApiCache::hasExtraFields() ||
-			(!$secureEntryHelper->isKsWidget() && $secureEntryHelper->hasRules(ContextType::THUMBNAIL)))
+		if (!$secureEntryHelper->hasKESRule() && ($secureEntryHelper->shouldDisableCache() || kApiCache::hasExtraFields() ||
```
yossipapi: I would probably change all the places where you named the functions or variables from the word KES to remoteServer (i.e hasKESRule hasServeRemoteServerRule). KES is not a terminology the Kaltura server should know about.

alpha/lib/model/accessControl.php
```diff
@@ -309,4 +330,38 @@ public function getCacheInvalidationKeys()
 	{
 		return array("accessControl:id=".strtolower($this->getId()));
 	}
+
+	public function getHasRemoteServer()
```
yossipapi: The function name is not that suitable, should be something like getHasServeRemoteServerRule same applies to other method signature as well.

PR #7608, opened 2018-09-25
Support custom thumbnails for image entries

abrararshad:
This adds support for image entries to have custom thumbnails and changes are made accordingly when fetching media metadata or calling thumbnail service.

Changes are tested with external storage enabled and disabled. It is working in both cases. 
However, in case of external storage enabled and once the file is exported, on the KMC content tab, thumbnails for image entries disappear because the URLs become like this "http://xxxxx.cloudfront.net/content/entry/data/0/0/0_j8y8fxw8_100002.jpg/bgcolor/F7F7F7/type/2". I couldn't look into this as KMC is in flash

Thanks

### DIFF
--- a/alpha/apps/kaltura/lib/batch2/bcdl/kBusinessConvertDL.php
+++ b/alpha/apps/kaltura/lib/batch2/bcdl/kBusinessConvertDL.php
@@ -300,15 +300,15 @@ class kBusinessConvertDL
 			$thumbAsset->addTags(array(thumbParams::TAG_DEFAULT_THUMB));
 			$thumbAsset->save();
 			KalturaLog::info("Setting entry [". $thumbAsset->getEntryId() ."] default ThumbAsset to [". $thumbAsset->getId() ."]");
+
+			$entry->setThumbnail(".jpg");
+			$entry->setCreateThumb(false, $thumbAsset);
+			$entry->save();
+
+			$thumbSyncKey = $thumbAsset->getSyncKey(thumbAsset::FILE_SYNC_FLAVOR_ASSET_SUB_TYPE_ASSET);
+			$entrySyncKey = $entry->getSyncKey(entry::FILE_SYNC_ENTRY_SUB_TYPE_THUMB);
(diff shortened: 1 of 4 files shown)


alpha/lib/model/entry.php
```diff
@@ -760,15 +760,24 @@ public function getDataUrl( $version = NULL )
 		}
 		else if ($media_type == self::ENTRY_MEDIA_TYPE_IMAGE  )
 		{
-			$width = self::DEFAULT_IMAGE_WIDTH;
-			$height = self::DEFAULT_IMAGE_HEIGHT;
+			$dataKey = $this->getSyncKey(self::ENTRY_MEDIA_SOURCE_FILE);
```
abrararshad: getDataUrl() should try to return the actual data file not thumbnail just like videos. If can't get the file then fall back to the thumbnail

alpha/lib/model/entry.php
```diff
@@ -760,15 +760,24 @@ public function getDataUrl( $version = NULL )
...
+			$file_sync = kFileSyncUtils::getReadyExternalFileSyncForKey($dataKey);
+			if($file_sync)
+			{
+				$url =  $file_sync->getExternalUrl($this->getId());
```
abrararshad: getDataUrl() should try to return the actual data file not thumbnail just like videos. If can't get the file then fall back to the thumbnail

alpha/apps/kaltura/lib/batch2/bcdl/kBusinessConvertDL.php
```diff
@@ -300,15 +300,15 @@ public static function setAsDefaultThumbAsset($thumbAsset)
 			$thumbAsset->addTags(array(thumbParams::TAG_DEFAULT_THUMB));
 			$thumbAsset->save();
 			KalturaLog::info("Setting entry [". $thumbAsset->getEntryId() ."] default ThumbAsset to [". $thumbAsset->getId() ."]");
-		}
 
-		$entry->setThumbnail(".jpg");
```
abrararshad: If the thumbnail is already a `default` one, the code starting from this line was being executed and adding more similar entries into file_sync table. So moved it up under the IF statement

PR #7619, opened 2018-09-30
PLAT-8342: limit reset password per email and ip to avoid flooding

inbal-ben-david:


infra/general/CountingReservation.php
```diff
@@ -0,0 +1,96 @@
+<?php
+/**
+ *
+ */
+
+class CountingReservation
+{
+	CONST DEFAULT_TIME_IN_CACHE_FOR_RESERVATION = 300;
```
MosheMaorKaltura: CONST should be lowercase.

infra/general/CountingReservation.php
```diff
@@ -0,0 +1,96 @@
...
+
+	/**
+	 * will try to acquire the resource, if we tried more then max times we will lock the resource
+	 * for the ttl set up for the reservator
+	 * @param string $resourceId
+	 * @return bool
+	 */
+	public function tryAcquire($resourceId)
+	{
+		$cacheCounter = $this->cache->get($this->getCacheKeyForResource($resourceId));
```
MosheMaorKaltura: Use Memcache::decrement  since it is atomic while get..store is not.

api_v3/lib/KalturaBaseUserService.php
```diff
@@ -293,6 +294,60 @@ protected function validateApiAccessControlByEmail($email)
...
 	}
+
+	/**
+	 * check if there were more than max_allowed calls for different resources
+	 * @param string $email
+	 * @throws KalturaErrors::FAILED_TO_INIT_OBJECT
+	 */
+	protected function validateRequestsAmount($email)
+	{
+		$cache = kCacheManager::getSingleLayerCache(kCacheManager::CACHE_TYPE_RESOURCE_RESERVATION);
```
MosheMaorKaltura: Add release note which states that this value must be set in cache.ini (for onprem users that did not set it so far)

PR #7629, opened 2018-10-04
kava - support filtering by customVar1

erankor:


alpha/apps/kaltura/lib/reports/myReportsMgr.class.php
```diff
@@ -1503,6 +1503,7 @@ class reportsInputFilter
 	public $extra_map;
 	public $categories;
 	public $categoriesIds;
+	public $custom_var1;
```
MosheMaorKaltura: Better use one format of member names, e.g. lowerCamelCase.

erankor: Yes, I saw that... however, this class already uses mixed conventions (before this pull) so I just went with the majority for now, which is underscores

PR #7678, opened 2018-10-14
Naos 14.6.0 plat 9228

noam-arad:


plugins/conference/services/ConferenceService.php
```diff
@@ -8,18 +8,20 @@
...
 	 * @action allocateConferenceRoom
 	 * @actionAlias liveStream.allocateConferenceRoom
 	 * @param string $entryId
+	 * @param string $tag
 	 * @return KalturaRoomDetails
 	 * @throws KalturaAPIException
 	 * @beta
 	 */
-	public function allocateConferenceRoomAction($entryId)
+	public function allocateConferenceRoomAction($entryId, $tag = '')
```
yossipapi: Better use null instead of ''

plugins/conference/services/ConferenceService.php
```diff
@@ -111,9 +113,17 @@ protected function findExistingConferenceRoom($entryId)
 		return null;
 	}
 
-	protected function findFreeServerNode()
+	protected function findFreeServerNode($tag, $partner)
 	{
-		$serverNodes = ServerNodePeer::retrieveActiveUnoccupiedServerNodesByType(ConferencePlugin::getCoreValue('serverNodeType',ConferenceServerNodeType::CONFERENCE_SERVER));
+		if (empty($tag))
+		{
+			$tag = $partner->getConfTagName();
```
yossipapi: Better move the if check inside the getConfTagName i.e:
public function getConfTagName()
	{
		return $this->getFromCustomData( self::CUSTOMER_DATA_CONF_TAG_NAME, null ,kConf::get(self::CONFERENCE_SERVER_NODE_TAG_NAME));
	}

configurations/base.ini
```diff
@@ -403,6 +403,8 @@ extend_list[] = BusinessProcessStartNotificationTemplate:mainObjectCode
 extend_list[] = BusinessProcessSignalNotificationTemplate:mainObjectCode
 extend_list[] = BulkUploadResultScheduleEvent:rowData
 
+conference_server_node_tag_name=prod
```
yossipapi: better keep same naming convention as in local.ini rtc_....

alpha/lib/model/ServerNodePeer.php
```diff
@@ -134,13 +134,17 @@ public static function getCacheInvalidationKeys()
...
+	public static function retrieveActiveUnoccupiedServerNodesByType($type, $tag, PropelPDO $con = null)
 	{
 		$c = new Criteria();
 		$c->add(ServerNodePeer::STATUS, ServerNodeStatus::ACTIVE);
 		$c->add(ServerNodePeer::TYPE, $type);
 		$c->add(ServerNodePeer::HEARTBEAT_TIME, time() - ServerNode::SERVER_NODE_TTL_TIME, Criteria::GREATER_EQUAL);
 		$c->addOr(ServerNodePeer::HEARTBEAT_TIME, null);
+		if (!empty($tag))
+		{
+			$c->add(ServerNodePeer::TAGS, $tag, Criteria::LIKE);
```
yossipapi: Using like query on string field which runs on the mysql side is not that good. I think it would be better to fetch all and filter by tag.

alpha/config/schema.xml
```diff
@@ -1716,6 +1716,7 @@
 		<column primaryKey="false" required="false" autoIncrement="false" lazyLoad="false" name="playback_host_name" type="varchar" size="256" />
 		<column primaryKey="false" required="false" autoIncrement="false" lazyLoad="false" name="parent_id" type="varchar" size="256" />
 		<column primaryKey="false" required="false" autoIncrement="false" lazyLoad="false" name="custom_data" type="longvarchar" />
+		<column primaryKey="false" required="false" autoIncrement="false" lazyLoad="false" name="environment" type="varchar" size="256" />
```
yossipapi: better set it to 32 it should be enough

deployment/updates/sql/2018_10_23_alter_server_node_table_add_environment_column.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE server_node	ADD  environment VARCHAR(256) NULL AFTER custom_data;
```
yossipapi: custom_data is usually last so better add before

alpha/data/sql/schema.sql
```diff
@@ -1920,6 +1920,7 @@ CREATE TABLE `server_node`
 	`playback_host_name` VARCHAR(256),
 	`parent_id` VARCHAR(256),
 	`custom_data` TEXT,
+	`environment` VARCHAR(256),
```
yossipapi: Add in deployment/base/sql/01.kaltura_ce_tables.sql as well

api_v3/lib/types/serverNode/base/KalturaServerNode.php
```diff
@@ -107,7 +107,17 @@
 	 * @filter like,mlikeor,mlikeand
 	 */
 	public $parentId;
-	
+
+	/**
+	 * Environment
+	 *
+	 * @var string
+	 * @filter eq
```
yossipapi: Generate filters

yossipapi: Think about adding in as well.

PR #7679, opened 2018-10-14
PLAT-9321: avoid warning

noam-arad:


alpha/apps/kaltura/lib/rules/kRule.php
```diff
@@ -354,6 +354,8 @@ public function hasActionType($actionTypes)
 		if ($actionTypes)
 		{
 			$ruleActions = $this->getActions();
+			if (!$ruleActions)
+				return false;
```
etameran: Please add curly braces

PR #7760, opened 2018-10-29
PLAT-9241 - optimizing flavors on playmanifest

gotlieb:
@MosheMaorKaltura 
@erankor

### DIFF
--- a/alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
+++ b/alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
@@ -988,11 +988,6 @@ class playManifestAction extends kalturaAction
 			}
 			$filteredFlavors[$currentFlavor->getId()] = $currentFlavor;
 		}
-		if (count($filteredFlavors) < 2)
-		{
-			$filteredFlavors[$firstFlavor->getId()] = $firstFlavor;
-		}
-
 		$this->deliveryAttributes->setFlavorAssets($filteredFlavors);
 	}
 
--- a/alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
+++ b/alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
@@ -959,7 +959,7 @@ class playManifestAction extends kalturaAction
 		return $this->deliveryProfile->serve();
 	}
 
-	private function optimizeFlavors()
+	protected function optimizeFlavors()
 	{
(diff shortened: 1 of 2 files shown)


alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
```diff
@@ -1323,4 +1357,23 @@ protected function setPlaylistFlavorAssets($durations, $mediaEntryId)
...
+	 * @param $b
+	 * @return int
+	 */
+	public static function sortFlavorsByFrameSizeAndBitrate($a, $b)
+	{
+		/* @var $a flavorAsset */
+		/* @var $b flavorAsset */
+		if ($a->getHeight() * $a->getWidth() == ($b->getHeight() * $b->getWidth()))
+		{
+			return $a->getBitrate() - $b->getBitrate();
```
yossipapi: In case the values are the same make sure to prioritize based on their original order in the array otherwise, you will get an inconsistent order for different requests.

SAme applies to "return ($a->getHeight() * $a->getWidth()) - ($b->getHeight() * $b->getWidth());"

gotlieb: Original order doesn't matter if the values are the same since in the end of the filtering you will be left with only one flavor to play.

gotlieb: fixed it anyway

alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
```diff
@@ -956,7 +958,39 @@ private function serveVodEntry()
...
+		$firstFlavor = array_shift($flavors);
+		$filteredFlavors = array($firstFlavor->getId() => $firstFlavor );
+		foreach ($flavors as $currentFlavor)
+		{
+			foreach ($filteredFlavors as $elementKey => $flavor)
+			{
+				/* @var $flavor flavorAsset */
+				if (abs(($currentFlavor->getBitrate() - $flavor->getBitrate())) <= ($currentFlavor->getBitrate() * self::FLAVOR_GROUPING_PERCENTAGE_FACTOR)
+					&& ($currentFlavor->getBitrate() >= $flavor->getBitrate())
+					&& (($currentFlavor->getHeight() * $currentFlavor->getWidth()) >= ($flavor->getHeight() * $flavor->getWidth()))
```
MosheMaorKaltura: add getFrameSize on flavorAsset and use it here

PR #7766, opened 2018-10-31
PLAT-9375:if default user does not exsit create it

MosheMaorKaltura:


plugins/vendor/services/ZoomVendorService.php
```diff
@@ -1,215 +1,223 @@
...
+		if (!$zoomCategory && $zoomIntegration->getZoomCategory() && $zoomIntegration->getZoomCategoryId())
+		{
+			$zoomIntegration->unsetCategory();
+			$zoomIntegration->unsetCategoryId();
+		}
+		$zoomIntegration->save();
+		//Validate that default user exist on partner otherwise create it.
+		if(!kuserPeer::getKuserByPartnerAndUid($partnerId, $defaultUserId, true))
+		{
+			$defaultZoomKuser = new kuser();
```
yossipapi: Better use kuserPeer::createKuserForPartner

PR #7774, opened 2018-11-01
No-Plat: Reduce bundle_name length by using md5 function to create the string hash value thus avoiding rename error when base64_encode generated long file name

yossipapi:


yossipapi:
@OrenMe FYI this change will cause the bundler to start recalculating requests since the cache key has changed.

PR #7779, opened 2018-11-04
PLAT-9207: add elastic index rotation script

NadavHarnik:


plugins/search/providers/elastic_search/scripts/aliases/ElasticIndexRotationWorker.php
```diff
@@ -0,0 +1,208 @@
...
+		{
+			$aliases = isset($arr['aliases']) ? $arr['aliases'] : array();
+			foreach ($aliases as $aliasName => $value)
+			{
+				if ($aliasName == $this->indexAlias)
+				{
+					$currentIndexingIndices[] = $indexName;
+				}
+
+				if ($aliasName == $this->searchAlias)
```
MosheMaorKaltura: Put this logic into internal sort function.

PR #7787, opened 2018-11-06
Naos 14.8.0 plat 9376

MosheMaorKaltura:


plugins/vendor/services/ZoomVendorService.php
```diff
@@ -198,6 +206,10 @@ public function recordingCompleteAction()
 		{
 			throw new KalturaAPIException('Zoom Integration data Does Not Exist for current Partner');
 		}
+		if($zoomIntegration->getStatus()==VendorStatus::DISABLED)
```
ZurKaltura: Maybe we should validated that status != ACTIVE since we don't want to process video for deleted vendor either

PR #7803, opened 2018-11-08
Naos 14.8.0 sup 15616

MosheMaorKaltura:


alpha/apps/kaltura/lib/clipconcat/kClipManager.php
```diff
@@ -306,9 +306,16 @@ private function addClipJobs($parentJob , $entryId, &$errDescription, $partnerId
 			$encryptionKey = $originalFlavorAsset->getEncryptionKey();
 		}
 
+		/* @var $singleAttribute kClipAttributes */
 		foreach($operationAttributes as $singleAttribute)
 		{
-			KalturaLog::info("Going To create Flavor for clip: " . print_r($singleAttribute));
+			KalturaLog::info("Going To create Flavor for clip: " . print_r($singleAttribute, true));
```
jessp01: @MosheMaorKaltura - IMHO, 
```php
KalturaLog::info("Going To create Flavor for clip: " . print_r($singleAttribute, true));
```
Should be moved to line 318 since if the condition in line 313 evaluates as true, execution will skip to the next iteration and "Ingnoring clip attribute with non-positive duration" will be logged.

PR #7815, opened 2018-11-12
SUP-16136: fix race condition on changing redirectEntryId

noam-arad:


api_v3/lib/types/entry/KalturaLiveEntry.php
```diff
@@ -208,15 +208,12 @@ public function doFromObject($dbObject, KalturaDetachedResponseProfile $response
 			$this->recordingOptions->fromObject($dbObject->getRecordingOptions());
 		}
 
-		if ($dbObject->getExplicitLive())
+		if ($dbObject->getViewMode() == ViewMode::ALLOW_ALL && in_array($dbObject->getLiveStatus(), array(KalturaEntryServerNodeStatus::PLAYABLE, KalturaEntryServerNodeStatus::BROADCASTING)))
```
MosheMaorKaltura: For better reading please move to small private function - shouldSetRedirectEntryId()
Or just getRedirectEntryId() and it can manage the logic

api_v3/lib/types/entry/KalturaLiveEntry.php
```diff
@@ -208,15 +208,12 @@ public function doFromObject($dbObject, KalturaDetachedResponseProfile $response
...
 		{
-			if ($dbObject->getViewMode() == ViewMode::ALLOW_ALL && in_array($dbObject->getLiveStatus(), array(KalturaEntryServerNodeStatus::PLAYABLE, KalturaEntryServerNodeStatus::BROADCASTING)))
```
yossipapi: Now that the object is no longer saved to the DB the cache won't expire on saving so, in my opinion, we need to reduce cache expiry (note this is an entry object and it is fetched quite a lot).

PR #7821, opened 2018-11-13
PLAT-9241 handle watermark flavors

gotlieb:
@erankor 
@MosheMaorKaltura


alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
```diff
@@ -1404,6 +1404,11 @@ protected function isRedundantFlavor($currentFlavor, $flavor )
 			return false;
 		}
 
+		if (!(strpos($flavor->getTags(), 'watermark') === false))
```
MosheMaorKaltura: watermark should be const somewhere.

PR #7829, opened 2018-11-15
PLAT-9410: Don't allow adding conversion profiles that contain mismatch between their type and the types of the flavor params assigned to them

yossipapi:


MosheMaorKaltura:
reviewed

PR #7832, opened 2018-11-18
PLAT-9228: config should be moved to correct location.

noam-arad:


MosheMaorKaltura:
@noam-arad  isn't this configuration private and should be located at local.ini ?

noam-arad:
@MosheMaorKaltura  it is not private, it should be able to override in local.ini or per partner.

PR #7838, opened 2018-11-19
SUP-15957 - fix missing fileExt on source flavor asset after clipping…

gotlieb:
…/trimming

@MosheMaorKaltura


MosheMaorKaltura:
reviewed

PR #7843, opened 2018-11-19
there are esoteric cases that access_control_id doesn't exist on an e…

noam-arad:
…ntry and the scope will be null


jessp01:
Hi @noam-arad,

Reviewed. Please merge into 14.8.0 as well.

PR #7899, opened 2018-12-05
SUP-16392: order delivery profiles based on user order in configurati…

inbal-ben-david:
…on when format is not given
before change userOrder for all delivery profiles was 0 when format was not given and there were differences in ordering for equal values in php5 and php7


inbal-ben-david:
also PLAT-9501

PR #7907, opened 2018-12-09
Adding check for tag in copied copy cue point

david-winder-kaltura:


plugins/cue_points/base/batch/copyCuePoints/Engine/KLiveClippingCopyCuePointEngine.php
```diff
@@ -14,7 +14,7 @@ protected function shouldCopyCuePoint($cuePoint)
         $cuePointEndTime = $this->getOffsetForTimestamp($cuePoint->calculatedEndTime * 1000, false);
         KalturaLog::debug("Checking times to know if copy is needed for id[". $cuePoint->id ."]: start- [$cuePointStartTime], end- [$cuePointEndTime], calculatedEndTime - " . $cuePoint->calculatedEndTime);
         if ($cuePointStartTime < 0 && is_null($cuePoint->calculatedEndTime))
```
MosheMaorKaltura: Add brackets for single if line (New coding standard)

plugins/cue_points/base/batch/copyCuePoints/Engine/KLiveClippingCopyCuePointEngine.php
```diff
@@ -24,4 +24,14 @@ public function getCuePointFilter($entryId, $status = CuePointStatus::READY)
         $filter = parent::getCuePointFilter($entryId, implode(",",$statuses));
         return $filter;
     }
+    
+    /**
+     * @param KalturaCuePoint $cuePoint
+     * @return boolean
+     */
```
MosheMaorKaltura: Function rename checkShouldCopyCuePointBeforeTimeWindow -> shouldCopyCuePointBeforeTimeWindow
or shouldCopyBasedOnTags and just send it the tags list, also - is it relevant to all cue point types or only to a specific type, if it is relevant to a specific type then I will set this function only on the relevant cue point object.

PR #7917, opened 2018-12-10
 PLAT-8192: use eSearch instead of Sphinx for caption search services

inbal-ben-david:


plugins/search/providers/elastic_search/lib/model/search/kBaseSearch.php
```diff
@@ -11,6 +11,9 @@
 	protected $queryAttributes;
 	protected $mainBoolQuery;
 
+	protected $filterOnlyContext = false;
```
MosheMaorKaltura: Init member in constructor

PR #7940, opened 2018-12-19
 SUP-16113: add broadcasting urls when ks user in member of a kuserEdit entitled group

inbal-ben-david:


MosheMaorKaltura:
reviewed

PR #7941, opened 2018-12-19
Change file_get_content to kFile::getFileContent

david-winder-kaltura:


plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -151,7 +151,7 @@ private function handleExtractMediaFinished(BatchJob $dbBatchJob, kExtractMediaJ
 				return $dbBatchJob;
 			}
 			
-			$rawSyncPointInfo = file_get_contents($rawSyncPointDataPath);
+			$rawSyncPointInfo = kFile::getFileContent($rawSyncPointDataPath);
```
noam-arad: Can return NULL

PR #7966, opened 2018-12-26
PLAT-9543 - add parsing entry_id with regex on thumbnailAction to red…

gotlieb:
…uce dumpiApi to other dc

@MosheMaorKaltura


MosheMaorKaltura:
reviewed

PR #7994, opened 2019-01-07
PSVAMB-5604

hilak:
New feature - add feature allowing bulk updates through the regular XML on multiple entries retrieved according to a filter.
Includes addition to update (and also true update - not override) co-editors and co-publishers of entries via bulk upload XML.
Includes bugfix for issue where XML update to an entry resets the entry owner.

### DIFF
--- a/api_v3/services/BatchService.php
+++ b/api_v3/services/BatchService.php
@@ -181,7 +181,7 @@ class BatchService extends KalturaBatchService
 				//TODO: find some better alternative, find out why the bulk upload result which reports error is
 				// returning objectId "null" for failed entry assets, rather than the entryId to which they pertain.
 				//$data->setNumOfEntries(BulkUploadResultPeer::countWithEntryByBulkUploadId($bulkUploadJobId));
-				$data->setNumOfObjects(BulkUploadResultPeer::countByBulkUploadId($bulkUploadJobId, $data->getBulkUploadObjectType()));
+				$data->setNumOfObjects(BulkUploadResultPeer::countByBulkUploadId($bulkUploadJobId));
 				$data->setNumOfErrorObjects($errorObjects);
 				$bulkUpload->setData($data);
 				$bulkUpload->save();
--- a/api_v3/services/BatchService.php
+++ b/api_v3/services/BatchService.php
@@ -181,7 +181,7 @@ class BatchService extends KalturaBatchService
 				//TODO: find some better alternative, find out why the bulk upload result which reports error is
 				// returning objectId "null" for failed entry assets, rather than the entryId to which they pertain.
 				//$data->setNumOfEntries(BulkUploadResultPeer::countWithEntryByBulkUploadId($bulkUploadJobId));
-				$data->setNumOfObjects(BulkUploadResultPeer::countWithObjectTypeByBulkUploadId($bulkUploadJobId, $data->getBulkUploadObjectType()));
+				$data->setNumOfObjects(BulkUploadResultPeer::countByBulkUploadId($bulkUploadJobId, $data->getBulkUploadObjectType()));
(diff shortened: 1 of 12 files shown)


plugins/bulk_upload/xml/batch/BulkUploadEngineXml.php
```diff
@@ -2196,6 +2213,21 @@ protected function createEntryFromItem(SimpleXMLElement $item, $type = null)
 		return $entry;
 	}
 	
+	protected function parseAndAddToExistingValues ($fieldName, SimpleXMLElement $item, KalturaBaseEntry $existingItem)
+	{
+		$values = $this->implodeChildElements($item->$fieldName);
+		if($existingItem && strpos($values, self::EXISTING_VALUE_DELIMITER) !== false)
```
MosheMaorKaltura: No need to check existingItem , since function signature validates that it has valid value (unless you meant giving it default=null)

plugins/bulk_upload/filter/batch/BulkUploadMediaEntryEngineFilter.php
```diff
@@ -0,0 +1,111 @@
...
+	const ACTION_TAG_XPATH = '/mrss/channel/item/action';
+	
+	protected function listObjects(KalturaFilter $filter, KalturaFilterPager $pager = null)
+	{
+		KBatchBase::impersonate($this->currentPartnerId);
+		
+		$filter->orderBy = "+createdAt";
+		if ($filter instanceof KalturaBaseEntryFilter)
+		{
+			return KBatchBase::$kClient->baseEntry->listAction($filter, $pager);
```
MosheMaorKaltura: if you return here , then the unimpersonate() will not run.

api_v3/services/BatchService.php
```diff
@@ -181,7 +181,7 @@ function updateBulkUploadResultsAction($bulkUploadJobId)
 				//TODO: find some better alternative, find out why the bulk upload result which reports error is
 				// returning objectId "null" for failed entry assets, rather than the entryId to which they pertain.
 				//$data->setNumOfEntries(BulkUploadResultPeer::countWithEntryByBulkUploadId($bulkUploadJobId));
-				$data->setNumOfObjects(BulkUploadResultPeer::countWithObjectTypeByBulkUploadId($bulkUploadJobId, $data->getBulkUploadObjectType()));
+				$data->setNumOfObjects(BulkUploadResultPeer::countByBulkUploadId($bulkUploadJobId, $data->getBulkUploadObjectType()));
```
MosheMaorKaltura: You sending type while function does not need it.

plugins/bulk_upload/filter/batch/BulkUploadCategoryEntryEngineFilter.php
```diff
@@ -86,8 +86,6 @@ public function getObjectTypeTitle()
 	 */
 	protected function listObjects(KalturaFilter $filter, KalturaFilterPager $pager = null) 
 	{
-		KBatchBase::impersonate($this->currentPartnerId);
```
MosheMaorKaltura: If impersonate is not needed remove the unimpersonate also.

plugins/bulk_upload/filter/batch/BulkUploadMediaEntryEngineFilter.php
```diff
@@ -0,0 +1,102 @@
...
+		}
+		else
+		{
+			throw new KalturaBatchException('Unsupported filter: {' . get_class($filter) . '}', KalturaBatchJobAppErrors::BULK_VALIDATION_FAILED);
+		}
+		
+		KBatchBase::unimpersonate();
```
MosheMaorKaltura: remove

PR #8001, opened 2019-01-09
PLAT-9539: Asset dimensions are not set when account is set with encryption + Asset size is incorrectly set with encrypted file size when account is set with encryption

yossipapi:


MosheMaorKaltura:
reviewed

kaltura-hooks:
Hi @yossipapi,

PR #8010, opened 2019-01-13
Naos 14.12.0-PLAT-9480-Blackout events

ZurKaltura:


plugins/schedule/base/lib/model/ScheduleEventPeer.php
```diff
@@ -249,6 +251,30 @@ public static function retrieveEventsByResourceIdsAndDateWindow($resourceIds, $s
...
 		return self::doSelect($c);
 	}
+
+	/**
+	 * @param date $startDate
+	 * @param date $endDate
+	 * @param string|null $scheduleEventIdToIgnore
+	 * @return array <ScheduleEvent>
+	 */
+	public static function retrieveBlackoutEventsByDateWindow($startDate, $endDate, $scheduleEventIdToIgnore = null)
```
MosheMaorKaltura: Since this function and the one above are 95% the same, it would be better to have one function that each over laps it and it is variate to the criteria - 
getEventCriteiaByDateWindow() that will return criteria , and each function will only add it variant of type.

plugins/schedule/base/services/ScheduleEventService.php
```diff
@@ -316,37 +320,58 @@ private static function setRecurringDates($dates, $dbScheduleEvent)
...
+		{
+			if($scheduleEventConflictType != KalturaScheduleEventConflictType::BLACKOUT_CONFLICT)
+			{
+				$events = array_merge($events, ScheduleEventPeer::retrieveEventsByResourceIdsAndDateWindow($resourceIds,
+					$date, $date + $duration, $scheduleEventIdToIgnore));
+			}
+
+			if($scheduleEventConflictType != KalturaScheduleEventConflictType::RESOURCE_CONFLICT)
+			{
+				$events = array_merge($events, ScheduleEventPeer::retrieveBlackoutEventsByDateWindow($date,
```
MosheMaorKaltura: Since this code is almost duplicated , better add function ScheduleEventPeer::getConflicts($dateStart , $dateEnd, $idToIgnore,$conflictType)

PR #8021, opened 2019-01-15
Add partner ID check for storage profile retrieval by ID

hilak:


api_v3/lib/KalturaAssetService.php
```diff
@@ -119,8 +119,8 @@ protected function exportAction ( $assetId , $storageProfileId )
 	    {
 	        throw new KalturaAPIException(KalturaErrors::INVALID_FLAVOR_ASSET_ID, $assetId);
 	    }
-	    
-	    $dbStorageProfile = StorageProfilePeer::retrieveByPK($storageProfileId);	    
+		
+		$dbStorageProfile = StorageProfilePeer::retrieveByIdAndPartnerId($storageProfileId, $dbAsset->getPartnerId());
```
MosheMaorKaltura: Need to make sure that the partner ID of the KS is the asset owner, otherwise we still have an issue

yossipapi:
Why not simply call $this->applyPartnerFilterForClass('storageProfile'); in the init service wouldn't it achive the same.

api_v3/lib/KalturaAssetService.php
yossipapi: Indent.

kaltura-hooks:
Hi @hilak,

PR #8039, opened 2019-01-23
PLAT-9602: extract thumbnail stripe from segment of the video

inbal-ben-david:


alpha/lib/model/entry.php
```diff
@@ -3487,7 +3487,7 @@ protected function getTrackEntryString ($namespace, $customDataColumn, $value)
 	 * @return the thumbnail path.
 	 */
 	public function getLocalThumbFilePath($version , $width , $height , $type , $bgcolor ="ffffff" , $crop_provider=null, $quality = 0,
-		$src_x = 0, $src_y = 0, $src_w = 0, $src_h = 0, $vid_sec = -1, $vid_slice = 0, $vid_slices = -1, $density = 0, $stripProfiles = false, $flavorId = null, $fileName = null)
```
noam-arad: You made a change to the signature of entry.php::getLocalThumbFilePath, this is overridden by LiveEntry.php::getLocalThumbFilePath, you should change the signature of LiveEntry.php as well

PR #8066, opened 2019-01-29
new email notifications

hilak:


kaltura-hooks:
Hi @hilak,

yossipapi:
This is no longer the way we add notification's in the system.
Should be added under deployment/updates/scripts/xml/notifications as a template and in Saas tag add the same file in the same path without the templates.

Also missing release notes.

yossipapi:
Also missing the Also missing the deployment script

PR #8088, opened 2019-02-05
PLAT-9642: when updating answer cue point, the fields should be added to the modified Columns

ravitshalem:


plugins/cue_points/quiz/lib/model/AnswerCuePoint.php
```diff
@@ -105,6 +105,7 @@ public function preUpdate(PropelPDO $con = null)
 			$this->setIsCorrect(in_array($this->getAnswerKey(), $this->getCorrectAnswerKeys()));
 			$this->setCustomDataObj();
 		}
+		parent::preUpdate();
```
yossipapi: You should do return parent:: preUpdate($con);
And remove the call below to preSave what is the reason for calling preSave from the preUpdate?

MosheMaorKaltura:
reviewed

PR #8092, opened 2019-02-06
eSearch - query boolean fields only with true/false

NadavHarnik:
- query on boolean fields in elastic 6.x+ with values other then true/false is deprecated


plugins/search/providers/elastic_search/lib/elasticSearchUtils.php
```diff
@@ -174,6 +176,14 @@ public static function getNumOfFragmentsByConfigKey($highlightConfigKey)
 		return $numOfFragments;
 	}
 
+	public static function getBooleanValue($value)
+	{
+		if (in_array(self::formatSearchTerm($value), self::$elastic_negation_booleans ,true))
+		{
+			return false;
```
MosheMaorKaltura: Much simpler - return !in_array($value ,self::$elastic_negation_booleans);

PR #8133, opened 2019-02-18
Copy poll data cue point regardless of its time

david-winder-kaltura:


plugins/cue_points/base/batch/copyCuePoints/Engine/KLiveClippingCopyCuePointEngine.php
```diff
@@ -10,6 +10,10 @@ protected static function postProcessCuePoints($copiedCuePointIds) {}
 
     protected function shouldCopyCuePoint($cuePoint)
     {
+        if (KCopyCuePointEngine::shouldCopyCuePoint($cuePoint))
```
MosheMaorKaltura: Parent::...

plugins/cue_points/base/batch/copyCuePoints/Engine/KCopyCuePointEngine.php
```diff
@@ -7,13 +7,24 @@ abstract class KCopyCuePointEngine
...
 	protected $data = null;
 	protected $partnerId = null;
 	private $lastCuePointPerType = null;
 
 	abstract public function copyCuePoints();
 
-	protected function shouldCopyCuePoint($cuePoint) {return true;}
+	protected function shouldCopyCuePoint($cuePoint)
+	{
+		if ($cuePoint->cuePointType == self::CUE_POINT_CODE && in_array("poll-data", explode(",", $cuePoint->tags)))
```
MosheMaorKaltura: Please create array - that will hold "poll-data" or any other future extention per cue point type , and just check if it exist ./
Array ( key (cuepoint_Type _ value (filed , expected value))
call it extendedShouldCopyArray

plugins/cue_points/base/batch/copyCuePoints/Engine/KLiveClippingCopyCuePointEngine.php
david-winder-kaltura: This is intentional as the parent (liveToVod engine) allow all cue point regardless of time (to match current behavior of the batch job and I would to exec the logic of the sub-type filtering

plugins/cue_points/base/batch/copyCuePoints/Engine/KCopyCuePointEngine.php
```diff
@@ -7,13 +7,28 @@ abstract class KCopyCuePointEngine
...
 	protected $data = null;
 	protected $partnerId = null;
 	private $lastCuePointPerType = null;
 
 	abstract public function copyCuePoints();
 
-	protected function shouldCopyCuePoint($cuePoint) {return true;}
+	protected function shouldCopyCuePoint($cuePoint)
+	{
+		$extendedShouldCopyTagsArray = array(self::CUE_POINT_CODE => array("poll-data"));
```
MosheMaorKaltura: cuePointType => (property , desired value)

return  isset($extendedShouldCopyTagsArray [$type] && $cuepoint->$extendedShouldCopyTagsArray [$type][0] == $extendedShouldCopyTagsArray [$type][1]

Use const as index.

PR #8134, opened 2019-02-18
use a protocol agnostic URL for embedPlaykitJs

jessp01:


alpha/apps/kaltura/modules/extwidget/templates/previewSuccess.php
```diff
@@ -161,7 +161,7 @@ function getParameterByName(name, url) {
     if (!width) {
         width = 600;
     }
-    var codeUrl = "https://" + data.securedHost + "/p/" + data.partnerId +"/embedPlaykitJs/uiconf_id/"+ data.uiConfId;
+    var codeUrl = "//" + data.securedHost + "/p/" + data.partnerId +"/embedPlaykitJs/uiconf_id/"+ data.uiConfId;
```
MosheMaorKaltura: if you do not specify the protocol , do you need the // prefix?

jessp01: Yes. That makes it protocol agnostic, i.e: if the page was loaded over HTTP, "http://" will be used, if HTTPs, then "https://" will be used.

jessp01: @MosheMaorKaltura - can you merge?

PR #8136, opened 2019-02-19
SUP-17265

hilak:
Eliminate situation where a client requests non-existent flavor param ID from the server and gets empty manifest as result; downgrade thrown errors to warnings.

### DIFF
--- a/alpha/lib/model/DeliveryProfileLive.php
+++ b/alpha/lib/model/DeliveryProfileLive.php
@@ -409,19 +409,19 @@ abstract class DeliveryProfileLive extends DeliveryProfile {
 	
 	protected function filterStreamIdsByBitrate ($streams)
 	{
-		$streamIds = array();
-		
-		$this->sanitizeRequestedFlavorParamsIds($streams);
-		
 		if (!$streams || !count($streams))
 		{
 			return;
 		}
-		elseif (!$this->getDynamicAttributes()->getMaxBitrate() && !$this->getDynamicAttributes()->getMinBitrate())
+		
+		$this->sanitizeRequestedFlavorParamsIds($streams);
+		
+		if (!$this->getDynamicAttributes()->getMaxBitrate() && !$this->getDynamicAttributes()->getMinBitrate())
 		{
 			return;
 		}
 		
+		$streamIds = array();
 		foreach ($streams as $stream)
 		{
 			/* @var $stream kLiveStreamParams */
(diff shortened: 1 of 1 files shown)


alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -442,13 +445,41 @@ protected function filterStreamIdsByBitrate ($streams)
 		
 		if (!count($streamIds))
 		{
-			// If the stream info is available to us, min and/or max bitrate params were passed, and no streams comply with them - throw error
-			KExternalErrors::dieError(KExternalErrors::FLAVOR_NOT_FOUND, "Entry [". $this->getDynamicAttributes()->getEntryId() ."] has no valid streams");
+			// If the stream info is available to us, min and/or max bitrate params were passed, and no streams comply with them - issue warning and cancel restrictions
+			KalturaLog::warning("Entry [". $this->getDynamicAttributes()->getEntryId() ."] has no streams which comply with the min/max bitrate limitations.");
```
MosheMaorKaltura: Use single quote instead of double since not processing is needed.

alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -410,6 +410,9 @@ protected function getRenderer($flavors)
 	protected function filterStreamIdsByBitrate ($streams)
 	{
 		$streamIds = array();
+		
+		$this->sanitizeRequestedFlavorParamsIds($streams);
```
MosheMaorKaltura: Move this call after validating streams exist , instead of validating it twice inside. Also ince move , the validation can be removed from internal protected function

hilak: resolved.

alpha/lib/model/DeliveryProfileLive.php
hilak: resolved.

MosheMaorKaltura:
approved

alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -409,16 +409,19 @@ protected function getRenderer($flavors)
...
 	protected function filterStreamIdsByBitrate ($streams)
 	{
-		$streamIds = array();
 		if (!$streams || !count($streams))
 		{
 			return;
 		}
-		elseif (!$this->getDynamicAttributes()->getMaxBitrate() && !$this->getDynamicAttributes()->getMinBitrate())
+		
+		$this->sanitizeRequestedFlavorParamsIds($streams);
```
yossipapi: Why not do this call after you already have the filtered out the streamIds.
Also currently, the min|max bitrate logic overrides the flavorParamsIds logic (i.e in case both are provided in the playManifest URL).

PR #8137, opened 2019-02-19
PLAT-9656:create new boolean event notification and expand the automatic rule i…

ravitshalem:
…n reach with it

### DIFF
--- a/plugins/reach/lib/kReachManager.php
+++ b/plugins/reach/lib/kReachManager.php
@@ -8,10 +8,11 @@ class kReachManager implements kObjectChangedEventConsumer, kObjectCreatedEventC
 	/**
 	 * @var array<booleanNotificationTemplate>
 	 */
-	protected $booleanNotificationTemplates;
+	protected static $booleanNotificationTemplatesFulfilled;
+	protected static $booleanNotificationTemplatesFromReachProfiles;
+	protected static $reachProfilesFilteredWithoutBooleanEventNotifications;
 
-	protected $gotBooleanNotificationTemplates;
-	protected $gotEmptyBooleanNotificationTemplates;
+	static $isInit = false;
 
 	protected function getObjectType($eventObjectClassName)
 	{
@@ -52,9 +53,9 @@ class kReachManager implements kObjectChangedEventConsumer, kObjectCreatedEventC
 		$partnerId = $scope->getPartnerId();
 		$object = $scope->getObject();
 		$entryId = $object->getEntryId();
-		if (count($this->booleanNotificationTemplates) == 0)
+		if (count(self::$booleanNotificationTemplatesFulfilled) == 0)
 			return false;
-		foreach ($this->booleanNotificationTemplates as $booleanNotificationTemplate)
+		foreach (self::$booleanNotificationTemplatesFulfilled as $booleanNotificationTemplate)
 		{
 			$profileId = $booleanNotificationTemplate[0];
 			$booleanNotificationTemaplateObjects = $booleanNotificationTemplate[1];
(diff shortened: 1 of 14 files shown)


plugins/reach/lib/kReachManager.php
```diff
@@ -3,8 +3,130 @@
...
+	 * @see kGenericEventConsumer::shouldConsumeEvent()
+	 */
+	public function shouldConsumeEvent(KalturaEvent $event)
+	{
+		$this->booleanNotificationTemplates = array();
+		$fulfilled = 0;
+
+		$scope = $event->getScope();
+		$partnerId = $scope->getPartnerId();
+		$eventType = kEventNotificationFlowManager::getEventType($event);
```
MosheMaorKaltura: Check here if partner ad reach enabled if not return false;

plugins/reach/lib/kReachManager.php
```diff
@@ -3,8 +3,130 @@
...
+		}
+		return true;
+	}
+
+	/* (non-PHPdoc)
+	 * @see kGenericEventConsumer::shouldConsumeEvent()
+	 */
+	public function shouldConsumeEvent(KalturaEvent $event)
+	{
+		$this->booleanNotificationTemplates = array();
```
MosheMaorKaltura: Run this code only once in init method that will generate static associative array that will contain the following info - key (object type , event type) -> value( profile id , catelogItem , boolean event notification ID).
This will running the loop multiple times.
Eventully , you just need to check if the object type and and event type are in the array

plugins/reach/lib/kReachManager.php
```diff
@@ -28,6 +150,33 @@ public function shouldConsumeCreatedEvent(BaseObject $object)
 		return false;
 	}
 
+	public function settingFlagsBooleanEventNotificationIds(KalturaEvent $event)
```
MosheMaorKaltura: Once you have the array set , this code is not needed

plugins/reach/lib/kReachManager.php
```diff
@@ -308,13 +474,41 @@ protected static function getTaskPuserId(entry $entry)
 		return $puserId;
 	}
 
+	protected function filterOnlyReachProfilesWithoutBooleanEventNotification($reachProfiles)
+	{
```
MosheMaorKaltura: Instead just do a fast scan pre-prepared array

plugins/reach/lib/kReachManager.php
```diff
@@ -3,8 +3,166 @@
...
+				{
+					$booleanEventNotificationIdArray = explode(',', $booleanNotificationTemplatesFromReachProfile[1]->getbooleanEventNotificationIds());
+					$boolEventNotificationObjectList = EventNotificationTemplatePeer::retrieveByEventTypeObjectTypeAndPKS($eventType, $objectType, $partnerId, $booleanEventNotificationIdArray);
+					foreach ($boolEventNotificationObjectList as $boolEventNotificationObject)
+					{
+						$fulfilled = $boolEventNotificationObject->fulfilled($scope);
+						if ($fulfilled)
+						{
+							self::$booleanNotificationTemplatesFulfilled[] = array($booleanNotificationTemplatesFromReachProfile[0], $booleanNotificationTemplatesFromReachProfile[1]);
+							$fulfilled = 0;
```
MosheMaorKaltura: remove, not needed

MosheMaorKaltura: Instead of 0 &1 , use CONST const PROFILE_ID ,

plugins/reach/lib/kReachManager.php
```diff
@@ -3,8 +3,166 @@
...
+					if ($profileWasAdded)
+					{
+						break;
+					}
+				}
+			}
+		}
+	}
+
+	/* (non-PHPdoc)
```
MosheMaorKaltura: Add notes - side effect : while checking the list, buildin.....

plugins/reach/lib/kReachManager.php
```diff
@@ -3,8 +3,166 @@
...
+			default:
+				return null;
+		}
+	}
+	/* (non-PHPdoc)
+	 * @see kGenericEventConsumer::consumeEvent()
+	 */
+	public function consumeEvent(KalturaEvent $event)
+	{
+		$scope = $event->getScope();
```
MosheMaorKaltura: 1. This function must return true.
2. Move its content to reachCosnumeEvent and you can use its return value,

plugins/reach/lib/kReachManager.php
```diff
@@ -54,6 +212,13 @@ public function shouldConsumeChangedEvent(BaseObject $object, array $modifiedCol
 		if ($object instanceof categoryEntry && $object->getStatus() == CategoryEntryStatus::ACTIVE)
 			return true;
 
+		if ($object instanceof entry && $object->getType() == entryType::MEDIA_CLIP && $object->isModified() )
```
MosheMaorKaltura: This rule must be validated before.

plugins/reach/lib/kReachManager.php
```diff
@@ -54,6 +212,13 @@ public function shouldConsumeChangedEvent(BaseObject $object, array $modifiedCol
...
+		{
+			$event = new kObjectChangedEvent($object,$modifiedColumns);
```
MosheMaorKaltura: If there are new rules, must act only according to the profile rules decision, else to support backward, check the legacy code.,

plugins/reach/lib/kReachManager.php
```diff
@@ -313,31 +485,34 @@ private function checkAutomaticRules($object, $checkEmptyRulesOnly = false)
...
-			$catalogItemIdsToAdd = array_unique(array_diff($allowedCatalogItemIds, $existingCatalogItemIds));
-			
-			foreach ($catalogItemIdsToAdd as $catalogItemIdToAdd)
-			{
-				//Pass the object Id as the context of the task
-				self::addEntryVendorTaskByObjectIds($entryId, $catalogItemIdToAdd, $profile->getId(), $this->getContextByObjectType($object));
+				/* @var $profile ReachProfile */
+				$fullFieldCatalogItemIds = $profile->fulfillsRules($scope, $checkEmptyRulesOnly);
+				if (!count($fullFieldCatalogItemIds))
+					continue;
```
MosheMaorKaltura: {

}

plugins/reach/lib/kReachManager.php
```diff
@@ -313,31 +485,34 @@ private function checkAutomaticRules($object, $checkEmptyRulesOnly = false)
...
+
+				$allowedCatalogItemIds = PartnerCatalogItemPeer::retrieveActiveCatalogItemIds($fullFieldCatalogItemIds, $object->getPartnerId());
```
MosheMaorKaltura: Consolidate this code since it repeats twice.

PR #8138, opened 2019-02-19
PLAT-9643 - As admin console user I would like to be able to view old…

gotlieb:
… versions of conf maps.

@MosheMaorKaltura

### DIFF
--- a/plugins/conf_maps/lib/api/filters/KalturaConfMapsFilter.php
+++ b/plugins/conf_maps/lib/api/filters/KalturaConfMapsFilter.php
@@ -61,12 +61,13 @@ class KalturaConfMapsFilter extends KalturaConfMapsBaseFilter
 		$hostPatern = str_replace('*','#', $this->relatedHostEqual);
 		/*  @var kRemoteMemCacheConf $remoteCache  */
 		$remoteCache = kCacheConfFactory::getInstance(kCacheConfFactory::REMOTE_MEM_CACHE);
+		$map = null;
 		if (!is_null($this->versionEqual))
 		{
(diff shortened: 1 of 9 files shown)


plugins/conf_maps/lib/model/ConfMapsPeer.php
```diff
@@ -27,17 +27,28 @@ public static function setDefaultCriteriaFilter()
...
 		$c->add(self::MAP_NAME ,$mapName );
 		if($hostNameRegex)
 		{
 			$c->add(self::HOST_NAME, $hostNameRegex);
 		}
+		if (!is_null($version))
+		{
+			$c->add(self::VERSION, $version);
+		}
 		$c->addDescendingOrderByColumn(self::VERSION);
```
MosheMaorKaltura: This should be added only if version is not set

plugins/conf_maps/lib/api/filters/KalturaConfMapsFilter.php
```diff
@@ -61,7 +61,21 @@ public function getMap()
 		$hostPatern = str_replace('*','#', $this->relatedHostEqual);
 		/*  @var kRemoteMemCacheConf $remoteCache  */
 		$remoteCache = kCacheConfFactory::getInstance(kCacheConfFactory::REMOTE_MEM_CACHE);
-		$map = $remoteCache->loadByHostName($this->nameEqual, $hostPatern);
+		if (!is_null($this->versionEqual))
+		{
+			$map = ConfMapsPeer::getMapByVersion($this->nameEqual, $hostPatern, $this->versionEqual);
+			if ($map)
```
MosheMaorKaltura: This entire block is redundant

MosheMaorKaltura:
reviewed

PR #8163, opened 2019-02-27
Was introduced by this commit b5d587e1eeb6e10fe972acf28c4cbb7471a03074

jessp01:
…4 but never added to the DB deployment file


jessp01:
@MosheMaorKaltura - FYI.

PR #8166, opened 2019-03-03
REACH2-493:adding boolean event notification condition to Reach

ravitshalem:


ravitshalem:
@MosheMaorKaltura @yossipapi

PR #8167, opened 2019-03-03
Naos 14.15.0 plat 8580 Adding managers to group user

ZurKaltura:


api_v3/lib/types/KalturaGroupUser.php
```diff
@@ -91,4 +103,31 @@ public function getFilterDocs()
 	{
 		return array();	
 	}
+
+	public function validateForUpdate($sourceObject, $propertiesToSkip = array())
+	{
+		parent::validateForUpdate($sourceObject, $propertiesToSkip);
+		if(!kCurrentContext::$is_admin_session)
+		{
+			if($sourceObject->getUserRole() != $this->userRole)
```
MosheMaorKaltura: User with group-user update permission can change or update any group as before this feature.

PR #8175, opened 2019-03-05
PLAT-9730 in regards to SUP-16999 in cas of importing during clip_con…

gotlieb:
…cat flow ignore virus scan since it blocks other consumers

@MosheMaorKaltura

### DIFF
--- a/plugins/virus_scan/lib/batch/kVirusScanFlowManager.php
+++ b/plugins/virus_scan/lib/batch/kVirusScanFlowManager.php
@@ -101,8 +101,7 @@ class kVirusScanFlowManager implements kBatchJobStatusEventConsumer, kObjectAdde
 			return false;
 		if($object instanceof asset)
 		{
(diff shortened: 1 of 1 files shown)


plugins/virus_scan/lib/batch/kVirusScanFlowManager.php
```diff
@@ -100,7 +100,15 @@ public function shouldConsumeAddedEvent(BaseObject $object)
 		if (!class_exists('kCurrentContext') || !kCurrentContext::isApiV3Context())
 			return false;
 		if($object instanceof asset)
+		{
+			$entryId = $object->getEntryId();
```
yossipapi: Asset has a getentry method better use it.

MosheMaorKaltura: As Yossi suggested , just use $object->getentry()

PR #8186, opened 2019-03-08
Update Live Params and remove ingest tag from Cloud Transcode flavors

DBezemer:
Having the ingest tag on the flavor and incorrect permissions causes 720p flavors to be unavailable for Cloud Transcode


DBezemer:
@MosheMaorKaltura please review and merge, this currently breaks Cloud Transcode in OnPrem

yossipapi:
@DBezemer I am merging this but please update release notes in a separate PR so it will be properly documented.

PR #8188, opened 2019-03-08
Support deleting files when using SFTP with key based authentication

DBezemer:
Support deleting files when using SFTP with key based authentication

### DIFF
--- a/batch/batches/Storage/Engine/KFileTransferExportEngine.php
+++ b/batch/batches/Storage/Engine/KFileTransferExportEngine.php
@@ -109,8 +109,24 @@ class KFileTransferExportEngine extends KExportEngine
         $engineOptions['passiveMode'] = $this->data->ftpPassiveMode;
         $engine = kFileTransferMgr::getInstance($this->protocol, $engineOptions);
         
-        try{
-            $engine->login($this->data->serverUrl, $this->data->serverUsername, $this->data->serverPassword);
+        try {
+            $keyPairLogin = false;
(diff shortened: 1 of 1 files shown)


DBezemer:
@MosheMaorKaltura please review, this currently blocks an issue for Deutsche Telekom

batch/batches/Storage/Engine/KFileTransferExportEngine.php
```diff
@@ -109,8 +109,24 @@ function delete()
         $engineOptions['passiveMode'] = $this->data->ftpPassiveMode;
         $engine = kFileTransferMgr::getInstance($this->protocol, $engineOptions);
         
-        try{
-            $engine->login($this->data->serverUrl, $this->data->serverUsername, $this->data->serverPassword);
+        try {
```
yossipapi: @DBezemer this code is exactly the same as in https://github.com/kaltura/server/blob/Naos-14.16.0/batch/batches/Storage/Engine/KFileTransferExportEngine.php#L47 better move it to a method and call it from both locations. Other than that seems ok.

DBezemer: @yossipapi I'm not going to improve on the original code. This pull request is purely to support a case that was omitted from the original code (SFTP with key based auth does not currently support deleting files even tho it does attempt it).

The getTempFileWithContent seems to have been removed in between versions, as it used to be part of the kFile class. I will amend the request with the same logic as in the export() function. If you wish to rewrite the method to a function instead of having this duplication feel free to close this pull and do so yourself.

PR #8215, opened 2019-03-14
PLAT-9645:validating the property entryId is not null before updating …

ravitshalem:
…the cue point


MosheMaorKaltura:
reviewed

PR #8232, opened 2019-03-20
PLAT-9421 - save cdnWhiteList results in memory in order to avoid mul…

gotlieb:
…tiple invokes during list calls

@MosheMaorKaltura 
@ZurKaltura


ZurKaltura:
reviewed

PR #8237, opened 2019-03-21
Naos-14.17.0-SUP-17688 Playlist update

ZurKaltura:


ZurKaltura:
Fix playlist update to work then when the partner have encryption files feature toggled on

PR #8267, opened 2019-04-01
making userEntry bulk delete work with bulk batch job

ravitshalem:


api_v3/lib/types/filters/KalturaUserEntryFilter.php
```diff
@@ -143,6 +150,11 @@ public function doFromObject($srcObj, KalturaDetachedResponseProfile $responsePr
 	 */
 	protected function fixFilterUserId()
 	{
+		if (kCurrentContext::$ks_partner_id == Partner::BATCH_PARTNER_ID)
+		{
+			kCurrentContext::$partner_id = $this->partnerId;;
```
MosheMaorKaltura: ;;

plugins/bulk_upload/filter/lib/api/KalturaBulkServiceFilterDataOnlyFilter.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+/**
+ * Represents the Bulk service input for filter bulk upload
+ * @package plugins.bulkUploadFilter
+ * @subpackage api.objects
+ */
+class KalturaBulkServiceFilterDataOnlyFilter extends KalturaBulkServiceData
```
MosheMaorKaltura: rename to KalturaBulkUploadFilterBase

yossipapi: Rename to KalturaBulkServiceFilterDataBase

plugins/bulk_upload/filter/batch/BulkUploadUserEntryEngineFilter.php
```diff
@@ -0,0 +1,81 @@
...
+
+		else
+		{
+			throw new KalturaBatchException("Unsupported filter: {get_class($filter)}", KalturaBatchJobAppErrors::BULK_VALIDATION_FAILED);
+		}
+	}
+
+	protected function createObjectFromResultAndJobData (KalturaBulkUploadResult $bulkUploadResult)
+	{
+
```
yossipapi: IS this function entry because the task itself does delete and there is no object to return?
If so add a comment explaining this and return null as a response.

PR #8281, opened 2019-04-03
 PLAT-9707: add idGT and createdAtGTE to partner filter

inbal-ben-david:


MosheMaorKaltura:
Need to make sure that created_at is added as index to the table else it will take forever.

 select count(*) from partner where created_at >'2019-01-01';
....
1 row in set (1 min 12.34 sec)

PR #8353, opened 2019-04-21
PLAT-9831: When returning KalturaNullableBoolean field value make sure to cast it to int value

yossipapi:


MosheMaorKaltura:
To avoid cases of inherting KalturaNullableBoolean - user typefo or limit the class by adding final

PR #8368, opened 2019-04-23
WEBC-1428: for vod entries choose specific domain prefix since it's n…

noam-arad:
…ot flavor related, will give same KES for all flavors


kaltura-hooks:
Hi @noam-arad,

PR #8372, opened 2019-04-25
PLAT-9836: We need to apply partner id on context when calling cue point count as we do in list

yossipapi:


plugins/cue_points/base/lib/api/filters/KalturaCuePointFilter.php
```diff
@@ -198,5 +189,25 @@ private function applyPartnerOnCurrentContext($entryIds)
 		{
 			kCurrentContext::$partner_id = $entry->getPartnerId();
 		}
+		else
+		{
+			KalturaLog::debug("Entry id not filtered, If partner id not correctly defined wront results may return");
```
MosheMaorKaltura: Typo

plugins/cue_points/base/lib/api/filters/KalturaCuePointFilter.php
```diff
@@ -198,5 +189,25 @@ private function applyPartnerOnCurrentContext($entryIds)
 		{
 			kCurrentContext::$partner_id = $entry->getPartnerId();
 		}
+		else
```
MosheMaorKaltura: WHat is no entryIds in list?

PR #8413, opened 2019-05-13
REACH2-581:enabling different rules boolean and regular on the same reach profile

ravitshalem:


plugins/reach/lib/kReachManager.php
```diff
@@ -100,14 +99,15 @@ protected static function initReachProfileForPartner($partnerId)
 					if ( $condition->getType()== ConditionType::BOOLEAN && $condition->getbooleanEventNotificationIds() && $condition->getbooleanEventNotificationIds() != "N/A")
 					{
 						self::$booleanNotificationTemplatesFromReachProfiles[] = array($profile->getId(), $condition, $rule->getActions());
-						$profileWasAdded++;
+					}
+					else if ( $condition->getType()== ConditionType::BOOLEAN &&
+						(!$condition->getbooleanEventNotificationIds() ||
+							($condition->getbooleanEventNotificationIds() && $condition->getbooleanEventNotificationIds() == "N/A" )))
```
MosheMaorKaltura: N/A as const
Use === or strcmp instead of ==

plugins/reach/lib/model/ReachProfile.php
```diff
@@ -440,15 +440,16 @@ public function syncCreditPercentageUsage()
...
 		$fullFilledCatalogItemIds = array();
 		if(!is_array($this->getRulesArray()) || !count($this->getRulesArray()))
 			return $fullFilledCatalogItemIds;
 		
 		$context = new kContextDataResult();
 		foreach ($this->getRulesArray() as $rule)
 		{
+			$gotBooleanCondition = false;
+			$gotNonBooleanCondition = false;
+
```
MosheMaorKaltura: No need to check count on empty array.

PR #8414, opened 2019-05-13
Add live fields for all event if the template is live entry

david-winder-kaltura:


david-winder-kaltura:
This change is due to PLAT-9865

MosheMaorKaltura:
reviewed

PR #8415, opened 2019-05-13
PLAT-9862:mediaService approveReplace action for images will return t…

ravitshalem:
…he entry as is,without approving


MosheMaorKaltura:
reviewed

PR #8417, opened 2019-05-13
Update XmlClientGenerator.php

hilak:


hilak:
@MosheMaorKaltura I've removed the argument in case the pluginName is empty, and also the creation of the 'prefix' var which appear completely unused. Please review and merge if acceptable

PR #8432, opened 2019-05-16
PLAT-9870: Add configuration for blocking specific partner from listing category Entry without providing proper filter

yossipapi:


api_v3/lib/types/filters/KalturaCategoryEntryFilter.php
```diff
@@ -20,11 +20,16 @@ protected function getCoreFilter()
 	 */
 	public function getListResponse(KalturaFilterPager $pager, KalturaDetachedResponseProfile $responseProfile = null)
 	{
+		$blockOnEmptyFilterPartners = kConf::getMap("partner_ids_require_category_entry_filter");
```
MosheMaorKaltura: Add this map to alpha/config/kConfMapNames.php and use const

PR #8472, opened 2019-05-27
feat(FEC-9147): pass KS to server preview template

shalommeoded:
Added a way to pass KS on the preview template.

Checked with QA servers:
`https://qa-apache-php7.dev.kaltura.com/index.php/extwidget/preview/partner_id/6502/uiconf_id/15222758/entry_id/0_4x8h8feu/embed/dynamic?flashvars[ks]=`

### DIFF
--- a/alpha/apps/kaltura/modules/extwidget/templates/previewSuccess.php
+++ b/alpha/apps/kaltura/modules/extwidget/templates/previewSuccess.php
@@ -154,6 +154,11 @@ if (isPlaykit === '1') {
         }
         catch(ee){}
     }
+    if (typeof data.flashVars !== 'undefined') {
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/modules/extwidget/templates/previewSuccess.php
```diff
@@ -153,6 +153,11 @@ function getParameterByName(name, url) {
             playerConfig = mergeDeep(playerConfig,parsedConfig);
         }
         catch(ee){}
+    }
+    if (typeof data.flashVars !== 'undefined') {
```
OrenMe: don't repeat the check, do it once and save the KS and then use it in the dynamic or iframe embeds

PR #8475, opened 2019-05-28
Orion-15.0.0-SipIntergration - add email support, add explicit live, …

gotlieb:
…add locations support.

@MosheMaorKaltura


plugins/sip/lib/model/providers/pexip/kPexipUtils.php
```diff
@@ -103,48 +105,55 @@ public static function retrieveAndValidateEntryForSipCall($queryParams, $pexipCo
 		if (!$dbLiveEntry)
 		{
 			$msg = "Entry was not found for sip token $sipToken";
+			self::sendSipEmailNotification($dbLiveEntry->getPartnerId(), $dbLiveEntry->getCreatorPuserId(), $msg, $dbLiveEntry->getId());
 			KalturaLog::err($msg);
```
MosheMaorKaltura: In all the following cases , err with backtrace is not required , It enough to  print Notice or Warning.
Applies to the entire page here

PR #8514, opened 2019-06-05
Update 2019_05_19_alter_table_sphinx_log.sql

DBezemer:
Not specifying the DB name makes this fail as update.php only runs agains `kaltura` database.


DBezemer:
See #8515

PR #8515, opened 2019-06-05
Fix update.php

jessp01:


jessp01:
Two other thoughts:
- Would be far better for the `version_management` table to include a `status` column so that we can insert records for failing SQL files as well and set the status accordingly. This would make reviewing the upgrade process far easier
- Instead of using `echo` for outputting various messages, we should use a proper logging mechanism, as we do elsewhere

PR #8524, opened 2019-06-07
Entry creation date update script.

hilak:


alpha/scripts/utils/setEntryCreationDate.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+if($argc < 3)
+{
+	echo "Arguments missing.\n\n";
+	echo "Usage: php " . __FILE__ . " {mapping} {isDateString} <realrun / dryrun> \n";
+	exit;
+}
+$mapping = $argv[1];
+$isDateString = ($argv[2] == "true");
```
MosheMaorKaltura: use === otherwise 0 can also work , 
try -> if (0 == 'true') print 'boo';

alpha/scripts/utils/setEntryCreationDate.php
```diff
@@ -0,0 +1,40 @@
...
+	$entry = entryPeer::retrieveByPK($entryId);
+	if(!$entry)
+	{
+		echo "Entry id [$entryId] not found\n";
+		continue;
+	}
+	
+	$entry->setCreatedAt($createdAt);
+	$entry->setAvailableFrom($createdAt);
+	$entry->save();
```
MosheMaorKaltura: Since we are loosing created at value , you must add new field to entry - orinigalCreatedAt and save the current value.
This will help us in future cases where we would like to understand when the record was created.

alpha/scripts/utils/setEntryCreationDate.php
```diff
@@ -0,0 +1,40 @@
...
+
+if($argc < 3)
+{
+	echo "Arguments missing.\n\n";
+	echo "Usage: php " . __FILE__ . " {mapping} {isDateString} <realrun / dryrun> \n";
+	exit;
+}
+$mapping = $argv[1];
+$isDateString = ($argv[2] == "true");
+$dryRun = ($argv[3] == "dryrun");
```
MosheMaorKaltura: Same here ===

hilak:
@MosheMaorKaltura  I have addressed your comments. Please re-review.

MosheMaorKaltura:
Reviewed

PR #8538, opened 2019-06-12
SUP-18090: When user is deleted, delete all dedicated appTokens

MosheMaorKaltura:


alpha/lib/model/AppTokenPeer.php
```diff
@@ -44,4 +44,11 @@ public static function getCacheInvalidationKeys()
 	{
 		return array(array("appToken:id=%s", self::ID));		
 	}
+
+	public static function retrieveByKuserId($kuserId)
+	{
+		$c = new Criteria();
+		$c->add(self::KUSER_ID, $kuserId);
+		return self::doSelect($c);
```
yossipapi: Add partner Id to the criteria since it is not applied by default when coming from the user service.

PR #8541, opened 2019-06-13
 PLAT-9841:block ability to use loginByLoginId with partner configuration

ravitshalem:


api_v3/lib/KalturaBaseUserService.php
```diff
@@ -214,6 +215,9 @@ protected function loginImpl($puserId, $loginEmail, $password, $partnerId = null
 			else if ($code == kUserException::MISSING_OTP) {
 					throw new KalturaAPIException(KalturaErrors::MISSING_OTP);
 			}
+			else if ($code == kUserException::DIRECT_LOGIN_BLOCKED) {
```
MosheMaorKaltura: String comparison is always === or strcmp

PR #8564, opened 2019-06-19
Same problem as with https://github.com/kaltura/server/pull/8515/commits/3718727ce18bfb519fa9cb0b103d3da5d1dc1a3a

jessp01:


jessp01:
@yossipapi, @MosheMaorKaltura - this is critical for CE/OP. Without this fix, the alteration will not take place for reasons previously explained. Please merge.

PR #8622, opened 2019-07-08
partner->getSecrets() should accept the optional $otp and pass it on to userLoginByEmail()

jessp01:
At the moment, attempting to call this function when OTP was enabled on the partner will fail.


jessp01:
@yossipapi, @MosheMaorKaltura - IMHO, this warrants a hotfix. 
This also prevents the developer.kaltura.com login flow from working when OTP has been enabled on the partner.

PR #8648, opened 2019-07-16
SUP-17451 - handle user list in case of more than 500 in results of q…

gotlieb:
…uery due to duplicates emails

@MosheMaorKaltura


plugins/metadata/lib/api/filters/KalturaMetadataFilter.php
```diff
@@ -98,7 +98,8 @@ private function validateObjectIdFiltered()
 			$kusers = !empty($objectIds) ? kuserPeer::getKuserByPartnerAndUids(kCurrentContext::getCurrentPartnerId(), $objectIds) : array();
 			$objectIds = array();
 			foreach($kusers as $kuser)
-				$objectIds[] = $kuser->getId();
+				$objectIds[$kuser->getPuserId()] = $kuser->getId();
+			$objectIds = array_values($objectIds);
```
MosheMaorKaltura: This action is not needed

PR #8681, opened 2019-07-29
awsIntegration-PLAT-10038: don't move file on moveFromFile when src

inbal-ben-david:
and target are both shared + small fixes


infra/storage/shared_file_system_managers/kSharedFileSystemMgr.php
```diff
@@ -636,4 +651,9 @@ public function copyDir($src, $dest, $deleteSrc)
 		return $this->doCopyDir($src, $dest, $deleteSrc);
 	}
 
+	public static function getCopySharedToSharedAllowed()
+	{
+		return self::$kCopySharedToShared;
```
yossipapi: Better call getInstance form here and than ave the static member value based on what it returned.

PR #8682, opened 2019-07-31
PLAT-9949:when permissions query returned empty, throw Exception befo…

ravitshalem:
…re storing the empty value in cache


MosheMaorKaltura:
reviewed

PR #8719, opened 2019-08-22
PLAT-10113: add script that checks for duplicated users created since…

inbal-ben-david:
… last run and merges them


alpha/scripts/utils/mergeNewlyCreatedDuplicatedUsers.php
```diff
@@ -165,3 +177,34 @@ function getLastId($currentTime, $startId)
...
+
+	if (!$toArray || count($toArray) < 1 || strlen($toArray[0]) == 0)
+		return true;
+
+	foreach ($toArray as $to)
+		$mailer->AddAddress($to);
+
+	$mailer->Subject = $subject;
+	$mailer->Body = $body;
+	$mailer->Sender = 'notifications@kaltura.com';
```
MosheMaorKaltura: Take the sender from kConf::get('batch_notification_sender_email')

alpha/scripts/utils/mergeNewlyCreatedDuplicatedUsers.php
```diff
@@ -7,25 +7,37 @@
...
+try
 {
-	KalturaLog::debug('Could not lock file. Exiting.');
-	return;
-}
-$lastRunFilePath = $argv[1];
+	$fp = fopen(__DIR__ . '/mergeNewlyCreatedDuplicatedUsers.php', "r+");
+	if (!flock($fp, LOCK_EX|LOCK_NB))
+	{
+		KalturaLog::debug('Could not lock file. Exiting.');
```
MosheMaorKaltura: What if we cannot run the script - how can we tell about it ? throw exception here.

PR #8741, opened 2019-09-02
PLAT-10100: The trim() function is not, by default, unicode safe.

jessp01:
Thus, on certain machines, if such trailing Unicode characters exist, a complete failure will occur as described in https://kaltura.atlassian.net/browse/PLAT-10100

This change fixes it. Verified on qa-apache-php7 where, before the
change, I was able to reproduce it.

### DIFF
--- a/plugins/bulk_upload/csv/batch/BulkUploadEngineCsv.php
+++ b/plugins/bulk_upload/csv/batch/BulkUploadEngineCsv.php
@@ -237,7 +237,7 @@ abstract class BulkUploadEngineCsv extends KBulkUploadEngine
 	 */
 	protected function trimArray(&$item)
 	{
-		$item = trim($item);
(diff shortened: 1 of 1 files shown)


jessp01:
For additional context, the actual failure occurs here:
https://github.com/kaltura/server/blob/Orion-15.6.0/plugins/bulk_upload/csv/batch/BulkUploadUserEngineCsv.php#L48

Following the trim() done here:
https://github.com/kaltura/server/blob/Orion-15.6.0/plugins/bulk_upload/csv/batch/BulkUploadUserEngineCsv.php#L33

In the batch/bulkupload-0-2019-09-02.log, for the input file attached to the ticket, you will see:
> [BATCH] [BulkUploadUserEngineCsv->createUploadResult] INFO: Value userId is empty
> [BATCH] [BulkUploadUserEngineCsv->createUploadResult] INFO: Value email is empty

PR #8772, opened 2019-09-16
PLAT-10120:adding option to remove synonyms in esearch

ravitshalem:


admin_console/forms/PartnerConfiguration.php
```diff
@@ -184,6 +184,10 @@ public function init()
 
 		$this->getElement('editDeliveryProfiles')->setAttrib('onClick', 'addDeliveryProfile()');
 
+		$this->addElement('checkbox', 'ignore_synonym_esearch', array(
+			'label'	  => 'Ignore Synonyms in eSearch',
```
MosheMaorKaltura: 1. Synonyms should be lower case.
2. Fix the space after the label

plugins/search/providers/elastic_search/lib/model/search/ESearchQueryAttributes.php
```diff
@@ -85,6 +85,11 @@ class ESearchQueryAttributes
 	 */
 	protected $searchHistoryMustNotContext;
 
+	/**
+	 * @var bool
+	 */
+	protected $ignoreSynonymsOnPartner;
```
MosheMaorKaltura: Until now it was called Synonym (single), keep the same name

PR #8784, opened 2019-09-24
PLAT-10200:Add ability search history indexing per partner

MosheMaorKaltura:


ZurKaltura:
Hi,

I think we should consider to also return an exception when we try to retrieve search history for a partner that it search history disabled so it will be clear for the partner why it doesn't get results.

This also affects the KMS since they query for last searches they need to change their to logic to see if the partner have history disabled otherwise we will get tickets from people who don't have the right terms when they look at their history.

MosheMaorKaltura:
Since this feature only disables new indexing I see no point in blocking history search that may contain valid content.

PR #8806, opened 2019-10-13
PLAT-10216:update last login when login with sso

ravitshalem:


alpha/lib/model/UserLoginDataPeer.php
```diff
@@ -557,6 +557,12 @@ private static function userLogin(UserLoginData $loginData = null, $password, $p
 			}
 		}
 
+		self::setLastLoginFields($loginData, $kuser);
+		return $kuser;
```
MosheMaorKaltura: In case of user that was just created, its id will be generated only after calling save.
Therefor ** setLastLoginFields** must return the kuser after save.

plugins/sso/services/SsoService.php
```diff
@@ -135,14 +135,27 @@ public function loginAction($userId = null, $applicationType , $partnerId = null
...
+			self::setLastLogin($partnerId, $userId);
 		}
 		$sso = new KalturaSso();
 		$sso->fromObject($dbSso, $this->getResponseProfile());
 		return $sso->redirectUrl;
 	}
 
+	protected static function setLastLogin($partnerId, $userId)
+	{
+		kuserPeer::setUseCriteriaFilter(false);	//action is without ks, need to prevent the addition of partnerId 0 into the criteria
```
MosheMaorKaltura: Need to validate partnerId is not 0

PR #8808, opened 2019-10-13
Orion 15.9.0-plat-10075-Tviinci distribution and moderations

ZurKaltura:


plugins/content_distribution/providers/tvinci/lib/model/TvinciDistributionProfile.php
```diff
@@ -50,7 +55,25 @@ public function getUpdateRequiredMetadataXPaths()
...
+	/**
+	 * @param entry $entry
+	 * @return bool
+	 */
+	public function shouldDistributeEntry($entry)
+	{
+		$result = true;
+		if($this->getAssetsType() == kTvinciAssetsType::REGULAR)
+		{
+			$result = $entry->getStatus() == entryStatus::READY;
```
MosheMaorKaltura: In this case call parent:: shouldDistributeEntry

plugins/content_distribution/lib/model/EntryDistribution.php
```diff
@@ -70,7 +70,7 @@ public function getNextReport($format = 'Y-m-d H:i:s')
 	/**
 	 * @return entry
 	 */
-	protected function getEntry()
+	public function getEntry()
 	{
 		$entryId = $this->getEntryId();
```
MosheMaorKaltura: This code is useless  , Since retrieveByPKNoFilter already searches in the pool.
Therefor you just call return retrieveByPKNoFilter(entryId)

PR #8811, opened 2019-10-15
SUP-19196:children_count on parent annotation cue point should decrease when deleting child

ravitshalem:


MosheMaorKaltura:
reviewed

PR #8839, opened 2019-10-28
Fix no more than one recepient is allowed bug

danielyag:


plugins/event_notification/providers/email/lib/EmailNotificationTemplateConfiguration.php
```diff
@@ -90,15 +90,27 @@ protected function getHeaderField( $headerName , $properties)
...
-
-				$recipient = new Kaltura_Client_EmailNotification_Type_EmailNotificationRecipient();
-				$recipient->email = $email;
-				$recipient->name = $name;
-
-				$recipientProvider = new Kaltura_Client_EmailNotification_Type_EmailNotificationStaticRecipientProvider();
-				$recipientProvider->emailRecipients = array();
-				$recipientProvider->emailRecipients[] = $recipient;
-
+				$delimiter = ";";
```
noam-arad: I want to be able to handle both in the same string so we'll catch also this
a@aol.com;b@aol.com,c@aol.com
You can either use preg_split or just replace all "," with ";" and use the one delimiter

plugins/event_notification/providers/email/lib/EmailNotificationTemplateConfiguration.php
```diff
@@ -110,44 +122,40 @@ protected function getHeaderField( $headerName , $properties)
...
+                /** @var Kaltura_Client_EmailNotification_Type_EmailNotificationRecipient $headerObject */
+                $fullEmailRecipientsValue .= trim($currentRecipient->email->value) . ';';
+            }
+            $headerObject = new Kaltura_Client_EmailNotification_Type_EmailNotificationRecipient();
+            $headerObject->email->value = $fullEmailRecipientsValue;
+        }
 
-			$objectEmailValue = $headerObject ? $headerObject->email->value : '';
-			$objectNameValue = $headerObject && $headerObject->name ? $headerObject->name->value : '';
+        $objectEmailValue = $headerObject ? $headerObject->email->value : '';
```
noam-arad: right now name will always be empty

plugins/event_notification/providers/email/lib/EmailNotificationTemplateConfiguration.php
```diff
@@ -133,6 +130,11 @@ protected function populateHeaderField($object , $headerName)
             }
             $headerObject = new Kaltura_Client_EmailNotification_Type_EmailNotificationRecipient();
             $headerObject->email->value = $fullEmailRecipientsValue;
+            if ($object->$headerName->emailRecipients)
+            {
+                $firstRecipient = $object->$headerName->emailRecipients[0];
```
MosheMaorKaltura: WHat is the usage of firstRecipient? WHat is the required behaviour here?

PR #8842, opened 2019-10-30
PLAT-10241: fix generate_filters error

ravitshalem:


hilak:
@ravitshalem looks good to me.

PR #8852, opened 2019-11-01
Orion 15.2.0 psvamb 6815

hilak:


hilak:
@MosheMaorKaltura  reviewed by @yossipapi

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/ApFeedDropFolderPlugin.php
```diff
@@ -0,0 +1,156 @@
...
+	/**
+	 * Returns a list of enumeration class names that implement the baseEnumName interface.
+	 * @param string $baseEnumName the base implemented enum interface, set to null to retrieve all plugin enums
+	 * @return array<string> A string listing the enum class names that extend baseEnumName
+	 */
+	public static function getEnums($baseEnumName = null)
+	{
+		if (!$baseEnumName)
+		{
+			return array('ApFeedDropFolderType');
```
MosheMaorKaltura: Use const

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/ApFeedDropFolderPlugin.php
```diff
@@ -0,0 +1,156 @@
...
+		}
+		if ($baseEnumName == 'DropFolderType')
```
MosheMaorKaltura: 1. Use === for string match
2. For 'DropFolderType' use KalturaDropFolderType::getEnumClass()

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/ApFeedDropFolderPlugin.php
```diff
@@ -0,0 +1,156 @@
...
+				{
+					return new Form_ApFeedDropFolderConfigureExtend_SubForm();
+				}
+				break;
+			case 'Kaltura_Client_DropFolder_Type_DropFolder':
+				if ($enumValue == Kaltura_Client_DropFolder_Enum_DropFolderType::AP_FEED)
+				{
+					return new Kaltura_Client_ApFeedDropFolder_Type_ApFeedDropFolder();
+				}
+				break;
```
MosheMaorKaltura: What should be the default behaviour of this switch case? should it print an message? throw exception do-nothing?
Please add default to the switch.

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/ApFeedDropFolderPlugin.php
```diff
@@ -0,0 +1,156 @@
...
+	 * @param string $baseClass The base class of the searched class
+	 * @param string $enumValue The enumeration value of the searched class
+	 * @return string The name of the searched object's class
+	 */
+	public static function getObjectClass($baseClass, $enumValue)
+	{
+		switch ($baseClass) {
+			case 'DropFolderFile':
+				if ($enumValue == self::getDropFolderTypeCoreValue(ApFeedDropFolderType::AP_FEED))
+					return 'FeedDropFolderFile';
```
MosheMaorKaltura: Use const values

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/batch/KApFeedDropFolderEngine.php
```diff
@@ -0,0 +1,150 @@
...
+	 * @var string
+	 */
+	protected $apiKey;
+	
+	const AP_FEED_PAGE_SIZE = 100;
+	
+	public function watchFolder(KalturaDropFolder $dropFolder)
+	{
+		/* @var $dropFolder KalturaApFeedDropFolder */
+		KalturaLog::info("Watching drop folder with ID [" . $dropFolder->id . "]");
```
MosheMaorKaltura: No need to use double quotes here since there is no processing , please use single.

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/batch/KApFeedDropFolderEngine.php
```diff
@@ -0,0 +1,150 @@
...
+		$this->apiKey = $dropFolder->apApiKey;
+		
+		$feedUrl = $this->dropFolder->path;
+		
+		$existingDropFolderFilesMap = $this->loadDropFolderFiles();
+		
+		$counter = 0;
+		$break = false;
+		
+		do{
```
MosheMaorKaltura: Brackets in new line.

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/batch/KApFeedDropFolderEngine.php
```diff
@@ -0,0 +1,150 @@
...
+				KalturaLog::info('Finished current run.');
+				break;
+			}
+			
+			if (isset($feed['data']['next_page']))
+			{
+				$feedUrl = $feed['data']['next_page'];
+			}
+			
+		}while ($lastPageCount == self::AP_FEED_PAGE_SIZE);
```
MosheMaorKaltura: This is a weak end case, since it is possible that due to changes that happen in parallel, one of the pages will return less than page size, still there will be more to come.
Safer approach will be until lastPageCount == 0.

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/lib/api/KalturaApFeedDropFolder.php
```diff
@@ -0,0 +1,36 @@
...
+	
+	public function getMapBetweenObjects()
+	{
+		return array_merge(parent::getMapBetweenObjects(), self::$map_between_objects);
+	}
+	
+	public function toObject($dbObject = null, $skip = array())
+	{
+		if (!$dbObject)
+			$dbObject = new ApFeedDropFolder();
```
MosheMaorKaltura: add {}

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/lib/api/KalturaApFeedDropFolder.php
```diff
@@ -0,0 +1,36 @@
...
+		
+		if ($this->feedItemInfo)
+			$dbObject->setFeedItemInfo($this->feedItemInfo->toObject());
```
MosheMaorKaltura: add {}

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/lib/model/ApFeedDropFolder.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @package plugins.FeedDropFolder
+ * @subpackage model
+ */
+class ApFeedDropFolder extends FeedDropFolder
+{
+	/**
+	 * @return the $apApiKey
+	 */
+	public function getApApiKey() {
```
MosheMaorKaltura: {]

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/batch/KApFeedDropFolderEngine.php
hilak: @MosheMaorKaltura  because this is a professionally generated news feed, odds are slim that we could catch it just at the time of day it gets updated, which is exactly why I cut off the interaction with it on what is perceived as the last page. Also I don't see how the same case can't happen even if the last page item count **is** 0. This interaction isn't a one-off - this is a drop folder pull, so it gets repeated periodically. Any content that was not ingested on a single run will likely get pulled in on the next one.

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/ApFeedDropFolderPlugin.php
hilak: @MosheMaorKaltura there is no default on any of the plugin classes that implement the loadObject through a switch case. The reason is that this method is being called from the KalturaPluginManager, which iterates over all the plugins that implement the loadObject function. If all plugins had a default logging a message, you would see several KalturaLog messages all reading "Plugin xyz doesn't handle this class". If you threw an exception, you would prevent the PluginManager from processing the next plugin class which **might** be the correct one. Any logic added in a default case would be bad here so there is no default.

plugins/bulk_upload/drop_folder_xml/plugins/providers/ap_feed/ApFeedDropFolderPlugin.php
hilak: @MosheMaorKaltura the reason this is done this way (in all of our plugins - not just this one) is because the IKalturaEnumerator::getEnums method is called from several endpoints - admin console, batch and generator. Each context has a different classmap loaded and in some cases this method is called in order to build the class map. In order to reference this value as a class enum, the class defining the enum needs to be part of the classmap, leading to an exception being thrown.

PR #8884, opened 2019-11-08
Orion 15.10.0 PSVAMB-8935 - New user-conscious 5-star rating functionality 

hilak:


plugins/rating/services/RatingService.php
```diff
@@ -0,0 +1,156 @@
...
+
+class RatingService extends KalturaBaseService
+{
+	public function initService($serviceId, $serviceName, $actionName)
+	{
+		parent::initService($serviceId, $serviceName, $actionName);
+		
+		if(!RatingPlugin::isAllowedPartner($this->getPartnerId()))
+		{
+			throw new KalturaAPIException(KalturaErrors::FEATURE_FORBIDDEN, RatingPlugin::PLUGIN_NAME);
```
MosheMaorKaltura: Since there not going to be a feature flip on this plugin, we better have an exception of plugin not enabled.

plugins/rating/services/RatingService.php
```diff
@@ -0,0 +1,156 @@
...
+		}
+		
+		if ((!kCurrentContext::$ks_uid || kCurrentContext::$ks_uid == "") && $actionName != "getRatingCounts")
```
MosheMaorKaltura: Better use kSessionBase::isAnonymousSession since you do cover all cases here. 

Use !== for string matching

plugins/rating/services/RatingService.php
```diff
@@ -0,0 +1,156 @@
...
+		{
+			throw new KalturaAPIException(KalturaErrors::INVALID_USER_ID);
```
MosheMaorKaltura: This exception does not imply on the real cause. 
Must add the following info to the API Exception - "Anonymous user is not allowed"

plugins/rating/services/RatingService.php
```diff
@@ -0,0 +1,156 @@
...
+	 * Action for current kuser ro rate a specific entry
+	 * @param string $entryId
+	 * @param int $rank
+	 * @throws KalturaErrors::ENTRY_ID_NOT_FOUND
+	 * @throws KalturaErrors::INVALID_RANK_VALUE
+	 * @return int
+	 */
+	public function rateAction ($entryId, $rank)
+	{
+		if (!$entryId)
```
MosheMaorKaltura: This check is not needed since entry has no default value , the dispatcher will validate that you have an string value in entry

plugins/rating/services/RatingService.php
```diff
@@ -0,0 +1,156 @@
...
+	 * @action removeRating
+	 * Action to remove current kuser's rating for a specific entry
+	 * @param string $entryId
+	 * @throws KalturaErrors::ENTRY_ID_NOT_FOUND
+	 * @throws KalturaRatingErrors::USER_RATING_FOR_ENTRY_NOT_FOUND
+	 * @return bool
+	 */
+	public function removeRatingAction ($entryId)
+	{
+		if (!$entryId)
```
MosheMaorKaltura: Same here , not needed.

plugins/rating/services/RatingService.php
```diff
@@ -0,0 +1,156 @@
...
+		}
+		
+		$existingKVote = kvotePeer::doSelectByEntryIdAndPuserId($entryId, $this->getPartnerId(), kCurrentContext::$ks_uid);
+		if (!$existingKVote)
+		{
+			throw new KalturaAPIException(KalturaRatingErrors::USER_RATING_FOR_ENTRY_NOT_FOUND);
+		}
+		
+		
+		if (kvotePeer::disableExistingKVote($entryId, $this->getPartnerId(), kCurrentContext::$ks_uid))
```
MosheMaorKaltura: {}

plugins/rating/services/RatingService.php
```diff
@@ -0,0 +1,156 @@
...
+			return true;
+		
+		return false;
```
MosheMaorKaltura: Would not it be more efficient to return the new rand as in ratingAction

plugins/rating/services/RatingService.php
```diff
@@ -0,0 +1,156 @@
...
+	/**
+	 * @action checkRating
+	 * Action to check current kuser's rating for a specific entry
+	 * @param string $entryId
+	 * @throws KalturaErrors::ENTRY_ID_NOT_FOUND
+	 * @return int
+	 */
+	public function checkRatingAction ($entryId)
+	{
+		if (!$entryId)
```
MosheMaorKaltura: Same here

plugins/rating/services/RatingService.php
```diff
@@ -0,0 +1,156 @@
...
+	
+	/**
+	 * @action getRatingCounts
+	 * Action to check entry's rating counts breakdown
+	 * @param KalturaRatingCountFilter $filter
+	 * @throws KalturaErrors::ENTRY_ID_NOT_FOUND
+	 * @throws KalturaRatingErrors::USER_RATING_FOR_ENTRY_NOT_FOUND
+	 * @return KalturaRatingCountListResponse
+	 */
+	public function getRatingCountsAction (KalturaRatingCountFilter $filter)
```
MosheMaorKaltura: If you want to allow calling this function with empty filter than you need to add default value null to the function signature
But, looking at getListResponse of this filter - you must have a valid filter with entry ID set , so this option is impossible

plugins/rating/services/RatingService.php
hilak: @MosheMaorKaltura  but we're **removing** the rank.

plugins/rating/services/RatingService.php
hilak: @MosheMaorKaltura  there actually will be a feature flip, but it needs to be added on the environment admin.ini file. I'll add to admin.ini.template as well.

plugins/rating/services/RatingService.php
hilak: @MosheMaorKaltura  kSessionBase::isAnonymousSession is a class method, not static.

plugins/rating/services/RatingService.php
MosheMaorKaltura: $ks = $this->getKs();
!$ks || $ks->isAnonymousSession()

plugins/rating/services/RatingService.php
MosheMaorKaltura: $ks = $this->getKs();
!$ks || $ks->isAnonymousSession()

hilak: @MosheMaorKaltura  slightly different implementation, but should still work.

PR #8885, opened 2019-11-10
PLAT-10295:remove reach vendor partner permission when performing ent…

ravitshalem:
…ryVendorTask list with admin console ks


MosheMaorKaltura:
reviewed

PR #8889, opened 2019-11-11
Kava - add QOE reports

NadavHarnik:


api_v3/lib/types/report/KalturaReportInputBaseFilter.php
```diff
@@ -79,4 +78,22 @@ public function toReportsInputFilter($reportInputFilter = null)
 		}
 		return $reportInputFilter;
 	}
+
+	protected function shouldRoundDate()
+	{
+		if (kConf::hasParam('kava_skip_date_rounding_client_tags'))
+		{
+			$skipDateRoundingClientTags = kConf::get('kava_skip_date_rounding_client_tags');
```
MosheMaorKaltura: Better usage call kConf::get with default value rather than calling kConf::hasParam and  kConf::get.

PR #8895, opened 2019-11-11
Update KFeedDropFolderEngine.php

hilak:


plugins/bulk_upload/drop_folder_xml/plugins/batch/Engine/KFeedDropFolderEngine.php
```diff
@@ -64,7 +64,7 @@ public function watchFolder(KalturaDropFolder $dropFolder) {
 		
 	}
 	
-	protected function watchProcessSingleItem (SimpleXMLElement $feedItem, array $existingDropFolderFilesMap)
+	protected function watchProcessSingleItem (SimpleXMLElement $feedItem, array &$existingDropFolderFilesMap)
```
MosheMaorKaltura: Notice that you are calling this function with third parameter - counter

PR #8899, opened 2019-11-12
SUP-18687: fix SRT line index when clipping the beginning of the entry

amiras89:


amiras89:
Closed since we need to handle all caption types (cap, srt, dfxp, webVTT) not assume it's SRT

PR #8900, opened 2019-11-12
SUP-19501:cloning file asset when cloning playlist

ravitshalem:


amiras89:
Closed since we need to handle all caption types (cap, srt, dfxp, webVTT)

PR #8930, opened 2019-11-19
orion-15.11.0-PLAT-10303-Privacy context and category names is not indexed on entry in setContent

ZurKaltura:


alpha/lib/model/categoryPeer.php
```diff
@@ -15,13 +15,40 @@ class categoryPeer extends BasecategoryPeer implements IRelatedObjectPeer
...
 	private static $ignoreDeleted = false;
-	
+
+	protected static $categoryEntitlementEnforcement = true;
+
+	public static function disableCategoryEntitlementEnforcement()
+	{
+		$currentState = self::$categoryEntitlementEnforcement;
+		self::$categoryEntitlementEnforcement = false;
+		if($currentState)
```
MosheMaorKaltura: this check is not needed

alpha/lib/model/categoryPeer.php
```diff
@@ -15,13 +15,40 @@ class categoryPeer extends BasecategoryPeer implements IRelatedObjectPeer
...
+		{
+			categoryPeer::setDefaultCriteriaFilter();
+		}
+	}
+
+	public static function enableCategoryEntitlementEnforcement()
```
MosheMaorKaltura: this function is not needed

PR #8933, opened 2019-11-20
SUP-20245: add Price > 0 when fetching all vendor task to avoid return unnecessary results which may exceed the php memory_limit

amiras89:


amiras89:
Per Moshe changing the fix from GREATER_THAN to NOT_EQUAL so that we can 'refund' clients in case we want to

PR #8934, opened 2019-11-20
SUP-20245: add Price <> 0 when fetching all vendor task to avoid return unnecessary results which may exceed the php memory_limit

amiras89:


MosheMaorKaltura:
reviewed

amiras89:
During the testing, the issue has been resolved (when we fetch the query, we no longer bring a whole year QUEUE_TIME)

Hot Fix aborted

PR #8943, opened 2019-11-24
SUP-20245: fetch SUM(price) + price <> 0 to avoid reaching PHP memory_limit due to too many results + Eran's improvements

amiras89:


amiras89:
@MosheMaorKaltura  , @erankor  , @yossipapi  I've added the modifications Eran suggested and also tested it. It works the same way as before (without returning matrix, returning regular array)

PR #8957, opened 2019-11-26
PLAT-10188: add login using sso to admin console

inbal-ben-david:


admin_console/views/scripts/user/logout.phtml
```diff
@@ -0,0 +1,10 @@
+<div class="box">
+	<div class="header">
+		<h1>Logged out Successfully</h1>
```
MosheMaorKaltura: Successfully to lower case

admin_console/controllers/UserController.php
```diff
@@ -193,14 +209,68 @@ public function loginAction()
...
 	
 	public function logoutAction()
 	{
 		Zend_Session::forgetMe();
 		$client = Infra_ClientHelper::getClient();
 		$client->session->end();
 		Infra_AuthHelper::getAuthInstance()->clearIdentity();
-		$this->_helper->redirector('index', 'index');
+
+		$settings = Zend_Registry::get('config')->settings;
```
MosheMaorKaltura: Since this code repeats twice (and may be more in the future), create a small function - shouldUseSso()

alpha/scripts/utils/enableAdminConsoleSso.php
```diff
@@ -0,0 +1,41 @@
...
+}
+
+function allowSsoFeature($partner)
+{
+	$partner->setUseSso(true);
+	$partner->setBlockDirectLogin(true);
+	$partner->save();
+}
+
+function addSsoProfile($domain, $redirectUrl)
```
MosheMaorKaltura: What about the SSO data used for the authentication, how do you add it to the profile?

MosheMaorKaltura:
reviewed

PR #8959, opened 2019-11-27
Orion 15.11.0-PLAT-10317 clip concat closer

ZurKaltura:


MosheMaorKaltura:
Reviewed. Make sure to set correct var in template \ config so this closer will not be activated too often

PR #8976, opened 2019-12-01
PLAT-10338: dont copy flavors in case replacement entry was dele…

inbal-ben-david:
…ted and remove cache invalidation

### DIFF
--- a/alpha/apps/kaltura/lib/batch2/bcdl/kReplacementHelper.php
+++ b/alpha/apps/kaltura/lib/batch2/bcdl/kReplacementHelper.php
@@ -276,13 +276,21 @@ class kReplacementHelper
 	{
 		$defaultThumbAssetOld = null;
 		$defaultThumbAssetNew = null;
+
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/lib/batch2/bcdl/kReplacementHelper.php
```diff
@@ -278,6 +278,11 @@ public static function copyReadyReplacingEntryAssetToReplacedEntry($replacingEnt
 		$defaultThumbAssetNew = null;
 		$replacingEntry = $replacingEntryAsset->getentry();
 		$replacedEntry = entryPeer::retrieveByPK($replacingEntry->getReplacedEntryId());
+		if(!$replacingEntry || !$replacedEntry)
+		{
+			KalturaLog::info("One of the replacement entries is missing");
```
MosheMaorKaltura: Add move valid info here - replacing - exits or not , and replaced - exit of not,

PR #8979, opened 2019-12-01
PLAT-10339: use replaced entry for flavors deletion on handleConvertF…

inbal-ben-david:
…ailed


alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2171,6 +2165,22 @@ public static function handleConvertProfileFinished(BatchJob $dbBatchJob, kConve
 		return $dbBatchJob;
 	}
 
+	protected static function getEntryToUpdate($dbBatchJob)
```
MosheMaorKaltura: rename - getEntryIdToUpdate

PR #9005, opened 2019-12-09
PLAT-10366:fixing event notifications not being triggered when vendors doing entryVendorTask updateJob

ravitshalem:
…teJob


MosheMaorKaltura:
Reviewed

PR #9037, opened 2019-12-18
Orion 15.13.0 psvamb 9373

TalDubovKaltura:


batch/batches/ExportCsv/Engine/KObjectExportEngine.php
```diff
@@ -8,7 +8,7 @@ abstract class KObjectExportEngine
 {
 	/**
 	 * @param int $objectType of enum KalturaCopyObjectType
-	 * @return KCopyingEngine
+	 * @return KUserExportEngine
```
hilak: @TalDubovKaltura this was originally my bad - could you please change the return type here to "KObjectExportEngine"? We're not necessarily retuning a user exporting engine.

batch/batches/ExportCsv/Engine/KObjectExportEngine.php
```diff
@@ -8,7 +8,7 @@ abstract class KObjectExportEngine
 {
 	/**
 	 * @param int $objectType of enum KalturaCopyObjectType
```
hilak: @TalDubovKaltura  this was originally my bad, could I ask you to change enum type in this comment to KalturaExportObjectType?

plugins/search/providers/elastic_search/batch/KExportMediaEsearchEngine.php
```diff
@@ -10,10 +10,12 @@ class KExportMediaEsearchEngine extends KObjectExportEngine
 	const LIMIT = 10000;
 	
 	const PAGE_SIZE = 500;
+
+	const DATE_FORMAT = 'c';
 	
 	public function fillCsv(&$csvFile, &$data)
 	{
-		KalturaLog::info ('Exporting content for media items through Esearch');
+		KalturaLog::info ('Exporting content for media items through Esearch, data:'.json_encode($data));
```
hilak: @TalDubovKaltura  why do you need to print the data here? It gets printed as part of the batch job object when the job starts running anyway.

plugins/search/providers/elastic_search/batch/KExportMediaEsearchEngine.php
```diff
@@ -128,12 +131,22 @@ protected function retrieveEntryCategories ($entryId)
...
 		return $result;
 	}
-}
\ No newline at end of file
+	/**
+	 * @param $timestamp
+	 * @param $data
+	 * @return false|string
+	 */
+	protected function makeTimeStampHumanReadableIfNeeded($timestamp, $data) {
```
hilak: @TalDubovKaltura please change the method name to something less specific - formatTimestamp for instance. You also don't need to pass the entire data object into this method - the $data->options would suffice.

plugins/search/providers/elastic_search/lib/api/types/data/KalturaMediaEsearchExportToCsvJobData.php
```diff
@@ -12,9 +12,15 @@ class KalturaMediaEsearchExportToCsvJobData extends KalturaExportCsvJobData
 	 * @var KalturaESearchEntryParams
 	 */
 	public $searchParams;
+	/**
+	 * options
+	 * @var KalturaExportToCsvOptionsType
+	 */
+	public $options;
```
hilak: @TalDubovKaltura we discussed this - the $options needs to be an array allowing to pass several different options, rather than just one.

batch/batches/ExportCsv/Engine/KObjectExportEngine.php
TalDubovKaltura: @hilak Done. actually it was my bad too because I didn't completely interpret which type of value the method returns, and it should be the parent class like you wrote.

batch/batches/ExportCsv/Engine/KObjectExportEngine.php
TalDubovKaltura: Done.

plugins/search/providers/elastic_search/batch/KExportMediaEsearchEngine.php
TalDubovKaltura: I just figured out at this point, when the process starting to fill the CSV that I don't have the info in the logs about the data that will be passed to the CSV generator. but if you think it's unuseful I can remove it

plugins/search/providers/elastic_search/batch/KExportMediaEsearchEngine.php
TalDubovKaltura: About the options parameter, I absolutely agree with you and I changed it.
as for the name, I changed it but I do think that function names should be meaningful and specific as possible, it is a known best practice.  while I'm calling a method called 'formatTimestamp' as a developer I'm expecting to pass it a parameter of format type, otherwise, I will ask myself which formatting the method using and I will assume that it will be formatted anyway(that's why I used the 'IfNeeded' in the name).

plugins/search/providers/elastic_search/lib/api/types/data/KalturaMediaEsearchExportToCsvJobData.php
TalDubovKaltura: You're right, we already spoke about that. I haven't reached it yet due to QA issues I had and this change will almost start the same process again so I will talk Oleg about that, to understand how much time can we have more.

plugins/search/providers/elastic_search/lib/model/data/kMediaEsearchExportToCsvJobData.php
```diff
@@ -6,24 +6,38 @@
 
 class kMediaEsearchExportToCsvJobData extends kExportCsvJobData
 {
+	/**
+	 * @var bool
```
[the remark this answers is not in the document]
TalDubovKaltura: Currently, the options is the value HUMAN_READABLE_DATES which is boolean, but after changing it to array it will not be boolean for sure

PR #9048, opened 2019-12-22
Orion 15.13.0 webc 1574

danielyag:


api_v3/services/LiveStreamService.php
```diff
@@ -577,4 +577,102 @@ public function regenerateStreamTokenAction($entryId)
...
+        }
+        $cuePoints = CuePointPeer::retrieveByEntryId($liveEntryId);
+
+        $shouldBeDeletedCuePoints = array();
+
+        $notDeletedCuePointTagsArray = explode(",", $archiveOptions->notDeletedCuePointTags);
+        foreach ($cuePoints as $cuePoint)
+        {
+            /** @var CuePoint $cuePoint */
+            KalturaLog::info("Iterating over cue point [" . $cuePoint->getId() . "]");
```
MosheMaorKaltura: Change to ' 
Coding standard

api_v3/services/LiveStreamService.php
```diff
@@ -577,4 +577,102 @@ public function regenerateStreamTokenAction($entryId)
...
+        if (!$liveEntry)
+        {
+            KalturaLog::err("Live entry id [" . $liveEntryId . "] not found");
+            throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $liveEntryId);
+        }
+        $cuePoints = CuePointPeer::retrieveByEntryId($liveEntryId);
+
+        $shouldBeDeletedCuePoints = array();
+
+        $notDeletedCuePointTagsArray = explode(",", $archiveOptions->notDeletedCuePointTags);
```
MosheMaorKaltura: Same '

api_v3/services/LiveStreamService.php
```diff
@@ -577,4 +577,102 @@ public function regenerateStreamTokenAction($entryId)
...
+     * @throws KalturaClientException
+     * @throws PropelException
+     */
+	public function archiveAction($liveEntryId, KalturaLiveEntryArchiveOptions $archiveOptions)
+    {
+        $liveEntry = entryPeer::retrieveByPK($liveEntryId);
+        /** @var LiveStreamEntry $liveEntry */
+        if (!$liveEntry)
+        {
+            KalturaLog::err("Live entry id [" . $liveEntryId . "] not found");
```
yossipapi: No need for the debug the exception will already print the error.

api_v3/services/LiveStreamService.php
```diff
@@ -577,4 +577,102 @@ public function regenerateStreamTokenAction($entryId)
...
+            throw new KalturaAPIException(KalturaErrors::ENTRY_ID_NOT_FOUND, $liveEntryId);
+        }
+        $cuePoints = CuePointPeer::retrieveByEntryId($liveEntryId);
```
yossipapi: This could return large result so better do in parts or even considering moving this entire logic to batch .

api_v3/services/LiveStreamService.php
yossipapi: Better use array intersect.

plugins/cue_points/base/lib/model/CuePointPeer.php
```diff
@@ -365,4 +365,17 @@ public static function validateMetadataObjectAccess($objectId)
...
+    {
+        $c = new Criteria();
+        $c->add(CuePointPeer::ID, $ids, Criteria::IN);
+        $c->add(CuePointPeer::STATUS, $newStatus, Criteria::NOT_EQUAL);
+
+        $update = new Criteria();
+        $update->add(CuePointPeer::STATUS, $newStatus);
+
+        $con = Propel::getConnection(MetadataPeer::DATABASE_NAME);
+        return BasePeer::doUpdate($c, $update, $con);
```
yossipapi: This wont update the object in sphinx so this is now good.

yossipapi: Look at  kCuePointManager to see how we do it there.

api_v3/services/LiveStreamService.php
```diff
@@ -577,4 +577,102 @@ public function regenerateStreamTokenAction($entryId)
...
+        $shouldBeDeletedCuePoints = array();
+
+        $notDeletedCuePointTagsArray = explode(",", $archiveOptions->notDeletedCuePointTags);
+        foreach ($cuePoints as $cuePoint)
+        {
+            /** @var CuePoint $cuePoint */
+            KalturaLog::info("Iterating over cue point [" . $cuePoint->getId() . "]");
+            $cuePointTags = explode(',', $cuePoint->getTags());
+            $containsImportantTag = false;
+            foreach ($notDeletedCuePointTagsArray as $importantCuePointTag)
```
MosheMaorKaltura: An optimization is required here.
1. If no tag was set on cue point iteration is not needed.
2. Since Since it we will less likely to have all on tags on a cuepointm than it is better to loop over the cuepoint tags, than to look each time on the list of tags.

api_v3/services/LiveStreamService.php
```diff
@@ -577,4 +577,102 @@ public function regenerateStreamTokenAction($entryId)
...
+            }
+            if (!$containsImportantTag)
+            {
+                $shouldBeDeletedCuePoints[] = $cuePoint->getId();
+            }
+        }
+
+        CuePointPeer::setCuePointsStatus($shouldBeDeletedCuePoints, CuePointStatus::DELETED);
+
+        $vodEntry = entryPeer::retrieveByPK($liveEntry->getRecordedEntryId());
```
yossipapi: Better check if recorded entry id exists and change the logic a bit to match this check.

api_v3/services/LiveStreamService.php
```diff
@@ -577,4 +577,102 @@ public function regenerateStreamTokenAction($entryId)
...
+        if (!$vodEntry)
+        {
+            KalturaLog::warning("Recorded entry id [" . $liveEntry->getRecordedEntryId() . "] not found");
+        }
+        else {
```
MosheMaorKaltura: { in new line.

api_v3/services/LiveStreamService.php
```diff
@@ -577,4 +577,102 @@ public function regenerateStreamTokenAction($entryId)
...
+                $shouldBeDeletedCuePoints[] = $cuePoint->getId();
+            }
+        }
+
+        CuePointPeer::setCuePointsStatus($shouldBeDeletedCuePoints, CuePointStatus::DELETED);
+
+        $vodEntry = entryPeer::retrieveByPK($liveEntry->getRecordedEntryId());
+        if (!$vodEntry)
+        {
+            KalturaLog::warning("Recorded entry id [" . $liveEntry->getRecordedEntryId() . "] not found");
```
MosheMaorKaltura: If no processing is used in string - use sinegle '

plugins/cue_points/base/lib/model/CuePointPeer.php
```diff
@@ -365,4 +365,17 @@ public static function validateMetadataObjectAccess($objectId)
 	
 		return true;
 	}
+
+	public static function setCuePointsStatus($ids, $newStatus)
```
MosheMaorKaltura: indentation

api_v3/services/LiveStreamService.php
```diff
@@ -577,4 +577,102 @@ public function regenerateStreamTokenAction($entryId)
...
+            $pushNotificationTemplateService->sendCommandAction('POLLS_PUSH_NOTIFICATIONS',
+                $pushNotificationParams, KalturaPushNotificationCommandType::CLEAR_QUEUE);
+            $pushNotificationTemplateService->sendCommandAction('PUBLIC_QNA_NOTIFICATIONS',
+                $pushNotificationParams, KalturaPushNotificationCommandType::CLEAR_QUEUE);
+            $pushNotificationTemplateService->sendCommandAction('USER_QNA_NOTIFICATIONS',
+                $pushNotificationParams, KalturaPushNotificationCommandType::CLEAR_QUEUE);
+        }
+        else
+        {
+            KalturaLog::info("Plugin [" . PushNotificationPlugin::getPluginName() . "] is not loaded");
```
yossipapi: If plugin not enabled no need to return false.
This method should always return true or throw exception.

plugins/cue_points/base/lib/model/CuePointPeer.php
MosheMaorKaltura: Should be changed to deleteCuePoints($ids, $entryId)
The entryID is for protection.
I see no case where we make vast change of cuepoints status plus it can open case for issues.

batch/batches/LiveToVod/KAsyncLiveToVod.class.php
```diff
@@ -37,7 +37,25 @@ protected function getJobType()
 	 */
 	protected function exec(KalturaBatchJob $job)
 	{
-        return $this->copyCuePoint($job, $job->data);
+		$jobRes = $this->copyCuePoint($job, $job->data);
```
MosheMaorKaltura: Indentation.

batch/batches/LiveToVod/KAsyncLiveToVod.class.php
```diff
@@ -37,7 +37,25 @@ protected function getJobType()
...
+        try {
+            $liveEntryId = $job->data->liveEntryId;
+            $liveEntry = KBatchBase::$kClient->baseEntry->get($liveEntryId);
+            /** @var KalturaLiveStreamEntry $liveEntry */
+            $recordStatus = $liveEntry->recordingStatus;
+            $shouldAutoArchive = $liveEntry->recordingOptions->shouldAutoArchive;
+            if ($recordStatus == KalturaRecordStatus::PER_SESSION && $shouldAutoArchive == KalturaNullableBoolean::TRUE_VALUE) {
+                $archiveOptions = new KalturaLiveEntryArchiveOptions();
+                $archiveOptions->notDeletedCuePointTags = "entry-reset-mode,webcast:internal-state,player-qna-settings-update,select-deck-document,poll-data,timeline-assets-state,rtc-settings";
```
MosheMaorKaltura: Single '

batch/batches/LiveToVod/KAsyncLiveToVod.class.php
```diff
@@ -37,7 +37,25 @@ protected function getJobType()
...
+                KBatchBase::$kClient->liveStream->archive($liveEntryId, $archiveOptions);
+            }
+        }
+        catch (Exception $e)
+        {
+            KalturaLog::err($e->getMessage());
+        }
+
+		return $jobRes;
```
MosheMaorKaltura: indentation

batch/batches/LiveToVod/KAsyncLiveToVod.class.php
```diff
@@ -37,7 +37,25 @@ protected function getJobType()
 	 */
 	protected function exec(KalturaBatchJob $job)
 	{
-        return $this->copyCuePoint($job, $job->data);
+		$jobRes = $this->copyCuePoint($job, $job->data);
+        try {
+            $liveEntryId = $job->data->liveEntryId;
+            $liveEntry = KBatchBase::$kClient->baseEntry->get($liveEntryId);
```
yossipapi: Check value returned.

batch/batches/LiveToVod/KAsyncLiveToVod.class.php
```diff
@@ -37,7 +37,25 @@ protected function getJobType()
...
+            /** @var KalturaLiveStreamEntry $liveEntry */
+            $recordStatus = $liveEntry->recordingStatus;
+            $shouldAutoArchive = $liveEntry->recordingOptions->shouldAutoArchive;
```
yossipapi: recordingOptions might not exist

batch/batches/LiveToVod/KAsyncLiveToVod.class.php
yossipapi: Better move the tags to the recording options and make thus making it more generic.

batch/batches/Archive/KAsyncLiveEntryArchive.class.php
```diff
@@ -0,0 +1,136 @@
...
+            try
+            {
+                $cuePointsList = KBatchBase::$kClient->cuePoint->listAction($filter, $pager);
+                foreach ($cuePointsList->objects as $cuePoint)
+                {
+                    $cuePointTags = explode(',', $cuePoint->tags);
+                    $relevantTags = array_intersect($notDeletedCuePointTagsArray, $cuePointTags);
+                    if (count($relevantTags) == 0)
+                    {
+                        KBatchBase::$kClient->cuePoint->updateStatus($cuePoint->id, CuePointStatus::DELETED);
```
MosheMaorKaltura: The try - catch should apply only on specific cue point rather then all the process, since it is possible that in other PAI call user will delete a cuepoint from your list , but you will still need to delete the others.

batch/batches/Archive/KAsyncLiveEntryArchive.class.php
```diff
@@ -0,0 +1,136 @@
...
+                    }
+                }
+            }
+            catch (Exception $ex)
+            {
+                KalturaLog::err('Failed to delete cue points for live entry ' . $entryId);
```
MosheMaorKaltura: In case of failure , add the spcific cue point who failre to the message

batch/batches/Archive/KAsyncLiveEntryArchive.class.php
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+    }
+
+    private static function getCuePointFilter($entryId)
```
MosheMaorKaltura: protected

batch/batches/Archive/KAsyncLiveEntryArchive.class.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        $filter = new KalturaCuePointFilter();
+        $filter->entryIdEqual = $entryId;
+        $filter->statusIn = CuePointStatus::READY;
+        return $filter;
+    }
+
+    private static function getCuePointCount($entryId)
```
MosheMaorKaltura: protected + avoid using static in this case

deployment/permissions/service.livestream.ini
```diff
@@ -130,3 +130,12 @@ permissionItem19.service = livestream
...
 permissionItem19.permissions = -5>MEDIA_SERVER_BASE, -5>PARTNER_-5_GROUP_*_PERMISSION,MEDIA_SERVER_PARTNER_LEVEL
+
+permissionItem20.service = livestream
+permissionItem20.action = archive
+permissionItem20.partnerId = -1
+permissionItem20.param3 =
+permissionItem20.param4 =
+permissionItem20.param5 =
+permissionItem20.tags =
+permissionItem20.permissions = BATCH_BASE, PARTNER_-1_GROUP_*_PERMISSION
```
MosheMaorKaltura: Who is supposed to call this api? is only the batch able to call it? does not make sense

deployment/permissions/service.pushnotification.pushnotificationtemplate.ini
```diff
@@ -8,3 +8,8 @@ permissionItem12.service = pushnotification_pushnotificationtemplate
 permissionItem12.action = sendCommand
 permissionItem12.partnerId = 0
 permissionItem12.permissions = WEBCAST_PRODUCER_PERMISSION
+
+permissionItem13.service = pushnotification_pushnotificationtemplate
+permissionItem13.action = sendCommand
```
MosheMaorKaltura: Every permission items presents an action, in this case you override sendCommand permissions, yo should add this permissions to the above item

batch/batches/Archive/KAsyncLiveEntryArchive.class.php
```diff
@@ -0,0 +1,136 @@
...
+        return KalturaBatchJobType::LIVE_ENTRY_ARCHIVE;
+    }
+
+    /**
+     * @param KalturaBatchJob $job
+     * @return KalturaBatchJob
+     */
+    protected function exec(KalturaBatchJob $job)
+    {
+        KalturaLog::info("Starting live entry archive! big news!");
```
noam-arad: no need for this log message, every batch already has an informative "starting" log line

batch/batches/Archive/KAsyncLiveEntryArchive.class.php
```diff
@@ -0,0 +1,136 @@
...
+        $jobData = $job->data;
+        /** @var KalturaLiveEntryArchiveJobData $jobData*/
+        $liveEntryId = $jobData->liveEntryId;
+        $notDeletedCuePointTags = $jobData->nonDeletedCuePointsTags;
```
noam-arad: get this from the liveEntry itself

batch/batches/Archive/KAsyncLiveEntryArchive.class.php
noam-arad: this updateStatus should be done as part of a multirequest

deployment/permissions/service.pushnotification.pushnotificationtemplate.ini
noam-arad: we tried that but it was not successful since we wanted this to happen on partner -1 and partner 0
also this will cause both permissions to work and not override

deployment/permissions/service.livestream.ini
noam-arad: Currently yes
in the future we want to be able to use this API as part of the webcasting app.

PR #9073, opened 2019-12-29
PLAT-10466 - handle invalid chars for partner describeYourself field

gotlieb:
@MosheMaorKaltura

### DIFF
--- a/api_v3/lib/types/partner/KalturaPartner.php
+++ b/api_v3/lib/types/partner/KalturaPartner.php
@@ -5,6 +5,8 @@
  */
 class KalturaPartner extends KalturaObject implements IFilterable
 {
+	const MAX_DESCRIBE_YOURSLEF_LEN = 64;
+
 	/**
 	 * @var int
 	 * @readonly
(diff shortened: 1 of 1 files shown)


api_v3/lib/types/partner/KalturaPartner.php
```diff
@@ -444,7 +445,24 @@ public function toPartner()
...
+	public function toObject($dbPartner = null, $skip = array())
+	{
+		if(!$dbPartner)
+		{
+			$dbPartner = new Partner();
+		}
+
+		if ($this->describeYourself && (strlen ($this->describeYourself) > self::MAX_DESCRIBE_YOURSLEF_LEN))
+		{
+			$this->describeYourself = kString::alignUtf8String($this->describeYourself, 64);
```
MosheMaorKaltura: 1. Use const here (instead of 64)
2. Apply this fix to all the relevant fields that can have an issue here.

api_v3/lib/types/partner/KalturaPartner.php
```diff
@@ -428,8 +430,7 @@ public function toPartner()
 		$this->validatePartnerPackageForInsert();
 		$this->validateForInsert();
 
-		$partner = new Partner();
-		$partner = parent::toObject( $partner );
+		$partner = parent::toObject();
```
MosheMaorKaltura: this should be changed to $this->toObject() isn't it?

PR #9089, opened 2020-01-06
Add auto archive batch configuration

danielyag:


configurations/batch/batch.ini.template
```diff
@@ -68,6 +68,7 @@ enabledWorkers.KAsyncCopyCaptions					= 1
 enabledWorkers.KAsyncCopyCuePoints                   = 1
 enabledWorkers.KSyncReachCreditTaskRunner	    = 0
 enabledWorkers.KAsyncExportCsv                      = 0
+enabledWorkers.KAsyncLiveEntryArchive               = 1
```
yossipapi: Any reason this was added to the template? Do we really need to run this batch on all prod machines?

danielyag: We added it to work as LiveToVOD

configurations/batch/batch.ini.template
```diff
@@ -800,6 +801,12 @@ scriptPath                                          = batches/ServerNodeMonitor/
 params.serverNodeTTL                                = 120
 params.typesToMonitor                               = 1
 
+[KAsyncLiveEntryArchive : JobHandlerWorker]
+id                                                  = 680
```
yossipapi: Please note this id is already taken by KAsyncCopyCaptions
[KAsyncCopyCaptions : JobHandlerWorker]
id                                    = 680

Please double check yourself before attaching an id to a worker

configurations/batch/batch.ini.template
yossipapi: So why do you need another worker if this is part of the liveTovod process?
Am I missing something?

danielyag: It has two scenarios: 
1) Work automatically after the live to VOD finished
2) Work when a proper API call was triggered

PR #9094, opened 2020-01-07
PLAT-10428 reorder audio flavors to end of flavors results in playman…

gotlieb:
…ifest request when muxedAudioLang params exist

@MosheMaorKaltura


alpha/lib/model/DeliveryProfileVodPackagerHlsManifest.php
```diff
@@ -26,7 +26,35 @@ public function buildServeFlavors()
...
+	{
+		$sortedFlavors = parent::sortFlavors($flavors);
+		if($this->params->getMuxedAudioLanguage())
+		{
+			//Order audio flavors after video flavors
+			$audioflavors = array();
+			$videoflavors = array();
+			foreach ($sortedFlavors as $flavor)
+			{
+				if (!isset($flavor['audioCodec']) && !isset($flavor['audioLanguageName']))
```
MosheMaorKaltura: Use const.

PR #9131, opened 2020-01-26
SUP-20262: increase process timeout to 20 min for entries > 2GB

amiras89:


amiras89:
forgot to add documentation
let me know if I should add
Thanks

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -265,6 +269,7 @@ protected function doSubmit(KalturaDistributionSubmitJobData $data, KalturaYoutu
 			$media->setFileSize(filesize($videoPath));
 			$ingestedVideo = self::uploadInChunks($media,$videoPath, self::DEFAULT_CHUNK_SIZE_BYTE);
 			$client->setDefer(false);
+			$this->shouldIncreaseProcessedTimeout(filesize($videoPath));
```
ZurKaltura: You call this function should which usually means it should return true or false but the function actually does change the timeout

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -41,6 +41,10 @@ class YoutubeApiDistributionEngine extends DistributionEngine implements
 
 	const TIME_TO_WAIT_FOR_YOUTUBE_TRANSCODING = 5;
 
+	const SIZE_2_GB = 2147483648;
+
+	const LONG_PROCESS_TIMEOUT = 1200;
```
ZurKaltura: I think it better to change it to a configuration value instead of const in the code

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
amiras89: Changed to "setLongProcessedTimeout($fileSize);"
But it doesn't always 'set' it (depends on fileSize)
If you have a better name suggestions - I'm open to it.

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
amiras89: changed to variables

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -682,5 +687,25 @@ private static function uploadChunk($media, $chunk)
...
 	}
 
+	/**
+	 * @param int $fileSize
+	 */
+	protected function setLongProcessedTimeout($fileSize)
+	{
+		if (isset(KBatchBase::$taskConfig->params->youtubeApi))
+		{
+			if (isset(KBatchBase::$taskConfig->params->youtubeApi->longProcessedTimeout))
```
MosheMaorKaltura: Use {} for every if

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -682,5 +687,25 @@ private static function uploadChunk($media, $chunk)
...
+				$this->longProcessedTimeout = KBatchBase::$taskConfig->params->youtubeApi->longProcessedTimeout;
```
MosheMaorKaltura: If value is not set in the configuration file - supply default value hard coded

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -682,5 +687,25 @@ private static function uploadChunk($media, $chunk)
...
+
+			if (isset(KBatchBase::$taskConfig->params->youtubeApi->bigFile))
```
MosheMaorKaltura: Same here - if no value , then set default hard coded.

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -682,5 +687,25 @@ private static function uploadChunk($media, $chunk)
...
+				$this->bigFile = KBatchBase::$taskConfig->params->youtubeApi->bigFile;
+		}
+
+		if ($fileSize > $this->bigFile)
+		{
+			$this->processedTimeout = $this->longProcessedTimeout;
+			KalturaLog::info('Increased processed timeout to '.$this->processedTimeout.' seconds for file larger than 2GB');
```
MosheMaorKaltura: Change the message - larger then $filesize

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
amiras89: done

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -30,7 +30,9 @@ class YoutubeApiDistributionEngine extends DistributionEngine implements
 {
 	protected $tempXmlPath;
 	protected $timeout = 90;
-	protected  $processedTimeout = 300;
+	protected $processedTimeout = 300;
+	protected $longProcessedTimeout = 1200;
+	protected $bigFile = 2147483648;
```
amiras89: @MosheMaorKaltura default values

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
amiras89: default values set lines 33-35

plugins/content_distribution/providers/youtube_api/lib/YoutubeApiDistributionEngine.php
```diff
@@ -30,7 +30,9 @@ class YoutubeApiDistributionEngine extends DistributionEngine implements
 {
 	protected $tempXmlPath;
 	protected $timeout = 90;
-	protected  $processedTimeout = 300;
+	protected $processedTimeout = 300;
+	protected $longProcessedTimeout = 1200;
```
MosheMaorKaltura: The idea was to use this default values if they are not set in configuration

amiras89: I set both properties on the class, then we either override them from config (if config exist) otherwise we keep the default values defined on class.

protected $processedTimeout = 300;
protected $longProcessedTimeout = 1200;

Let me know if I confused it.

MosheMaorKaltura:
reviewed

PR #9153, opened 2020-02-03
PLAT-10547: add timezone to liveentry recording options for archive

noam-arad:


alpha/lib/data/live/kLiveEntryRecordingOptions.php
```diff
@@ -39,6 +39,12 @@ class kLiveEntryRecordingOptions
      */
 	protected $nonDeletedCuePointsTags;
 
+	/**
+	 * @var string
+	 */
+	public $archiveVodSuffixTimezone;
```
david-winder-kaltura: Better be "protected"

PR #9163, opened 2020-02-06
Update KFeedDropFolderEngine.php

hilak:
When a feed item has no valid content URL or GUID, a drop folder file object should not be created.

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9163)


MosheMaorKaltura:
reviewed

PR #9164, opened 2020-02-06
CSV Bulk Event Scheduling - Add option to specify resource by ID

hilak:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9164)


plugins/schedule/bulk_upload/batch/BulkUploadScheduleEventCsvEngine.php
```diff
@@ -462,7 +472,8 @@ protected function getColumns()
 			'tags',
 			'categoryIds',
 			'categoryPaths',
-			'resource',
+			'resourceID',
```
MosheMaorKaltura: Changing the name of the field though to a better name , will cause incompatibility with customer already using it.
Must return to use the old name, or add the name and always check both.
Column names in CSV is same as property names in API.

PR #9167, opened 2020-02-09
SUP-19612: change Zoom video upload to add categoryEntry instead of entry->cateogy which is deprecated

amiras89:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9167)

### DIFF
--- a/alpha/lib/model/categoryEntry.php
+++ b/alpha/lib/model/categoryEntry.php
@@ -291,6 +291,8 @@ class categoryEntry extends BasecategoryEntry implements IRelatedObject
 		if(!$categoryEntryExists)
 		{
 			$categoryEntry = new categoryEntry();
+			$categoryEntry->setCategoryId($categoryId);
+			$categoryEntry->setEntryId($entryId);
 		}
 		else
 		{
--- a/alpha/lib/model/categoryEntry.php
+++ b/alpha/lib/model/categoryEntry.php
@@ -219,5 +219,117 @@ class categoryEntry extends BasecategoryEntry implements IRelatedObject
 	{
 		$this->putInCustomData('creatorPuserId', $v);
 	}
+
+	/**
+	 * @param $entryId
+	 * @param $categoryId
+	 * @throws KalturaErrors::CATEGORY_NOT_FOUND
+	 * @throws KalturaErrors::MAX_CATEGORIES_FOR_ENTRY_REACHED
+	 * @throws KalturaErrors::CANNOT_ASSIGN_ENTRY_TO_CATEGORY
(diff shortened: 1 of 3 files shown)


plugins/vendor/zoom/lib/model/kZoomEngine.php
```diff
@@ -353,11 +353,18 @@ protected function createEntryDescriptionFromMeeting($meeting)
...
 		{
-			$entry->setCategories($zoomIntegration->getZoomCategory());
+			$categoryEntry = new categoryEntry();
+			$categoryEntry->setEntryId($entry);
+			$categoryEntry->setCategoryId($zoomIntegration->getZoomCategoryId());
+			$categoryEntry->setPartnerId($entry->getPartnerId());
+			$categoryEntry->setCreatorKuserId($entry->getKuserId());
+			$categoryEntry->setStatus(CategoryEntryStatus::ACTIVE);
+			$categoryEntry->save();
+			KalturaLog::info("Entry Id '" . $entry->getId() ."' added to Zoom category Id " . $zoomIntegration->getZoomCategoryId());
```
MosheMaorKaltura: If a string does not process values , don't use " prefer single quote '

amiras89: Changed to single quotes

plugins/vendor/zoom/lib/model/kZoomEngine.php
```diff
@@ -353,11 +353,18 @@ protected function createEntryDescriptionFromMeeting($meeting)
 	 * @param entry $entry
 	 * @throws kCoreException
 	 */
-	protected function setEntryCategory($zoomIntegration, $entry)
+	protected function createCategoryEntry($zoomIntegration, $entry)
```
MosheMaorKaltura: This code should move to categoryentry core object
Adding new funino categoryEntry::add(entry, category)

When it is added , you must behave the same way cateogryEntryService::add handles the moderation.

alpha/lib/model/categoryEntry.php
```diff
@@ -219,5 +219,120 @@ public function setCreatorPuserId ($v)
...
+	public function add($entryId, $categoryId)
+	{
+		/*=========================================================*/
+		$entry = entryPeer::retrieveByPK($entryId);
+		if(!$entry)
+		{
+			throw new KalturaAPIException(KalturaErrors::INVALID_ENTRY_ID, $entryId);
+		}
+
+		$category = categoryPeer::retrieveByPK($categoryId);
```
MosheMaorKaltura: The entire function must be broken to smaller parts - 
function validateMaxCategoryEntries
function validateAccess 
function validateNotExist
function handleModeration
function assignUsers

PR #9197, opened 2020-02-19
Orion-15.18.0-PLAT-10706-Add support for isQuiz filter

ZurKaltura:
to the query eSearch adapter

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9197)


plugins/search/providers/elastic_search/lib/model/filters/ESearchQueryFromAdvancedSearch.php
```diff
@@ -100,6 +103,25 @@ protected function createESearchQueryFromEntryCatpionAdvancedFilter(kEntryCaptio
...
+		$item = new ESearchEntryItem();
+		$item->setFieldName(ESearchEntryFieldName::IS_QUIZ);
+		$item->setItemType(ESearchItemType::EXISTS);
+		if($filter->getIsQuiz())
+		{
+			$result = $item;
+		}
+		else
+		{
+			$result = new ESearchOperator();
```
MosheMaorKaltura: Since the NOT option repeat itself, and will repeat more, lets consolidate it into a smaller function - 
$query = createNegativeQuery(array $items,$operator=null)
And in your case you can call it -
$item =  createNegativeQuery(array($item));

MosheMaorKaltura:
reviewed

PR #9200, opened 2020-02-23
Orion-15.18.0-PLAT-10708 duration type support for query esearch adpater

ZurKaltura:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9200)

### DIFF
--- a/alpha/lib/enums/durationType.php
+++ b/alpha/lib/enums/durationType.php
@@ -5,8 +5,8 @@
  */
 interface durationType extends BaseEnum
 {
-	const ENTRY_DURATION_TYPE_NOTAVAILABLE = "notavailable";
-	const ENTRY_DURATION_TYPE_SHORT = "short";
-	const ENTRY_DURATION_TYPE_MEDIUM = "medium";
-	const ENTRY_DURATION_TYPE_LONG = "long";
+	const NOT_AVAILABLE = 'notavailable';
+	const SHORT = 'short';
+	const MEDIUM = 'medium';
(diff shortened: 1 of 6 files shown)


alpha/lib/enums/durationType.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * @package Core
+ * @subpackage model.enum
+ */
+interface durationType extends BaseEnum
+{
+	const ENTRY_DURATION_TYPE_NOTAVAILABLE = "notavailable";
```
MosheMaorKaltura: remove name prefix so the API will not change

alpha/lib/enums/durationType.php
```diff
@@ -0,0 +1,12 @@
...
+	const ENTRY_DURATION_TYPE_SHORT = "short";
```
MosheMaorKaltura: Use '

plugins/search/providers/elastic_search/lib/model/filters/ESearchEntryQueryFromFilter.php
```diff
@@ -213,6 +228,12 @@ public function createElasticQueryFromFilter(baseObjectFilter $filter)
 		$this->prepareEntriesCriteriaFilter($filter);
 		foreach($filter->fields as $field => $fieldValue)
 		{
+			if(in_array($field, static::getSpecialFields()))
```
MosheMaorKaltura: rename to self

PR #9217, opened 2020-03-04
PLAT-10732: fix a fatal error in auto archive

danielyag:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9217)

### DIFF
--- a/alpha/apps/kaltura/lib/batch2/kFlowHelper.php
+++ b/alpha/apps/kaltura/lib/batch2/kFlowHelper.php
@@ -3217,22 +3217,20 @@ class kFlowHelper
 		if (!$liveEntry){
 			throw new APIException(APIErrors::ENTRY_ID_NOT_FOUND, $liveEntryId);
 		}
-		else {
-			$recordStatus = $liveEntry->getRecordStatus();
-			$shouldAutoArchive = $liveEntry->getRecordingOptions() ?
-				$liveEntry->getRecordingOptions()->getShouldAutoArchive() : false;
-			if ($recordStatus == RecordStatus::PER_SESSION && $shouldAutoArchive == true) {
-				$liveEntryArchiveJobData = new kLiveEntryArchiveJobData();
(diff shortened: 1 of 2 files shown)


alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -3207,41 +3207,45 @@ public static function handleBulkDownloadRetried(BatchJob $dbBatchJob, kBulkDown
...
+	{
+		KalturaLog::info('Handling live to vod finished');
+		$liveEntryId = $data->getLiveEntryId();
+		$vodEntryId = $data->getVodEntryId();
+		$liveEntry = entryPeer::retrieveByPK($liveEntryId);
+		/** @var LiveStreamEntry $liveEntry */
+		if (!$liveEntry){
+			throw new APIException(APIErrors::ENTRY_ID_NOT_FOUND, $liveEntryId);
+		}
+		else {
```
noam-arad: no need for this else, throw is like return and will stop the running of the metho

batch/batches/Archive/KAsyncLiveEntryArchive.class.php
```diff
@@ -7,79 +7,89 @@
...
+	 */
+	protected function exec(KalturaBatchJob $job)
+	{
+		$jobData = $job->data;
+		/** @var KalturaLiveEntryArchiveJobData $jobData*/
+		$liveEntryId = $jobData->liveEntryId;
+		$liveEntry = KBatchBase::$kClient->baseEntry->get($liveEntryId);
+		/** @var KalturaLiveStreamEntry $liveEntry */
+		if (!$liveEntry){
+			KalturaLog::err("Live entry [" . $liveEntryId . "] not found");
```
noam-arad: if this is the case I think we should close the job with a different status than finished and a different error message
"Could not get live entry with id <live_entry_id>"
Status should be KalturaBatchJobStatus::FAILED

PR #9234, opened 2020-03-09
PLAT-10736 directory cleanup fix

gotlieb:
@MosheMaorKaltura

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9234)

### DIFF
--- a/batch/batches/Cleanup/KAsyncDirectoryCleanup.class.php
+++ b/batch/batches/Cleanup/KAsyncDirectoryCleanup.class.php
@@ -12,6 +12,8 @@
  */
 class KAsyncDirectoryCleanup extends KPeriodicWorker
 {
+	const CHUNK_ENCODEING_POSTFIX = 'chunkenc';
+
 	/* (non-PHPdoc)
 	 * @see KBatchBase::getType()
 	 */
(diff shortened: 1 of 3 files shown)


batch/batches/Cleanup/KAsyncDirectoryCleanup.class.php
```diff
@@ -12,6 +12,8 @@
  */
 class KAsyncDirectoryCleanup extends KPeriodicWorker
 {
+	const CHUNK_ENCODEING_POSTFIX = 'chunkenc';
```
MosheMaorKaltura: You should get it from  infra/chunkedEncode/KChunkedEncode.php 
By adding const to this class.

batch/batches/Cleanup/KAsyncDirectoryCleanup.class.php
```diff
@@ -25,64 +27,161 @@ public static function getType()
...
-		else
+	}
+
+	/**
+	 * @param $path
+	 * @param $now
+	 * @param $secondsOld
+	 * @return bool
+	 */
+	protected function shouldDeleteDirectory($path, $now, $secondsOld)
```
MosheMaorKaltura: Logic problem.
If dir is not chunk , this function returns always true.
The logic should be - always not, if it is chunk - check internal content and decide yes of no.

batch/batches/Cleanup/KAsyncDirectoryCleanup.class.php
```diff
@@ -25,64 +27,161 @@ public static function getType()
...
+				}
+				else
+				{
+					continue;
+				}
+			}
+			else
+			{
+				$this->deleteFile($file, $usePHP);
+			}
```
MosheMaorKaltura: for statistics and improvement - must add debug print here the states how many files were deleted.

PR #9259, opened 2020-03-18
PLAT-10735: add periodic storage export batch

inbal-ben-david:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9259)


api_v3/services/StorageProfileService.php
```diff
@@ -130,5 +138,255 @@ public function listAction(KalturaStorageProfileFilter $filter = null, KalturaFi
...
+		// Note: disabling the criteria because it accumulates more and more criterions, and the status was already explicitly added
+		//		once that bug is fixed, this can be removed
+		FileSyncPeer::setUseCriteriaFilter(false);
+		$fileSyncs = FileSyncPeer::doSelect($c, myDbHelper::getConnection(myDbHelper::DB_HELPER_CONN_PROPEL2));
+		FileSyncPeer::setUseCriteriaFilter(true);
+
+		return $fileSyncs;
+	}
+
+	protected static function updateLastId($fileSyncs, $idLimit)
```
MosheMaorKaltura: Since the last id is just returned here, no update is done to a member or static value - please rename the fucntion to getlastId

batch/batches/Storage/Periodic/KAsyncStoragePeriodicExport.class.php
```diff
@@ -0,0 +1,267 @@
...
+/**
+ * process remote storage files and export using file handlers as required
+ *
+ * @package Scheduler
+ * @subpackage Storage
+ */
+class KAsyncStoragePeriodicExport extends KPeriodicWorker
+{
+	const MAX_EXECUTION_TIME = 3600;        // can be exceeded by one file sync
+	const IDLE_SLEEP_INTERVAL = 10;
```
MosheMaorKaltura: This value also needs to come from the batch config and if not use this default

batch/batches/Storage/Periodic/KAsyncStoragePeriodicExport.class.php
```diff
@@ -0,0 +1,267 @@
...
+
+		while (time() < $timeLimit)
+		{
+			$storageProfile = $this->getStorageProfile();
+			if(!$storageProfile)
+			{
+				continue;
+			}
+
+			self::$kClient->setResponseProfile($responseProfile);
```
MosheMaorKaltura: This can be done once outside the loop

batch/batches/Storage/Periodic/KAsyncStoragePeriodicExport.class.php
```diff
@@ -0,0 +1,267 @@
...
+				$exported = $this->export($storageProfile, $fileSync);
+				if($exported)
+				{
+					$status = KalturaFileSyncStatus::READY;
+				}
+				else
+				{
+					$status = KalturaFileSyncStatus::ERROR;
+				}
+				$this->changeFileSyncStatus($fileSync, $status);
```
MosheMaorKaltura: Since there file sync status will happen many times - use multi request for optimisation

batch/batches/Storage/Periodic/KAsyncStoragePeriodicExport.class.php
```diff
@@ -0,0 +1,267 @@
...
+				$this->changeFileSyncStatus($fileSync, KalturaFileSyncStatus::ERROR);
+			}
+		}
+	}
+
+	protected function handleAppException($e, $fileSync)
+	{
+		if($e->getCode() == KalturaBatchJobAppErrors::FILE_ALREADY_EXISTS)
+		{
+			KalturaLog::debug("File already exists in remote storage");
```
MosheMaorKaltura: Use '

batch/batches/Storage/Periodic/KAsyncStoragePeriodicExport.class.php
```diff
@@ -0,0 +1,267 @@
...
+			$status = KalturaFileSyncStatus::READY;
+		}
+		else
+		{
+			KalturaLog::err("Could not export file. Error: [{$e->getMessage()}]");
```
MosheMaorKaltura: Will the file sync ID + path will be easy to understand from the logs, if not add them here.

batch/batches/Storage/Periodic/KAsyncStoragePeriodicExport.class.php
```diff
@@ -0,0 +1,267 @@
...
+		$exportResult = $engine->export();
+		if(!$exportResult)
+		{
+			KalturaLog::debug('File export failed');
+		}
+		return $exportResult;
+	}
+
+
+	protected function createExportData($storageProfile, $fileSync)
```
MosheMaorKaltura: This priodic exported must not have anything dedicated to S3 or any other storage system. Try move the specific type handling to different file, most likely the engine code is a good place.

batch/batches/Storage/Periodic/KAsyncStoragePeriodicExport.class.php
inbal-ben-david: we cant extract it outside as there are other actions that use different response profile after we call this action so in the next iteration we must set the relevant response profile for the lock command

PR #9276, opened 2020-03-23
SUP-21111 handle syndication feed according to syndication confmap

gotlieb:
@MosheMaorKaltura

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9276)

### DIFF
--- a/api_v3/lib/syndication/KalturaSyndicationFeedRenderer.php
+++ b/api_v3/lib/syndication/KalturaSyndicationFeedRenderer.php
@@ -17,7 +17,10 @@ class KalturaSyndicationFeedRenderer
 	const PAGE_SIZE_MAX_VALUE = 500;
 
 	const DYNAMIC_TOTAL_RESULTS = 200;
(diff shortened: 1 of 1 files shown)


api_v3/lib/syndication/KalturaSyndicationFeedRenderer.php
```diff
@@ -408,6 +408,18 @@ public function getNextEntryByPlaylistType()
 		}
 	}
 
+	protected function shouldUseDynamicPlaylist()
+	{
+		$syndicationMap = kConf::getMap('syndication');
```
MosheMaorKaltura: 1. Better practice - 
kConf::get('playlistExecutionLimitPartners','syndication',array());
And then just check if the partner is in this array.

2. Add the map name + section name  as const to syndycation class

gotlieb: done

PR #9291, opened 2020-03-25
PLAT-10758

gotlieb:
@MosheMaorKaltura

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9291)


plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -559,7 +559,10 @@ protected function cuePointDeleted(CuePoint $cuePoint)
 		$c = new Criteria();
```
MosheMaorKaltura: All in the if

PR #9292, opened 2020-03-25
SUP-19377: fix for php 5

ravitshalem:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9292)

### DIFF
--- a/plugins/reach/lib/model/ReachProfile.php
+++ b/plugins/reach/lib/model/ReachProfile.php
@@ -491,8 +491,8 @@ class ReachProfile extends BaseReachProfile
 				}
 				else if (count($rule->getConditions()) == 1)
 				{
-					$condition = $rule->getConditions()[0];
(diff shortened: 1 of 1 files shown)


plugins/reach/lib/model/ReachProfile.php
```diff
@@ -483,12 +483,20 @@ public function fulfillsRules(kScope $scope, $checkEmptyRulesOnly = false)
...
+			if(!$checkEmptyRulesOnly)
 			{
-				continue;
+				if (!count($rule->getConditions()))
+				{
+					continue;
+				}
+				else if (count($rule->getConditions()) == 1)
+				{
+					$condition = $rule->getConditions()[0];
```
MosheMaorKaltura: This wont work in old php 
try - 
php -r "function a() {return array(1,2,33);}print a()[0];"

But you can do this
$condition[0]->getType();

MosheMaorKaltura:
reviewed

PR #9322, opened 2020-04-12
REACH-836 - export vendor catalog items to csv

gotlieb:
@MosheMaorKaltura

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9322)

### DIFF
--- a/plugins/reach/lib/kReachUtils.php
+++ b/plugins/reach/lib/kReachUtils.php
@@ -4,10 +4,18 @@
  */
 class kReachUtils
 {
-	static private $vendorCatalogItemStatusEnumTranslate = array(
+	static private $dateFields = array('createdAt', 'updatedAt');
+	static private $catalogItemTranslateableFields = array('status','serviceType','serviceFeature','turnAroundTime','outputFormat');
+
+	static private $statusEnumTranslate = array(
 		1 => "DEPRECATED",
 		2 => "ACTIVE",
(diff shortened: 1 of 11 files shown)


plugins/reach/lib/kReachUtils.php
```diff
@@ -177,4 +214,81 @@ public static function reachStrToTime($offset , $value)
...
+		return $headers;
+	}
+
+	/**
+	 * @param $csvData
+	 * @return array
+	 */
+	public static function validateAndTranslateCatalogItemCsvData($csvData)
+	{
+		if (isset($csvData['createdAt']))
```
MosheMaorKaltura, line 232: Also here - you can hold an array of string with the names of the fields you want, then just foreach them

plugins/reach/services/VendorCatalogItemService.php
```diff
@@ -167,4 +166,60 @@ public function deleteAction($id)
...
+	public function exportAction($vendorId = null)
+	{
+		$filter = new KalturaVendorCatalogItemFilter();
+		if($vendorId)
+		{
+			$filter->vendorPartnerIdEqual = $vendorId;
+		}
+
+		$pager = new KalturaFilterPager();
+		$pager->pageSize = 500;
```
MosheMaorKaltura, line 184: pageIndex must be init to 1, otherwise 0 and 1 are the same page

PR #9324, opened 2020-04-12
REACH2-835: take into account weekends and work hours

avichainoach:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9324)


plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -235,9 +240,83 @@ public function getServiceFeature()
 		return $this->getFromCustomData(self::CUSTOM_DATA_SERVICE_FEATURE);
 	}
 
-	/* (non-PHPdoc)
- 	 * @see BaseEntryVendorTask::preSave()
- 	 */
+	protected function calculateNextBusinessDay()
```
MosheMaorKaltura, line 243: This function should be generic - calculate next day from timestamp and not from now.
Add timestamp in the API call.

plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -235,9 +240,83 @@ public function getServiceFeature()
...
+
+		//Set to EST Time
+		date_default_timezone_set('America/New_York');
+		$currentDateTime = new DateTime("now");
+
+		//Calculate the next business day
+		$expTime = $this->calculateNextBusinessDay();
+
+		//Calculate the second business day
+		if ($numBusinessDays == 2)
```
MosheMaorKaltura, line 295: This function should handle also N generic business days and not just 1 / 2, what if we want to add 7 business days?

plugins/reach/lib/model/enums/VendorServiceTurnAroundTime.php
```diff
@@ -7,16 +7,16 @@ interface VendorServiceTurnAroundTime extends BaseEnum
 {
 	const BEST_EFFORT           = -1;
 	const IMMEDIATE             = 0;
+	const ONE_BUSINESS_DAY      = 1;
```
MosheMaorKaltura, line 10: Since the numeric value was number of seconds so far, in case of Business days better use value that cannot be used otherwise - '1B' , '2B' etc....

plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -235,9 +240,84 @@ public function getServiceFeature()
...
+
+		$currentDateTime = new DateTime("now");
+		$currentDay = date("N");
+		$currentTime = date("His");
+
+		//Calculate the next (1st) business day
+		$expTime = $this->calculateNextBusinessDay($currentDay, $currentTime);
+
+		//Calculate the other business days
+		if ($numBusinessDays > 1)
```
MosheMaorKaltura, line 295: Why not just do - 

while(numBusinessDays)
{
       $time = $this->calculateNextBusinessDay($day, $cme);
      //calculate the day
     numBusinessDays--
}

In the current code you are limited up to 5 business days.

plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -252,6 +332,11 @@ public function preSave(PropelPDO $con = null)
 				{
 					$turnAroundTime = self::SEVEN_DAYS;
 				}
+				elseif( ($turnAroundTime == VendorServiceTurnAroundTime::ONE_BUSINESS_DAY) ||
```
MosheMaorKaltura, line 335: Add up to 7 business days.
If the value is between 1 and max days, send the value , so we can support more days then  1 & 2

PR #9329, opened 2020-04-13
PLAT-10778 handle commas in kmcng free text search

gotlieb:
@MosheMaorKaltura

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9329)


plugins/search/providers/elastic_search/lib/model/filters/ESearchEntryQueryFromFilter.php
```diff
@@ -404,13 +404,26 @@ protected function createUnifiedSearchItem($value, $removeWildCard = false)
 			$freeTextSearchExact->setSearchTerm($value);
 			return $freeTextSearchExact;
 		}
-
+		$items = explode(',',$value);
```
MosheMaorKaltura, line 407: Add const PHRASE_SEPERATOR

PR #9351, opened 2020-04-21
Propus-16.1.0-PLAT-10795 enhancing interactivity schema validation

ZurKaltura:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9351)


plugins/interactivity/lib/model/validators/kInteractivityNodeValidator.php
```diff
@@ -7,12 +7,15 @@
 class kInteractivityNodeValidator extends kInteractivityBaseValidator
 {
 	const ID = 'id';
```
MosheMaorKaltura, line 9: Rename to NODE_ID or use ID for both

PR #9354, opened 2020-04-22
PLAT-10817: fix call to getEntryId on an entryId

inbal-ben-david:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9354)


moshemery:
reviewed

PR #9361, opened 2020-04-23
PLAT-10824: when choosing remote delivery profile always prefer perio…

inbal-ben-david:
…dic profile

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9361)

### DIFF
--- a/alpha/apps/kaltura/lib/storage/kFileSyncUtils.class.php
+++ b/alpha/apps/kaltura/lib/storage/kFileSyncUtils.class.php
@@ -908,49 +908,16 @@ class kFileSyncUtils implements kObjectChangedEventConsumer, kObjectAddedEventCo
 			}
 		}
 
-		$periodicStorageIds = array();
-		if($fetch_from_remote_if_no_local)
-		{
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/lib/storage/kFileSyncUtils.class.php
```diff
@@ -908,7 +908,14 @@ public static function getReadyFileSyncForKey ( FileSyncKey $key , $fetch_from_r
 			}
 		}
 
+		$periodicStorageIds = array();
+		if($fetch_from_remote_if_no_local)
+		{
+			$periodicStorageIds = kStorageExporter::getPeriodicStorageIdsByPartner($key->partner_id);
```
MosheMaorKaltura, line 914: At this point you should have only sort the list of file_sync_list where the local DC remains first un touched, and the non local should be order where the periodic is first.
file_sync_list = sortFileSyncList($file_sync_list,$priodicRemoteStorageIds)

All the same logic remains exactly the same

PR #9371, opened 2020-04-27
PLAT-10836:changing func getLineFromFileTail to get the last line wit…

ravitshalem:
…hout reading the entire file

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9371)

### DIFF
--- a/infra/storage/kFile.class.php
+++ b/infra/storage/kFile.class.php
@@ -563,30 +563,45 @@ class kFile extends kFileBase
 		return is_dir($path);
 	}
 
-	public static function getLineFromFileTail($file)
+	public static function getLinesFromFileTail($file, $numberOfLines = 1, $ignoreEmptyEnds = true)
 	{
+		$eol = self::getArrEndOfLine();
 		$f = fopen($file, 'r');
 		$index = -1;
 		fseek($f, $index, SEEK_END);
 		$char = fgetc($f);	//get the last char in the file
 
-		while ($char === "\n" || $char === "\r")
+		while ($ignoreEmptyEnds && in_array($char, $eol))
 		{
 			fseek($f, $index--, SEEK_END);
 			$char = fgetc($f);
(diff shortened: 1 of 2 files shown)


infra/storage/kFile.class.php
```diff
@@ -563,14 +563,28 @@ public static function checkIsDir($path)
 		return is_dir($path);
 	}
 
-	public static function getLineFromFileTail($file , $lineNum)
+	public static function getLineFromFileTail($file)
```
MosheMaorKaltura, line 566: rename - getLinesFromFileTail so it can be more generic
add param numberOfLines (default 1)
add param ignoreEmptyEnds = true //this flag indicate if we need ignore ending with new lines.

infra/storage/kFile.class.php
```diff
@@ -563,14 +563,28 @@ public static function checkIsDir($path)
...
 	{
-		$lines = file($file);
-		if ($lines && count($lines))
+		$f = fopen($file, 'r');
+		$index = -1;
+		fseek($f, $index, SEEK_END);
+		$char = fgetc($f);	//get the last char in the file
+
+		while ($char === "\n" || $char === "\r")
```
MosheMaorKaltura, line 573: this section should be used - only if we want to ignore ending with NEW LINEs

infra/storage/kFile.class.php
```diff
@@ -563,14 +563,28 @@ public static function checkIsDir($path)
...
 		{
-			return $lines[count($lines) - $lineNum];
+			fseek($f, $index--, SEEK_END);
+			$char = fgetc($f);
 		}
-		return $lines;
+
+		$line = '';
+		while ($char !== false && $char !== "\n" && $char !== "\r")
```
MosheMaorKaltura, line 580: This loop should accumulate number of requested lines.

PR #9377, opened 2020-04-28
SUP-21891: do not update entry since this can cause race condition be…

noam-arad:
…tween DCs

### DIFF
--- a/alpha/lib/model/LiveEntryServerNode.php
+++ b/alpha/lib/model/LiveEntryServerNode.php
@@ -39,11 +39,6 @@ class LiveEntryServerNode extends EntryServerNode
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


alpha/lib/model/LiveEntryServerNode.php
```diff
@@ -111,7 +110,13 @@ public function postDelete(PropelPDO $con = null)
 			/* @var $liveEntry LiveEntry */
 			$entryServerNodes = EntryServerNodePeer::retrievePlayableByEntryId($liveEntry->getId());
```
MosheMaorKaltura, line 111: Since the 2 files have duplicate code #111 -> #126, consider moving the code to a single function that will be called from the 2 places

PR #9380, opened 2020-04-30
PLAT-10838: add entry server node id to the m3u8 url as sessionId

Amirhalfon1:
- add entry server node id to the m3u8 url in case of MediaServerNode of type LiveCluster.
- change getHttpUrl function in DeliveryProfileLive.php to receive entryServerNode instead of serverNode

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9380)

### DIFF
--- a/alpha/lib/model/DeliveryProfileLive.php
+++ b/alpha/lib/model/DeliveryProfileLive.php
@@ -135,7 +135,7 @@ abstract class DeliveryProfileLive extends DeliveryProfile {
 		$streams = $liveEntryServerNode->getStreams();
 		$this->sanitizeAndFilterStreamIdsByBitrate($streams);
 		
-		$this->liveStreamConfig->setUrl($this->getHttpUrl($liveEntryServerNode->serverNode));
+		$this->liveStreamConfig->setUrl($this->getHttpUrl($liveEntryServerNode));
 		$this->liveStreamConfig->setPrimaryStreamInfo($liveEntryServerNode->getStreams());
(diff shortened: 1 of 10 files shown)


alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -354,7 +355,10 @@ protected function getLivePackagerUrl($serverNode, $streamFormat = null)
 			$entryId = $this->getDynamicAttributes()->getEntryId();
 		}
 		
-		$livePackagerUrl = "$livePackagerUrl/p/$partnerID/e/$entryId/" . $serverNode->getSegmentDurationUrlString($segmentDuration);
+		$livePackagerUrl = "$livePackagerUrl/p/$partnerID/e/$entryId/";
```
noam-arad, line 358: Is there a chance in the future that we'll want both segmentDurationUrlString and sessionIdUrlString on the same url?
Isn't it always one or the other? seems like there should be one call and the relevant serverNode will handle it accordingly.

david-winder-kaltura: You are correct the currently one is for the old live and the other for the new but can in the future be server that need them both? yes
we wonder if the wrap both as one function of get extraParams or something - then each server will call all his releated function - do you think that better?

noam-arad: Better to leave the legacy code untouched I guess.

PR #9395, opened 2020-05-05
SUP-21596: remove deleted entries from elastic results

amiras89:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9395)

### DIFF
--- a/plugins/search/providers/elastic_search/lib/model/search/kEntrySearch.php
+++ b/plugins/search/providers/elastic_search/lib/model/search/kEntrySearch.php
@@ -107,6 +107,17 @@ class kEntrySearch extends kBaseESearch
     public function fetchCoreObjectsByIds($ids)
     {
         $entries = entryPeer::retrieveByPKsNoFilter($ids);
+
+        // remove deleted entries from results (in case there is major lag - elastic may return IDs that are deleted in DB)
+		foreach ($entries as $key => $entry)
+		{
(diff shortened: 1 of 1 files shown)


amiras89:
Disclaimer:
We fetch 10 objects at a time
If one of them is deleted - we will fetch 9 results (while the request said pageSize=10)

plugins/search/providers/elastic_search/lib/model/search/kEntrySearch.php
```diff
@@ -107,6 +107,17 @@ public function getElasticTypeName()
     public function fetchCoreObjectsByIds($ids)
     {
         $entries = entryPeer::retrieveByPKsNoFilter($ids);
+
+        // remove deleted entries from results (in case there is major lag - elastic may return IDs that are deleted in DB)
+		foreach ($entries as $key => $entry)
```
MosheMaorKaltura, line 112: Fix indentation.
Verify that if you have related objects such as caption on entry , and it was removed - that the result will not break.

amiras89: Due to complexity and chance of breaking flow - we decided to write a script to fix it.

amiras89:
Due to complexity and chance of breaking backward flow - we decided to write a script to delete entries from elastic manually

PR #9397, opened 2020-05-06
PLAT-10865: exclude invalid flavors when checking if we should delete…

inbal-ben-david:
… local flavprs

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9397)


MosheMaorKaltura:
reviewed

PR #9398, opened 2020-05-06
Fix for cases where the uploaded chunk partially overrides content that was already concatenated to the output file

yossipapi:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9398)


alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -341,10 +341,17 @@ protected function handleResume($fileData, $resumeAt)
 		return $currentFileSize;
 	}
 	
-	private function appendCurrentChunk($targetFilePath, $chunkFilePath)
+	private function appendCurrentChunk($targetFilePath, $chunkFilePath, $resumeAt)
 	{
 		$targetFileResource = fopen($targetFilePath, 'r+b');
-		fseek($targetFileResource, 0, SEEK_END);
+		if($resumeAt == -1)
```
MosheMaorKaltura, line 347: If resume at == -1 is an application assumption of last chunk, than it should be defined in the class (using const or something)

PR #9414, opened 2020-05-13
PLAT-10890 - retrieve lastFileSyncId from DB in case he is not the ca…

gotlieb:
…che...
@MosheMaorKaltura

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9414)

### DIFF
--- a/api_v3/services/StorageProfileService.php
+++ b/api_v3/services/StorageProfileService.php
@@ -168,7 +168,7 @@ class StorageProfileService extends KalturaBaseService
 		$lastIdLoopAddition = self::getConfigVal($cloudStorageConfig, self::LAST_ID_LOOP_ADDITION, self::DEFAULT_LAST_ID_LOOP_ADDITION);
 		$maxIdDelay = self::getConfigVal($cloudStorageConfig, self::MAX_ID_DELAY, self::DEFAULT_MAX_ID_DELAY);
 		$maxId = self::getMaxId($keysCache, $storageProfileId, $maxIdDelay, $workerId);
-		$initialLastId = self::getLastFileSyncIdToLock($keysCache, $storageProfileId, $workerId);
+		$initialLastId = self::getLastFileSyncIdToLock($keysCache, $storageProfileId, $filter, $workerId);
 		$lastId = $initialLastId ? $initialLastId : $maxId;
 
 		// created at less than handled explicitly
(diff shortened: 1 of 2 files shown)


api_v3/services/StorageProfileService.php
```diff
@@ -410,4 +416,40 @@ protected static function getConfigVal($configMap, $configField, $defaultVal)
...
+	 * @param $storageProfileId
+	 * @return int
+	 * @throws PropelException
+	 */
+	protected static function getLastFileSyncIdToLock($keysCache, $storageProfileId, $workerId)
+	{
+		$initialLastId = $keysCache->get(self::LAST_FILESYNC_ID_PREFIX . $workerId);
+		if (!$initialLastId)
+		{
+			$fileSyncThresholdCreationTime = time() - kConf::get('lastFileSyncIdCreationTimeThreshold', 'cloud_storage', 0);
```
MosheMaorKaltura, line 432: This value should be relative to the createdAt time of the filter, in the calling function
Since the idea is that the filter will ask for last 6 hours , but since we have no last ID , and your gap is 2Hors (for exampe) , then you should try to find 1 FS from * hours till now

api_v3/services/StorageProfileService.php
```diff
@@ -410,4 +416,41 @@ protected static function getConfigVal($configMap, $configField, $defaultVal)
...
+		{
+			$fileSyncThresholdCreationTime = $filter->createdAtGreaterThanOrEqual - kConf::get('lastFileSyncIdCreationTimeThreshold', 'cloud_storage', 0);
+
+			$c = new Criteria();
+			$c->add(FileSyncPeer::STATUS, FileSync::FILE_SYNC_STATUS_PENDING);
+			$c->add(FileSyncPeer::FILE_TYPE, FileSync::FILE_SYNC_FILE_TYPE_URL);
+			$c->add(FileSyncPeer::DC, $storageProfileId, Criteria::IN);
+			$c->add(FileSyncPeer::LINKED_ID, NULL, Criteria::ISNULL);
+			$c->addAscendingOrderByColumn(FileSyncPeer::ID);
+			$c->add(FileSyncPeer::CREATED_AT, $fileSyncThresholdCreationTime, Criteria::GREATER_THAN);
```
MosheMaorKaltura, line 441: Use updated at as its an index

PR #9439, opened 2020-05-25
Propus-16.2.0-PLAT-10869-thumbnail action through new thumbnail API

ZurKaltura:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9439)

### DIFF
--- a/plugins/thumbnail/lib/model/thumbnailFactory/kBaseResizeAdapter.php
+++ b/plugins/thumbnail/lib/model/thumbnailFactory/kBaseResizeAdapter.php
@@ -131,7 +131,7 @@ class kBaseResizeAdapter
 		{
 			$entry = $this->getEntry();
 			$version = $this->parameters->get(kThumbFactoryFieldName::VERSION);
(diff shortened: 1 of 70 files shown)


plugins/thumbnail/lib/model/thumbnailFactory/kBaseResizeAdapter.php
```diff
@@ -136,6 +134,20 @@ protected function checkIfOldApiCachedExists($contentPath, $thumbDirs)
...
+				if($currPath != $this->finalThumbPath)
+				{
+					$moveFileSuccess = kFile::moveFile($currPath, $this->finalThumbPath);
+					if($moveFileSuccess)
+					{
+						return array (true, $this->finalThumbPath);
+					}
+					else
+					{
+						KalturaLog::Error("Failed to move thumbnail from [$currPath] to [$this->finalThumbPath], will return oldPath");
```
MosheMaorKaltura, line 146: In this case - 
1. Error is not requires as we do not really need the traceback nor the parameters 
2. Is it correct to return value rather then flase?

MosheMaorKaltura:
reviewed

PR #9456, opened 2020-05-31
PLAT-10943: periodic export - update lastId when there're no hits in …

avichainoach:
…range

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9456)

### DIFF
--- a/api_v3/services/StorageProfileService.php
+++ b/api_v3/services/StorageProfileService.php
@@ -193,17 +193,19 @@ class StorageProfileService extends KalturaBaseService
 			$idLimit = min($lastId + self::MAX_FILESYNC_ID_RANGE, $maxId);
 			$fileSyncs = FileSync::getFileSyncsChunkNoCriteria($baseCriteria, $lastId, $idLimit);
 
-			if (!$fileSyncs)
-			{
-				break;
-			}
-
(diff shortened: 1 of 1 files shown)


api_v3/services/StorageProfileService.php
```diff
@@ -193,17 +193,19 @@ function lockPendingFileSyncsAction(KalturaFileSyncFilter $filter, $workerId, $s
...
 			{
-				break;
+				$lastId = $idLimit;
 			}
-
-			if (count($fileSyncs) < KalturaFileSyncFilter::MAX_FILESYNCS_PER_CHUNK)
+			else
 			{
-				$done = true;
+				$lastId = end($fileSyncs)->getId();
```
MosheMaorKaltura, line 202: $done is useless - need to remove from the while()

avichainoach: There is one use in $done.
The query does not contain ''createdAtLessThanOrEqual".
We check that later in 'filterFileSyncs'.
if we find file syncs that are valid for this filter, we don't want to continue in the loop, but the lastID will be smaller than maxId.
Hence the $done.

MosheMaorKaltura:
reviewed

PR #9462, opened 2020-06-01
Fix Fatal: Call to undefined method kFileSyncUtils::getReadyKalturaInternalFileSyncsForKey()

yossipapi:
This method was removed as a part of larger refactor

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9462)


yossipapi:
Noi longer needed

PR #9471, opened 2020-06-03
Propus-16.4.0-Zoom-get-token-script

ZurKaltura:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9471)

### DIFF
--- a/alpha/scripts/utils/getZoomToken.php
+++ b/alpha/scripts/utils/getZoomToken.php
@@ -3,7 +3,7 @@
 if($argc != 2)
 {
 	echo "Arguments missing.\n\n";
-	echo "Usage: php getZoomToken.php {vendor account id}\n";
+	echo "Usage: php {$argv[0]} <vendor account id>\n";
 	exit;
 }
 
--- a/alpha/scripts/utils/getZoomToken.php
+++ b/alpha/scripts/utils/getZoomToken.php
@@ -1,0 +1,17 @@
+<?php
+
+if($argc != 2)
+{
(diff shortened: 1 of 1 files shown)


alpha/scripts/utils/getZoomToken.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+if($argc != 2)
+{
+	echo "Arguments missing.\n\n";
+	echo "Usage: php getZoomToken.php {vendor account id}\n";
```
MosheMaorKaltura, line 6: instead of writing the script name - use argv[0] so we can rename it if needed.

MosheMaorKaltura, line 6: When you want to specify fields in the usage you should use < description > and not { }

alpha/scripts/utils/getZoomToken.php
```diff
@@ -0,0 +1,17 @@
...
+	exit;
+}
+
+$vendorAccountId = $argv[1];
+
+require_once(__DIR__ . '/../bootstrap.php');
+
+$zoomIntegration = ZoomHelper::getZoomIntegrationByAccountId($vendorAccountId);
```
MosheMaorKaltura, line 14: If not object return - let's print appropriate message and stop the scipt.

ZurKaltura: VerifyZoomIntegration take care of that

MosheMaorKaltura:
reviewed

PR #9481, opened 2020-06-07
Propus 16.4.0 plat 10914

gotlieb:
@MosheMaorKaltura

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9481)


plugins/search/providers/elastic_search/lib/model/items/ESearchUnifiedItem.php
```diff
@@ -111,6 +112,24 @@ private static function addEntryFieldsToUnifiedQuery($eSearchUnifiedItem, &$entr
 
 	}
 
+	/**
+	 * @param $entryAllowedFields
+	 */
+	private static function handleExcludedGroups(&$entryAllowedFields)
```
MosheMaorKaltura, line 118: 1. use protected
2. Rename - why not call it unsetExcludedGroups

PR #9482, opened 2020-06-07
No-Plat: When destination entry not found delete temp entry so it wont be left as a zombie

yossipapi:


alpha/apps/kaltura/lib/clipconcat/kClipManager.php
```diff
@@ -548,7 +548,19 @@ private function concatDone(BatchJob $batchJob)
 		$concatAsset = assetPeer::retrieveById($concatJobData->getFlavorAssetId());
 		/** @var kClipConcatJobData $clipConcatJobData */
 		$clipConcatJobData = $batchJob->getRootJob()->getData();
-		$this->addDestinationEntryAsset($clipConcatJobData->getDestEntryId(), $concatAsset);
+		try {
```
MosheMaorKaltura, line 551: indentation

PR #9483, opened 2020-06-08
REACH2-678 do not retrieve captions for getPlaybackContext that has d…

gotlieb:
…isplayOnPlayer set to false
@MosheMaorKaltura

---
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9483)


MosheMaorKaltura:
reviewed

yossipapi:
@MosheMaorKaltura @gotlieb shouldn't we also order them in the following way:
1. The caption which is set as default should be the first in the list
2. group all captions with the same language to be sequential.

For example if I have the following 4 captions:
1. Spanish
2. english
3. Hebrew version 2
4. Hebrew version 1 (default)

The result should be like:
1. Hebrew version 1 (default)
2. Hebrew version 2
3. Spanish
4. English

What do you think?

PR #9490, opened 2020-06-09
SUP-22077: trim leading slashes from Request URI and privilege URI to avoid substr() compare mismatch

amiras89:
This change is [Reviewable](https://reviewable.io/reviews/kaltura/server/9490)


amiras89:
Suggested fix from server side
When a request with leading slashed comes in the URI, like:
//api_v3/service/report/action/getCsvFromStringParams?

Or if we'll get a privilege with a leading slashed in the uri, like:
urirestrict://api_v3/service/report/action/*

The comparison will fail

Also possible to fix from client side, check why KMS ended up with (SUP-22077):
https://www.kaltura.com//api_v3/service/report/action/getCsvFromStringParams?. . . .

PR #9523, opened 2020-06-22
PLAT-10997: export if delete flag is true but partner doesn't have additional storage

inbal-ben-david:


MosheMaorKaltura:
reviewed

PR #9530, opened 2020-06-25
KMS-20244: add codeCuePoint insert permission to capture

david-winder-kaltura:
Need also to add release note


MosheMaorKaltura:
reviewed

PR #9534, opened 2020-06-28
PLAT-10995:conversion logic - when the output file doesnt exist, will…

ravitshalem:
… monitor the progress of all the files with the same perfix

### DIFF
--- a/batch/batches/Convert/Engines/KJobConversionEngine.class.php
+++ b/batch/batches/Convert/Engines/KJobConversionEngine.class.php
@@ -163,10 +163,15 @@ abstract class KJobConversionEngine extends KConversionEngine
 		if (kFile::checkFileExists($this->inFilePath))
 		{
 			$newModificationTime = kFile::getFileLastUpdatedTime($this->inFilePath);
-			if ($newModificationTime !== false && $newModificationTime > $currentModificationTime)
-			{
-				return $newModificationTime;
-			}
+		}
+		else
+		{
+			$newModificationTime = kFileUtils::getMostRecentModificationTimeFromDir($this->inFilePath);
+		}
+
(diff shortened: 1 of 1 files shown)


batch/batches/Convert/Engines/KJobConversionEngine.class.php
```diff
@@ -238,7 +243,7 @@ protected function executeConversionCmdlineSmartTimeout($command, &$return_var,
 				}
 				$currentModificationTime = $newModificationTime;
 			}
-			sleep(1);
+			sleep(10);
```
MosheMaorKaltura, line 246: Use const!

batch/batches/Convert/Engines/KJobConversionEngine.class.php
```diff
@@ -163,10 +163,15 @@ protected function isConversionProgressing($currentModificationTime)
 		if (kFile::checkFileExists($this->inFilePath))
```
MosheMaorKaltura, line 163: THis will return true also on existence of a dir, better use checkIsDir also , to verify it is actual file and not a dir

batch/batches/Convert/Engines/KJobConversionEngine.class.php
```diff
@@ -270,7 +275,7 @@ protected static function extendExpiration($jobId, $maximumExecutionTime, $timeo
 
 	protected static function isReachedTimeout(&$timeout)
 	{
-		$timeout--;
+		$timeout = $timeout - 10;
```
MosheMaorKaltura, line 278: $timeout-=self::kkkconsttimeout

MosheMaorKaltura:
reviewed

PR #9552, opened 2020-07-05
SUP-20297: add REACH2 support for external_media of type YouTube

amiras89:


MosheMaorKaltura:
reviewed

plugins/reach/lib/kReachManager.php
```diff
@@ -184,6 +184,10 @@ public function shouldConsumeAddedEvent(BaseObject $object)
 			$this->buildingReachArrays($event, $event->getScope()->getPartnerId(), $event->getScope(), false);
 			return true;
 		}
+		if ($object instanceof entry && $object->getExternalSourceType() == KalturaExternalMediaSourceType::YOUTUBE)
```
yossipapi, line 187: I don't think this change is correct.
We already handle entries in https://github.com/kaltura/server/blob/Propus-16.6.0/plugins/reach/lib/kReachManager.php#L204

The only thing is that we handle only entires of type media_clip and this is maybe this is the condition youtube fail to answer.

So instead of adding a new consumer I think it would be better to add your case to the current one.

amiras89: Thanks @yossipapi will add it later and resolve this

amiras89: @yossipapi following our Teams chat - the flow above isn't being consumed when using external_media service which immediatlly set the entry status to Ready
plugins/content/external_media/services/ExternalMediaService.php:30

amiras89:
@MosheMaorKaltura / @yossipapi 

the current code will support REACH2 in 2 use cases:
1. Adding external_media via API (external_media->add)
2. Adding external_media via KMS (**I couldn't test KMS in my local env**)

But from the logs, KMS first add the entry and only then it updates the duration - therefore, added to objectChangedEvent consumer

Please review

PR #9567, opened 2020-07-08
SUP-20297: add REACH2 support for external_media of type YouTube

amiras89:


amiras89:
@MosheMaorKaltura @yossipapi 
getting the apiToCore value is long an ugly:
ExternalMediaPlugin::getExtendedTypes(entry::class, entryType::MEDIA_CLIP)

So I pushed it into a function + added the regular entryType::MEDIA_CLIP

That way, if in future we'll want to also check changedEvent - we have the function to return all the supported types

Awaiting your feedback
Thanks

PR #9581, opened 2020-07-13
Add new permission allowing admin KS to set userId on new cue point

hilak:


hilak:
@MosheMaorKaltura same code that was approved last sprint.

PR #9584, opened 2020-07-14
Plat-11051 Releasing lock in case we failed processing thumbnail

ZurKaltura:


plugins/thumbnail/lib/model/thumbnailFactory/kBaseResizeAdapter.php
```diff
@@ -194,17 +194,31 @@ public function resizeEntryImage($params)
...
+			$entry = $this->getEntry();
+			if ($entry->getType() == entryType::PLAYLIST)
+			{
+				myPlaylistUtils::updatePlaylistStatistics($entry->getPartnerId(), $entry);
+			}
+
+			$adapter = new kImageTransformationAdapter();
+			$imageTransformation = $adapter->getImageTransformation($this->parameters);
+
+			$imagick = $imageTransformation->execute();
```
MosheMaorKaltura, line 209: What if we spend time on execution , do we really want to allow user do it again and again?
Consider releasing the cache only if we failed before execution, if after - then don't!

ZurKaltura: As far as I understand the lock is for preventing to have multiple processes trying to take the same resource  so to prevent someone putting the link in their site and suddenly all their users trigger the same process, I think we can handle it if it just one process, we do lock it again it the old thumbnail code, this is why I added the release here so if we fail the execution we could still have the fallback to the old code

MosheMaorKaltura:
reviewed

PR #9589, opened 2020-07-15
SUP-21781: for VOD get single edge server for all flavors

noam-arad:


alpha/lib/model/DeliveryProfile.php
```diff
@@ -466,12 +466,12 @@ public function getCacheInvalidationKeys()
 	
 	public function applyFlavorsDomainPrefix(&$flavors)
 	{
+		$domainPrefix = $this->getDeliveryServerNodeUrl();
```
MosheMaorKaltura, line 469: reviewed

PR #9591, opened 2020-07-15
PLAT-11027 - compatcheck new alerts on new server errors

gotlieb:
@MosheMaorKaltura

### DIFF
--- a/alpha/scripts/compatCheck.php
+++ b/alpha/scripts/compatCheck.php
@@ -624,12 +624,17 @@ function compareResults($resultNew, $resultOld)
 
 	if (!$resultNew)
 	{
-		return array('failed to parse new XML');
+		return array('#ERROR - failed to parse new XML');
 	}
 
 	if (!$resultOld)
 	{
 		return array('failed to parse old XML');
+	}
+
+	if (isset($resultNew['result']) && isset($resultNew['result']['error']) && !isset($resultOld['result']['error']))
+	{
+		return array("#ERROR - in new result and not in old result");
 	}
 
 	return compareArrays($resultNew, $resultOld, "");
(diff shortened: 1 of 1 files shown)


alpha/scripts/compatCheck.php
```diff
@@ -624,14 +624,19 @@ function compareResults($resultNew, $resultOld)
 
 	if (!$resultNew)
 	{
-		return array('failed to parse new XML');
+		return array('#ERROR - failed to parse new XML');
```
MosheMaorKaltura, line 627: The alert is really needed only when the old is parse and the new fails.
Otherwise there are many api's that both now and old failed on parse

alpha/scripts/compatCheck.php
```diff
@@ -624,14 +624,19 @@ function compareResults($resultNew, $resultOld)
...
 	}
 
 	if (!$resultOld)
 	{
 		return array('failed to parse old XML');
 	}
 
+	if (isset($resultNew['result']) && isset($resultNew['result']['error']) && !isset($resultOld['result']['error']))
```
MosheMaorKaltura, line 635: No need to check isset(isset($resultNew['result']['error']))
If checking isset($resultNew['result']['error']) and isset($resultNew['result']['error']) is not set - no error or notice is raised.

alpha/scripts/compatCheck.php
```diff
@@ -766,6 +771,11 @@ function testAction($ipAddress, $fullActionName, $parsedParams, $uri, $postParam
 			return;
 		}
 
+		if( !trim($resultNew) && $resultOld )
```
MosheMaorKaltura, line 774: why not trim the old also, to make sure that we raise this error only when old is not empty and new is emptry

alpha/scripts/compatCheck.php
gotlieb: look a few lines above... this what happens.

PR #9601, opened 2020-07-20
PLAT-11057: Delete local file syncs siblings only if partner is dedic…

avichainoach:
…ated AWS


MosheMaorKaltura:
reviewed - also prepare to 16.6

PR #9608, opened 2020-07-22
PLAT-11057: remove the check for the existence of private storages up…

avichainoach:
…on export to periodic storage


MosheMaorKaltura:
reviewed

PR #9610, opened 2020-07-22
PLAT-11067 fixing thumbnail adapter with encrypted file

ZurKaltura:


plugins/thumbnail/lib/model/thumbnailFactory/kImageTransformationAdapter.php
```diff
@@ -420,7 +420,19 @@ protected function addSource($step)
 	{
 		if(kFile::checkFileExists($this->parameters->get(kThumbFactoryFieldName::ORIG_IMAGE_PATH)) && $this->parameters->get(kThumbFactoryFieldName::VID_SEC) === kThumbAdapterParameters::UNSET_PARAMETER)
 		{
-			$source = new kFileSource($this->parameters->get(kThumbFactoryFieldName::ORIG_IMAGE_PATH));
+			/* @var $file_sync FileSync */
```
MosheMaorKaltura, line 423: Since source now in brackets , define it before the brackets to avoid lint warnings.

PR #9611, opened 2020-07-23
PLAT-11013: create a valid PDF file upon quiz->serve

avichainoach:
A valid PDF is a one that begins with "%PDF-[version]".


MosheMaorKaltura:
reviewed

PR #9622, opened 2020-07-27
LIV-167: add session type to playManifest url

Amirhalfon1:
add session type to playManifest url (only in case of LiveClusterMediaServerNode)
remove session_id (sid) from playManifest url


MosheMaorKaltura:
reviewed

PR #9631, opened 2020-07-30
REACH2-911:reach job cleaner pending and processing stuck jobs

ravitshalem:


batch/batches/ReachJobCleaner/KAsyncReachJobCleaner.php
```diff
@@ -0,0 +1,156 @@
...
+		$filter->orderBy = KalturaEntryVendorTaskOrderBy::UPDATED_AT_ASC;
+
+		$pager = new KalturaFilterPager();
+		$pager->pageSize = self::MAX_PAGE_SIZE;
+		$pager->pageIndex = 1;
+
+		$lastUpdatedAt = 0;
+		$totalCount = 0;
+		do {
+			if ($lastUpdatedAt)
```
MosheMaorKaltura, line 67: the if is not needed, nothing will happen if we search for updatedat>=0

batch/batches/ReachJobCleaner/KAsyncReachJobCleaner.php
```diff
@@ -0,0 +1,156 @@
...
+			{
+				$filter->updatedAtGreaterThanOrEqual = $lastUpdatedAt;
+			}
+			try
+			{
+				$entryVendorTaskList = $this->reachClientPlugin->entryVendorTask->listAction($filter, $pager);
+				$tasksCount = count($entryVendorTaskList->objects);
```
MosheMaorKaltura, line 74: Since this function (count...) can't throw, take it out of the try catch

batch/batches/ReachJobCleaner/KAsyncReachJobCleaner.php
```diff
@@ -0,0 +1,156 @@
...
+			}
+			catch (Exception $e)
+			{
+				KalturaLog::debug("Couldn't list entry Vendor Tasks " . $e->getMessage());
```
MosheMaorKaltura, line 78: string that do not process, use single qoute

batch/batches/ReachJobCleaner/KAsyncReachJobCleaner.php
```diff
@@ -0,0 +1,156 @@
...
+			}
+
+			$this->updateStuckJobs($entryVendorTaskList);
+
+			$totalCount += $tasksCount;
+			KalturaLog::debug("Handled More tasks - $tasksCount totalCount - " . $totalCount);
+			if ($tasksCount)
+			{
+				$lastObject = end($entryVendorTaskList->objects);
+				$lastUpdatedAt = $lastObject->updatedAt + 1;
```
MosheMaorKaltura, line 89: If the last job ends to at 17:00 , and we have 10 jobs at the same second, we will loose them.
You need to have $lastUpdatedAt = $lastObject->updatedAt

batch/batches/ReachJobCleaner/KAsyncReachJobCleaner.php
```diff
@@ -0,0 +1,156 @@
...
+			}
+
+			unset($entryVendorTaskList);
+			if (function_exists('gc_collect_cycles')) // php 5.3 and above
+			{
+				gc_collect_cycles();
+			}
+		} while ($pager->pageSize == $tasksCount);
```
MosheMaorKaltura, line 97: Instead use tasksCount>0, since it is possible that 1 job was deleted or moved, while we asked for the second page from the first page, and we will get 499 jobs/.

batch/batches/ReachJobCleaner/KAsyncReachJobCleaner.php
```diff
@@ -0,0 +1,156 @@
...
+			{
+				KalturaLog::debug('Going to update EntryVendorTask id ' . $entryVendorTask->id . ' to status ERROR. '
+					. self::JOB_DUE_DATE_EXCEEDED . $this->overdueTimePercentage . '%');
+
+				$kEntryVendorTask = new KalturaEntryVendorTask();
+				$kEntryVendorTask->status = KalturaEntryVendorTaskStatus::ERROR;
+				$kEntryVendorTask->errDescription = self::JOB_DUE_DATE_EXCEEDED . $this->overdueTimePercentage . '%';
+				try
+				{
+					KBatchBase::impersonate($entryVendorTask->partnerId);
```
MosheMaorKaltura, line 117: No need to do impersonate & impersonate before every call since there are no calls to the API in between.
You can just call impersonate once at the beginning of the function and one at the end

batch/batches/ReachJobCleaner/KAsyncReachJobCleaner.php
ravitshalem: in the first round updatedAtGreaterThanOrEqual = time() - 14 days.
from the sec round the updatedAtGreaterThanOrEqual will be $lastUpdatedAt

batch/batches/ReachJobCleaner/KAsyncReachJobCleaner.php
```diff
@@ -0,0 +1,150 @@
...
+			$this->updateStuckJobs($entryVendorTaskList);
+
+			$pager->pageIndex++;
+			$totalCount = $totalCount - $pager->pageSize;
+			$pager->pageSize = min(self::MAX_PAGE_SIZE, $totalCount);
+			if ($pager->pageSize > 0)
+			{
+				try
+				{
+					$entryVendorTaskList = $this->reachClientPlugin->entryVendorTask->listAction($filter, $pager);
```
MosheMaorKaltura, line 86: If list size returned zero - break the flow

MosheMaorKaltura:
reviewed

MosheMaorKaltura:
reviewed

PR #9632, opened 2020-07-30
Propus 16.8.0 elastic populate fix

gotlieb:
@MosheMaorKaltura

### DIFF
--- a/plugins/search/providers/elastic_search/scripts/populateElasticFromLog.php
+++ b/plugins/search/providers/elastic_search/scripts/populateElasticFromLog.php
@@ -32,12 +32,6 @@ error_reporting(E_ALL);
 KalturaLog::setLogger(new KalturaStdoutLogger());
 
 $hostname = (isset($_SERVER["HOSTNAME"]) ? $_SERVER["HOSTNAME"] : gethostname());
-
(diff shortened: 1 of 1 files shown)


plugins/search/providers/elastic_search/scripts/populateElasticFromLog.php
```diff
@@ -32,15 +32,19 @@ function filter($i)
 KalturaLog::setLogger(new KalturaStdoutLogger());
 
 $hostname = (isset($_SERVER["HOSTNAME"]) ? $_SERVER["HOSTNAME"] : gethostname());
-$configFile = ROOT_DIR . "/configurations/elastic/populate/$hostname.ini";
-
-if (!file_exists($configFile))
+$config = kConf::get('elasticPopulateSettings', 'elastic_populate', array());
```
MosheMaorKaltura, line 35: No need to use new map, we already have elasticDynamicMap, just add another section there.

gotlieb: This is not a regular map....  this is a map in the hosts hierarchy by server name (e.g ny-#.ini , pa-#.ini)

PR #9636, opened 2020-07-30
Update restoreDeletedEntriesByEntryList.php

hilak:
Deleted entries acquire the DELETED moderation status upon deletion (in addition to the regular entryStatus::DELETED). The moderation status needs to be reverted for the entry to reappear in specific searches.


MosheMaorKaltura:
reviewed

PR #9641, opened 2020-08-02
LIV-197: add dynamically live limits (limit by admin tags)

Amirhalfon1:
add dynamically live limits (limit by admin tags) - taken from admin.ini
take default limit values from local.ini
checking limit not exceeding in liveStream->authenticate
split overagedLimit and limitByAdminTag to subclasses of PartnerConfigurationLimitSubForm


plugins/admin_console/system_partner/lib/api/KalturaSystemPartnerLiveAdminTagLimit.php
```diff
@@ -0,0 +1,55 @@
...
+ * @subpackage api.objects
+ */
+class KalturaSystemPartnerLiveAdminTagLimit extends KalturaSystemPartnerLimit
+{
+	/**
+	 * @var string
+	 */
+	public $adminTag;
+
+	public static function buildFrom($adminTag, $value)
```
david-winder-kaltura, line 13: I think better naming can be use here (as createLiveAdminTagLimit)

api_v3/services/LiveStreamService.php
```diff
@@ -724,4 +736,38 @@ public function duplicateTemplateEntry($conversionProfileId, $templateEntryId, $
...
 	}
 
+	/**
+	 * Check if entry contains adminTag
+	 *
+	 * @param LiveEntry $entry
+	 * @param string $adminTag
+	 * @return boolean
+	 */
+	private function entryContainsAdminTag(LiveEntry $entry, $adminTag)
```
david-winder-kaltura, line 746: This function better be on entry.php (as the argument is regular entry and not live one in particular)
so signature should be like: public bool isContainsAdminTag(string $adminTag)

plugins/admin_console/system_partner/lib/api/KalturaSystemPartnerLimitArray.php
```diff
@@ -15,11 +15,20 @@ public static function fromPartner(Partner $partner)
...
+			{
+				$adminTagLimits =  KalturaSystemPartnerLiveAdminTagLimit::getArrayFromPartner($partner);
+				foreach ($adminTagLimits as $adminTagLimit)
+				{
+					$arr[] = $adminTagLimit;
+				}
+			} 
+			else
+			{
+				$arr[] = KalturaSystemPartnerOveragedLimit::fromPartner($typeValue, $partner);
```
david-winder-kaltura, line 29: We can add function "getArrayFromPartner" on KalturaSystemPartnerOveragedLimit which wrap "fromPartner" and just insert to array - in this case the signature will be the same and the "KalturaSystemPartnerLimit" should have getArrayFromPartner as well and call the sub file according the type

admin_console/forms/PartnerConfigurationLimitSubForm.php
```diff
@@ -135,10 +97,15 @@ public function populateFromObject(Form_PartnerConfiguration $form, Kaltura_Clie
 	public function getObject($objectType, array $properties, $add_underscore = true, $include_empty_fields = false)
 	{
 		$object = null;
-		if($this->withOverage)
-			$object = new Kaltura_Client_SystemPartner_Type_SystemPartnerOveragedLimit();
-		else
+
+		if($objectType && $objectType != '')
```
david-winder-kaltura, line 101: $object = ($objectType && $objectType != '') ? new $objectType() : new Kaltura_Client_SystemPartner_Type_SystemPartnerLimit();

api_v3/services/LiveStreamService.php
david-winder-kaltura, line 746: This function can be removed

PR #9672, opened 2020-08-10
PLAT-11086: add EnforceDelivery for RemotePlaybackSources

avichainoach:


alpha/apps/kaltura/lib/kPlaybackContextDataHelper.php
```diff
@@ -397,6 +397,18 @@ private function constructRemotePlaybackSources(entry $dbEntry, kContextDataHelp
 		if (!count($this->remoteFlavorsByDc))
 			return;
 
+		$partner = $dbEntry->getPartner();
+		if ($partner->getEnforceDelivery())
+		{
+			$deliveryAttributes = DeliveryProfileDynamicAttributes::init(null, $dbEntry->getId(), null);
```
MosheMaorKaltura, line 403: if this line is called anyway, they just move its call before this block and remove the call in line 412

PR #9676, opened 2020-08-11
PLAT-11098 - show diff of conf maps in admin console before submittin…

gotlieb:
…g a change

@MosheMaorKaltura


MosheMaorKaltura:
reviewed

PR #9698, opened 2020-08-16
PLAT-11125: add support for fallback storage id in serverFlavor action

inbal-ben-david:


alpha/apps/kaltura/lib/myPlaylistUtils.class.php
```diff
@@ -1093,7 +1093,8 @@ protected static function getCaptionFilePath($captionAsset, &$localFilePath)
 	{
 		$captionFileSyncKey = $captionAsset->getSyncKey(asset::FILE_SYNC_ASSET_SUB_TYPE_ASSET);
 		$preferredStorageId = serveFlavorAction::getPreferredStorageProfileId();
-		list ($captionFileSync, $path) = kFileSyncUtils::getFileSyncAndPathForFlavor($captionFileSyncKey, $captionAsset, $preferredStorageId);
+		$fallbackStorageId = serveFlavorAction::getFallbackStorageProfileId();
```
MosheMaorKaltura, line 1096: If this parameter does not exist, then we need to have the other DC as preferred , for cases where the request came from PA vod

alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
```diff
@@ -366,6 +368,7 @@ public function execute()
 		$captionLanguages = $this->getRequestParameter('captions', '');
 		$this->pathOnly = $this->getRequestParameter('pathOnly', false);
 		self::$preferredStorageId = $this->getRequestParameter('preferredStorageId');
+		self::$fallbackStorageId = $this->getRequestParameter('fallbackStorageId');
```
MosheMaorKaltura, line 371: Must add default if param not exist here

alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
```diff
@@ -926,4 +929,9 @@ public static function getPreferredStorageProfileId()
 	{
 		return self::$preferredStorageId;
 	}
+
+	public static function getFallbackStorageProfileId()
```
MosheMaorKaltura, line 933: use this function in all places in this file, instead of going to the self::$fall..

inside this function we handle the case where the call did not contained the fallbackSt

alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
inbal-ben-david: getRequestParameter already have a default value of null in case the parm is not provided in the request

PR #9705, opened 2020-08-17
PLAT-11082:Adding new parameter on flavor actualFileSizeOnDisk

ravitshalem:


alpha/apps/kaltura/lib/storage/kFileSyncUtils.class.php
```diff
@@ -1163,6 +1168,16 @@ public static function createSyncFileForKey ( $rootPath, $filePath, FileSyncKey
 		return $currentDCFileSync;
 	}
 
+	public static function setActualFileSizeOnAsset($currentDCFileSync)
+	{
+		$asset = assetPeer::retrieveById($currentDCFileSync->getObjectId());
+		if ($asset && $asset->getType() == assetType::FLAVOR && $currentDCFileSync->getFileSize())
+		{
+			$asset->setActualFileSizeOnDisk((int) ($currentDCFileSync->getFileSize() / 1024));
```
MosheMaorKaltura, line 1176: Why do we need the size in kbytes, already have it  and its not accurate.
Lets leave the size as is, in bytes

alpha/apps/kaltura/lib/storage/kFileSyncUtils.class.php
```diff
@@ -1117,6 +1117,11 @@ public static function createSyncFileForKey ( $rootPath, $filePath, FileSyncKey
 		$currentDCFileSync->setIsDir($isDir);
 		$currentDCFileSync->save();
 
+		if ($currentDCFileSync->getObjectType() == FileSyncObjectType::ASSET && $currentDCFileSync->getObjectSubType() == asset::FILE_SYNC_ASSET_SUB_TYPE_ASSET)
+		{
+			self::setActualFileSizeOnAsset($currentDCFileSync);
```
MosheMaorKaltura, line 1122: No need to limit it only to sub type asset, lets do it for all assets.
For all asset , we have new property called sizeBytes, and it is the actual files on disk.

MosheMaorKaltura:
reviewed

PR #9708, opened 2020-08-17
PLAT-11134: return delivery profiles of cloud storage only if partner…

avichainoach:
… is dedicated AWS


MosheMaorKaltura:
reviewed

PR #9728, opened 2020-08-23
PLAT-11145 - playmanifest delivery profile for periodic storage

ZurKaltura:


ZurKaltura:
was merged as part of the merged from propus-16.8.5 to propus-16.9.0

PR #9758, opened 2020-08-27
LIV-216: detach live limits validation

Amirhalfon1:
validate limits of the current entry streaming, regardless of the other limit exceedings

### DIFF
--- a/api_v3/services/LiveStreamService.php
+++ b/api_v3/services/LiveStreamService.php
@@ -196,6 +196,7 @@ class LiveStreamService extends KalturaLiveEntryService
 		$maxPassthroughStreams = $liveEntryPartner->getMaxLiveStreamInputs();
 		KalturaLog::debug("Max Passthrough streams [$maxPassthroughStreams]");
 		$adminTagsLimits = $liveEntryPartner->getMaxConcurrentLiveByAdminTag();
+		KalturaLog::debug('Current AdminTags: [' . $liveEntry->getAdminTags() . '] AdminTag limits : [' . print_r($adminTagsLimits, true) . ']');
 		$isCloudTranscode = $this->isCloudTranscode($liveEntry->getConversionProfileId());
 		
 		// If the entry is limited by adminTag, no other limit will be checked
@@ -791,7 +792,6 @@ class LiveStreamService extends KalturaLiveEntryService
 	 */
 	protected function validateAdminTagLimits(LiveEntry $currentEntry, $liveEntries, $adminTagsLimits)
 	{
-		KalturaLog::debug('Current AdminTags: [' . $currentEntry->getAdminTags() . '] AdminTag limits : [' . print_r($adminTagsLimits, true) . ']');
 		$adminTagsCounters = $adminTagsLimits;
 		foreach($liveEntries as $entry)
 		{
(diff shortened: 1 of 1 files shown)


api_v3/services/LiveStreamService.php
```diff
@@ -196,9 +196,13 @@ private function validateMaxStreamsNotReached(LiveEntry $liveEntry)
 		$maxPassthroughStreams = $liveEntryPartner->getMaxLiveStreamInputs();
 		KalturaLog::debug("Max Passthrough streams [$maxPassthroughStreams]");
 		$adminTagsOrigLimits = $liveEntryPartner->getMaxConcurrentLiveByAdminTag();
-		KalturaLog::debug('Current AdminTags: [' . $liveEntry->getAdminTags() . '] AdminTag limits : [' . print_r($adminTagsOrigLimits, true) . ']');
-		$adminTagsCounters = $adminTagsOrigLimits;
+		$isCloudTranscode = $this->isCloudTranscode($liveEntry->getConversionProfileId());
 		
+		$isAdminTagLimited = $this->isAdminTagLimited($liveEntry, array_keys($adminTagsOrigLimits));
```
david-winder-kaltura, line 201: Can move that to the "if" as this bool is not use afterward.
And prefer not to have var and function with the same name as here

api_v3/services/LiveStreamService.php
```diff
@@ -196,9 +196,13 @@ private function validateMaxStreamsNotReached(LiveEntry $liveEntry)
...
+		if($isAdminTagLimited)
+		{
+			return $this->validateAdminTagLimits($liveEntry, $liveEntries, $adminTagsOrigLimits);
```
david-winder-kaltura, line 204: Can rename $adminTagsOrigLimits => $adminTagsLimits
As you don't copy to counter array in this function anymore

api_v3/services/LiveStreamService.php
```diff
@@ -196,9 +196,13 @@ private function validateMaxStreamsNotReached(LiveEntry $liveEntry)
 		$maxPassthroughStreams = $liveEntryPartner->getMaxLiveStreamInputs();
 		KalturaLog::debug("Max Passthrough streams [$maxPassthroughStreams]");
 		$adminTagsOrigLimits = $liveEntryPartner->getMaxConcurrentLiveByAdminTag();
-		KalturaLog::debug('Current AdminTags: [' . $liveEntry->getAdminTags() . '] AdminTag limits : [' . print_r($adminTagsOrigLimits, true) . ']');
-		$adminTagsCounters = $adminTagsOrigLimits;
+		$isCloudTranscode = $this->isCloudTranscode($liveEntry->getConversionProfileId());
 		
+		$isAdminTagLimited = $this->isAdminTagLimited($liveEntry, array_keys($adminTagsOrigLimits));
+		if($isAdminTagLimited)
+		{
```
david-winder-kaltura, line 203: better add comment: if has admin tag limit we don't going to check other limitation...

api_v3/services/LiveStreamService.php
```diff
@@ -760,4 +748,65 @@ protected function updateAdminTagsCounters(LiveEntry $entry, &$adminTagsCounters
...
+	 * Validating adminTag limits not reached.
+	 *
+	 * @param LiveEntry $currentEntry
+	 * @param array $liveEntries
+	 * @param array $adminTagsOrigLimits
+	 * @throws KalturaAPIException
+	 */
+	protected function validateAdminTagLimits(LiveEntry $currentEntry, $liveEntries, $adminTagsOrigLimits)
+	{
+		KalturaLog::debug('Current AdminTags: [' . $currentEntry->getAdminTags() . '] AdminTag limits : [' . print_r($adminTagsOrigLimits, true) . ']');
```
david-winder-kaltura, line 794: Better move this log to before the "if" that call this function - as if this function will not call but we think we should we never see the log

david-winder-kaltura (changes requested):

david-winder-kaltura (approved):

PR #9773, opened 2020-08-30
REACH2-948:rule with boolean event notification condition should not …

ravitshalem:
…be clone between different partners


MosheMaorKaltura:
reviewed

plugins/reach/admin/ReachProfileCloneAction.php
```diff
@@ -72,4 +74,17 @@ public function doAction(Zend_Controller_Action $action)
 		}
 		Infra_ClientHelper::unimpersonate();
 	}
+
+	protected function gotBooleanEventNotificationCondition($rule)
```
MosheMaorKaltura, line 78: if you refer it as self, declare static

PR #9778, opened 2020-08-31
PLAT-11153: always support cloud periodic file syncs for all partners

inbal-ben-david:


MosheMaorKaltura (commented):
Also - 

At  alpha/apps/kaltura/lib/kPlaybackContextDataHelper.php , Line ~420, if (in_array($dcId, $this->storageIds)) , should be replaced with list of localstorageIds

PR #9785, opened 2020-09-01
CLOUDMGT-770:Support windows machine to put file via http POST

gotlieb:
@MosheMaorKaltura 
@erankor

### DIFF
--- a/api_v3/services/BatchService.php
+++ b/api_v3/services/BatchService.php
@@ -18,6 +18,7 @@
  */
 class BatchService extends KalturaBatchService
 {
+    const DEFAULT_MAX_DATA_SIZE = 20000000;
 
 // --------------------------------- BulkUploadJob functions 	--------------------------------- //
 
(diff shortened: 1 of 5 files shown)


api_v3/services/BatchService.php
```diff
@@ -631,4 +632,45 @@ public function extendLockExpirationAction($jobId, $maxExecutionTime)
...
+		{
+			throw new KalturaAPIException(KalturaErrors::FILE_SIZE_EXCEEDED, $size);
+		}
+
+		KalturaLog::debug("Going to save file in path - $path");
+		if(kFile::checkFileExists($path))
+		{
+			throw new KalturaAPIException(KalturaErrors::FILE_ALREADY_EXISTS, $path);
+		}
+		$fileContent = file_get_contents($data['tmp_name']);
```
yossipapi, line 671: Why not call rename to move the file from apache tmp location to the destination?

batch/batches/Convert/KAsyncConvert.class.php
```diff
@@ -295,8 +295,16 @@ private function moveFile(KalturaBatchJob $job, KalturaConvertJobData $data)
 					$fileSize=KBatchBase::foldersize($data->destFileSyncLocalPath);
 				else
 					$fileSize = kFile::fileSize($data->destFileSyncLocalPath);
-			
-				kFile::moveFile($data->destFileSyncLocalPath, $sharedFile);
+
+				if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN')
```
yossipapi, line 299: Instead of checking if its windows why not add a config on the worker config to force it to go via this flow. This way we can re use it the future if needed without being dependent on if its windows or not.

PR #9787, opened 2020-09-02
cloudmgt-770 fixing file upload to share path in ptt conversion

ZurKaltura:


ZurKaltura:
We are reverting the change instead of fixing this flow

PR #9801, opened 2020-09-07
PSVAMB-15535 - Update DB regardless of cache

GElkayam:
in order to avoid loop.


GElkayam:
Hi @NadavHarnik ,
Can you please review and merge?

PR #9895, opened 2020-09-30
Propus 16.9.0 plat 11193

MosheMaorKaltura:


batch/batches/Import/KAsyncImport.class.php
```diff
@@ -88,13 +88,35 @@ private function curlExec($sourceUrl,$localPath,$resumeOffset=0)
 		return array($res,$responseStatusCode,$errorMessage,$errorNumber);
 	}
 
+
+	/*
+	 * Will take a single KalturaBatchJob and fetch the URL to the job's destFile
+	 */
+	protected function fetchFromS3(KalturaBatchJob $job, KalturaImportJobData $data)
+	{
+		$s3Mgr = s3Mgr::getInstance(kFileTransferMgrType::S3);
```
inbal-ben-david, line 97: better put the logic in try catch and move to error status in case we failed

PR #9915, opened 2020-10-05
PLAT-11196: support s3 files download using downloadAction

inbal-ben-david:


alpha/apps/kaltura/lib/request/infraRequestUtils.class.php
```diff
@@ -531,18 +531,23 @@ public static function getRequestParams()
...
-			fseek($fh, $range_from);
-			while($range_length > 0)
-			{
-				$content = fread($fh, min($chunk_size, $range_length));
-				echo $content;
-				$range_length -= $chunk_size;
-			}
-			fclose($fh);
+			KalturaLog::err("Could not open source file [$file_name] for read");
+			self::unregisterStreamWrappers();
```
MosheMaorKaltura, line 539: need to return here

PR #9927, opened 2020-10-06
VIRTC-166: simulive - isLive conditional cache

Amirhalfon1:
setting the conditional cache of isLive action according to the event scheduling


MosheMaorKaltura:
reviewed

PR #9930, opened 2020-10-07
VIRTC-166: simulive - change duration calculation

Amirhalfon1:
change duration calculation to be the min between the entry length the event length (in order to put EXT-X-ENDLIST on time)


david-winder-kaltura (approved):

MosheMaorKaltura:
approved

PR #9934, opened 2020-10-07
PLAT-11182: take batch version from dc config in case we add job with…

inbal-ben-david:
… opposite dc


MosheMaorKaltura:
reviewed

PR #9941, opened 2020-10-08
SUP-24198 and SUP-24358: support generating download url based on con…

inbal-ben-david:
…figuration for remote file syncs


MosheMaorKaltura:
reviewed

PR #9959, opened 2020-10-12
aws: When using s3 there is not need to create the directory tree + i…

yossipapi:
…n case final status is passed dont call handleS3Exception

### DIFF
--- a/infra/storage/shared_file_system_managers/kS3SharedFileSystemMgr.php
+++ b/infra/storage/shared_file_system_managers/kS3SharedFileSystemMgr.php
@@ -733,6 +733,7 @@ class kS3SharedFileSystemMgr extends kSharedFileSystemMgr
 			if ($success)
 				break;
 			
+			sleep(rand(2,5));
 			$retries--;
 		}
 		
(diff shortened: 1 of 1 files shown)


infra/storage/shared_file_system_managers/kS3SharedFileSystemMgr.php
```diff
@@ -743,7 +724,17 @@ protected function doMoveLocalToShared($src, $dest, $copy = false)
 			return false;
 		}
 		
-		list($success, $res) = $this->doPutFileHelper($dest, $fp, $params);
+		$retries = $this->retriesNum;
```
MosheMaorKaltura, line 727: Must add while ($retries>0)

To avoid getting  -1 by mistake

infra/storage/shared_file_system_managers/kS3SharedFileSystemMgr.php
```diff
@@ -229,6 +230,7 @@ private function doPutFileHelper($filePath , $fileContent, $params)
 		}
 		catch (Exception $e)
 		{
+			KalturaLog::debug("Failed to uploaded to s3, info with message: " . $e->getMessage());
```
MosheMaorKaltura, line 233: log level in this case needs to be WARNING

PR #9964, opened 2020-10-13
SUP-22605:do not copy parentEntryId to targetEntry in Cross Distribution

ravitshalem:


plugins/content_distribution/providers/cross_kaltura/lib/batch/CrossKalturaDistributionEngine.php
```diff
@@ -513,6 +513,8 @@ protected function transformEntry(KalturaBaseEntry $sourceEntry, $forUpdate = fa
 		if (!$targetEntry->endDate) { $targetEntry->endDate = null; }
 		if (!$targetEntry->referenceId) { $targetEntry->referenceId = null; }
 
+		if (isset($targetEntry->parentEntryId)) { $targetEntry->parentEntryId = null; }
```
MosheMaorKaltura, line 516: No need to check status - just nullify it,

PR #9966, opened 2020-10-13
PLAT-11207: Fix export flavorAsset when the file sync does not exist …

avichainoach:
…in current DC


alpha/lib/model/StorageProfile.php
```diff
@@ -382,12 +382,6 @@ public function isValidFileSync(FileSyncKey $key)
 			return false;
 		}
```
MosheMaorKaltura, line 383: if $local is not in use anymore than remove it - list($kalturaFileSync,) = kFileSyncUtils::getReadyFileSyncForKey($key, true, false);

PR #10004, opened 2020-10-19
VIRTC-490: add postEnd to event

david-winder-kaltura:


plugins/schedule/base/lib/model/LiveStreamScheduleEvent.php
```diff
@@ -107,4 +124,10 @@ public function getCalculatedStartTime()
 		return parent::getCalculatedStartTime() - $preStartTime;
 	}
 
+	public function getCalculatedEndTime()
+	{
+		$postEndTime = !is_null($this->getPostEndTime()) ? $this->getPostEndTime() : 0;
```
MosheMaorKaltura, line 129: All the test is  useless , you can just make sure that getPostEndTime return default value of 0

MosheMaorKaltura:
reviewed

PR #10021, opened 2020-10-21
Super critical for conversion jobs

jessp01:
Without this, since it's the batch daemon (running as the `kaltura` user) that creates the dir and the API (so basically the Apache user) that does the move and deletion operations, things will fail.


jessp01:
@yossipapi, @gotlieb - I need this merged so I can release 16.10.0. 16.9.0 has the same problem.

PR #10067, opened 2020-11-01
aws: shared dc file syncs should also served as remote DC

yossipapi:


alpha/apps/kaltura/lib/storage/kFileSyncUtils.class.php
```diff
@@ -2113,7 +2113,7 @@ public static function getPathAndSourceTypeByFileSync($fileSync, $preferredStora
 		$forceRemoteServePattern = kConf::get('force_remote_serve_pattern', 'local', '');
 
 		// handle remote dc
-		if(!in_array($fileSync->getDc(), kDataCenterMgr::getDcIds()))
+		if(!in_array($fileSync->getDc(), kDataCenterMgr::getDcIds(false)))
```
MosheMaorKaltura, line 2116: Instead we can just just use $fileSync->getRemotePath()

PR #10073, opened 2020-11-02
aws: Use legacy cloud storage id when pushing src to remote and conve…

yossipapi:
…rting from remote


MosheMaorKaltura:
reviewed

PR #10086, opened 2020-11-03
Reach: Allow vendor accounts to fetch catalog item data as part of ge…

yossipapi:
…t jobs when using response profile


plugins/reach/lib/api/data/vendorCatalogItem/filters/KalturaVendorCatalogItemFilter.php
```diff
@@ -40,7 +40,8 @@ public function doGetListResponse(KalturaFilterPager $pager, KalturaDetachedResp
 		{
 			$partnerIdEqual = $this->partnerIdEqual;
 		}
-		elseif (!$this->partnerIdEqual && kCurrentContext::$ks_partner_id != Partner::ADMIN_CONSOLE_PARTNER_ID)
+		// Dont filter on partner if requesting partner id is admin console or has the vendor permission
+		elseif (!$this->partnerIdEqual && (kCurrentContext::$ks_partner_id != Partner::ADMIN_CONSOLE_PARTNER_ID || kCurrentContext::$action == "getJobs"))
```
MosheMaorKaltura, line 44: php string compare ===
string without processing use '

PR #10087, opened 2020-11-03
PLAT-11257: if admin user doesnt exists in account we try to login to with otp throw an error

inbal-ben-david:


MosheMaorKaltura:
reviewed

gonenradai:
in case we have such users in the system (that changed their email) - this will prevent them from login...
wouldn't a better solution be to search for the correct kuser using login_data_id?

PR #10091, opened 2020-11-04
PLAT-11257: if admin user doesnt exists in account we try to login to…

inbal-ben-david:
… with otp throw an error


MosheMaorKaltura:
reviewed

PR #10114, opened 2020-11-08
Propus-16.11.0-PLAT-11190-HF creating the folder before the packager …

ZurKaltura:
…try to save the image to it


ZurKaltura:
Will be separated into 2 PR

PR #10139, opened 2020-11-11
FEC-10567: support bundle builder server new response

dan-ziv:


alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -106,10 +106,19 @@ public static function buildBundleLocked($context)
...
-		{
-			KExternalErrors::dieError(KExternalErrors::BUNDLE_CREATION_FAILED, $config . " bundle created with wrong content");
-		}
+
+        if (isset($content['status'])) {
+            if ($content['status'] != 0) {
+                $message = $content['message'];
+                KExternalErrors::dieError(KExternalErrors::BUNDLE_CREATION_FAILED, $config . ". " . $message);
+            } else {
+                $content = $content['payload'];
```
OrenMe, line 115: do we need to check here for `bundle` field?

OrenMe (changes requested):
please check indentation

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
dan-ziv: don't think so, if the status is 0 (OK) so the bundle field should be there.
before we needed to check because we didn't knew what type of response we will get.

OrenMe: ok, sounds good.
Please fix the indentation.

PR #10149, opened 2020-11-15
PLAT-11239: Add download from directory support to download action using serveFlavor url

inbal-ben-david:


MosheMaorKaltura:
reviewed

PR #10156, opened 2020-11-15
SUP-24970: get max_pages_to_scan value from config, default is 7

amiras89:


amiras89:
FYI @MosheMaorKaltura

MosheMaorKaltura:
reviewed

PR #10160, opened 2020-11-16
PLAT-11221: add option to delete the recorded live entry automatically once deleting live entry

inbal-ben-david:


MosheMaorKaltura:
reviewed

PR #10167, opened 2020-11-16
VIRTC-920: Support manual live status

david-winder-kaltura:


Amirhalfon1 (approved):

alpha/lib/model/LiveEntry.php
```diff
@@ -499,41 +499,133 @@ public function getScheduleEvents($startTime, $endTime)
...
 	/**
 	 * @return boolean
 	 */
-	public function isCurrentlyLive($currentDcOnly = false)
+	public function isCurrentlyLive($currentDcOnly = false, $protocol = null)
+	{
+		return $this->getLiveStatus(true, $protocol) === EntryServerNodeStatus::PLAYABLE;
+	}
+
+	private function isExternalCurrentlyLive($reqProtocol = null)
```
MosheMaorKaltura, line 528: change private to protected (coding standard)

alpha/lib/model/LiveEntry.php
```diff
@@ -499,41 +499,133 @@ public function getScheduleEvents($startTime, $endTime)
...
 		{
-			return true;
+			$isProtocolLive = $this->checkIsLiveByProtocol($protocol);
+			if ($isProtocolLive !== null)
+			{
+				if ($isProtocolLive || $takeFirst)
+				{
+					return $isProtocolLive;
+				}
+				$isLive = $isProtocolLive; // as false
```
MosheMaorKaltura, line 563: should we break the loop once $isLive is set?

alpha/lib/model/LiveEntry.php
```diff
@@ -499,41 +499,133 @@ public function getScheduleEvents($startTime, $endTime)
...
 
-		$liveEntryServerNodes = $this->getPlayableEntryServerNodes();
-		if(!count($liveEntryServerNodes))
-			return false;
+		return $isLive;
+	}
 
-		/* @var LiveEntryServerNode $liveEntryServerNode*/
-		foreach($liveEntryServerNodes as $liveEntryServerNode)
+	private function checkIsLiveByProtocol($protocol) {
```
MosheMaorKaltura, line 570: protected

alpha/lib/model/LiveEntry.php
```diff
@@ -499,41 +499,133 @@ public function getScheduleEvents($startTime, $endTime)
...
-				if ($this->getExplicitLive() && !$this->canViewExplicitLive() && !$liveEntryServerNode->getIsPlayableUser())
-					return false;
 				return true;
 			}
+			$urlManagerBackup = DeliveryProfilePeer::getLiveDeliveryProfileByHostName(parse_url($backupUrl, PHP_URL_HOST), $dpda);
+			return $urlManagerBackup ? $urlManagerBackup->isLive($backupUrl) : false;
 		}
-		
-		return !$currentDcOnly;
+		return null;
```
MosheMaorKaltura, line 590: if function is called checkIsLive, why not return true or false?

alpha/lib/model/LiveEntry.php
```diff
@@ -499,41 +499,133 @@ public function getScheduleEvents($startTime, $endTime)
...
+			{
+				$maxKey = max(array_search($status, $statusOrder), array_search($entryServerNode->getStatus(), $statusOrder));
+				$status = $statusOrder[$maxKey];
+			}
+			//TODO case where on other dc has the explicit protected node we were return true even it is not viewable
+		}
+		return $status;
+	}
+
+	private function shouldConsiderEntryServerNodeStatusForLiveStatus($entryServerNode, $checkExplicitLive)
```
MosheMaorKaltura, line 617: protected

alpha/lib/model/LiveEntry.php
```diff
@@ -499,41 +499,133 @@ public function getScheduleEvents($startTime, $endTime)
...
+	{
+		if (!$entryServerNode || !($entryServerNode instanceof LiveEntryServerNode))
```
MosheMaorKaltura, line 619: use brackets of if
if(bla)
{
  //do
}

alpha/lib/model/LiveEntry.php
david-winder-kaltura: No - as is isLive is false by one protocol we want to keep check all other optional protocol (that the behavior we had till now)

alpha/lib/model/LiveEntry.php
david-winder-kaltura: NULL is mark the the protocol is not defined - because in that case we want to know when no protocol is set - the behavior is difference between protocol is defined but not live and when the requested protocol is not defined

PR #10168, opened 2020-11-16
No-Plat: Remove tmp files during convert once they are no longer needed

yossipapi:


infra/chunkedEncode/KChunkedEncode.php
```diff
@@ -1046,7 +1046,8 @@ public function getSessionName($mode="merged")
 		{
 			switch($mode){
 			case "merged":
-				$name = $this->setup->output."_merged";
+				$name = $this->params->output;
```
MosheMaorKaltura, line 1049: was the usage of params intent?

yossipapi: yes, it's the params saved on the class object

PR #10177, opened 2020-11-17
PLAT-22454: add url restriction

inbal-ben-david:


plugins/content/caption/base/services/CaptionAssetService.php
```diff
@@ -289,7 +289,7 @@ protected function attachUrl(CaptionAsset $captionAsset, $url)
 	{
 		$destPath = md5($url);
 		$fullPath = myContentStorage::getFSUploadsPath() . '/' . $destPath;
-		if (KCurlWrapper::getDataFromFile($url, $fullPath))
+		if (KCurlWrapper::getDataFromFile($url, $fullPath) && !myUploadUtils::isFileTypeRestricted($fullPath))
```
MosheMaorKaltura, line 292: This function has just been delete from this service and the captions imported from URL will be added by import Job (which you covered).

PR #10181, opened 2020-11-18
REACH2-973-Multiple zoom entries to have 1 parent entry

ZurKaltura:


plugins/vendor/zoom/lib/model/processors/kZoomRecordingProcessor.php
```diff
@@ -10,6 +10,11 @@ abstract class kZoomRecordingProcessor extends kZoomProcessor
 
 	protected $zoomClient;
 
+	/**
+	 * @var entry
+	 */
+	protected $mainEntry;
```
MosheMaorKaltura, line 16: Clear the value in the constructor

PR #10200, opened 2020-11-23
SUP-24970: get maxPagesToScan value from worker params config, default is 7

amiras89:
FYI @MosheMaorKaltura moved to workers config


MosheMaorKaltura:
reviewed

PR #10210, opened 2020-11-25
PLAT-22483:new script that will update tag on entries according to re…

ravitshalem:
…ference id

### DIFF
--- a/alpha/scripts/utils/updateTagAccordingToRefenceId.php
+++ b/alpha/scripts/utils/updateTagAccordingToRefenceId.php
@@ -28,11 +28,21 @@ foreach ($entryIds as $entryId)
 			list($referenceId, $tag) = explode(',', $referenceIdsTag);
 			if ($referenceId && $tag && $entryReferenceId === $referenceId)
 			{
-				$entryTags = $entry->getTags();
-				$entry->setTags($entryTags . ',__' . $tag);
-				$entry->save();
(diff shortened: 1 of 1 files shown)


alpha/scripts/utils/updateTagAccordingToRefenceId.php
```diff
@@ -0,0 +1,41 @@
...
+	{
+		$entryReferenceId = $entry->getReferenceID();
+		foreach ($referenceIdsTagList as $referenceIdsTag)
+		{
+			$referenceIdsTag = trim($referenceIdsTag);
+			list($referenceId, $tag) = explode(',', $referenceIdsTag);
+			if ($referenceId && $tag && $entryReferenceId === $referenceId)
+			{
+				$entryTags = $entry->getTags();
+				$entry->setTags($entryTags . ',__' . $tag);
```
assafb, line 32: @ravitshalem  
1. We want to add the new tag to the existing tags so just making sure, setTags doesn't override existing tags, right? it just adding the new one?
2. please use these lines to convert the Content Group value into tag: (to be consistent with how we handle other "hidden" tags in KMS)
$pattern = array('/^/', '/[^[:alnum:]]/u');
$replcement = array('__', '_');
$tag = preg_replace($pattern, $replcement, $tag);

assafb (changes requested):

etameran:
Plz read the translation array once outside the loop into a dictionary array  and then in the foreach on the entry just check if the reference is in your array

PR #10211, opened 2020-11-26
PLAT-22486: bumper- ignore access control and show sources when main …

avichainoach:
…entry has sources as well


MosheMaorKaltura:
reviewed

PR #10241, opened 2020-12-02
SUP-25292-Zoom-different records in same event

ZurKaltura:
Split handling the record file by their starting time


MosheMaorKaltura:
reviewed

PR #10251, opened 2020-12-06
PLAT-22458 - Add update script to align non SaaS

GElkayam:


GElkayam:
Deploys changes from https://github.com/kaltura/server/pull/10179

PR #10265, opened 2020-12-09
SUP-25292: if parentEntryId is null - do not enter this function

amiras89:


plugins/vendor/zoom/lib/model/processors/kZoomRecordingProcessor.php
```diff
@@ -75,6 +75,7 @@ public function handleRecordingVideoComplete($event)
 		$extraUsers = $this->getAdditionalUsers($recording->id, $dbUser->getPuserId());
 		foreach ($recording->recordingFiles as $recordingFilesPerTimeSlot)
 		{
+			$this->mainEntry = null;
```
amiras89, line 78: @MosheMaorKaltura @ZurKaltura 
Is this one related to the zoom 'parent-chiled' feature we removed

I'm not sure it should be here.

MosheMaorKaltura:
please close

PR #10284, opened 2020-12-14
PLAT-22519: add live entry custom data on clone

nadav-ravid:
LiveEntry custom data fields now get copied when cloning an entry
Small consistency change for live entry dvr fields


david-winder-kaltura (approved):

nadav-ravid:
This is a fix for specific paramaters in the live entry type . A wider solution should be implemented when possible to prevent similar problems from occurring with other objects or paramaters

PR #10288, opened 2020-12-15
FOUN-67: Handle shared storage dir serving as if they were local

yossipapi:


alpha/lib/model/asset.php
```diff
@@ -545,7 +545,10 @@ public function getDownloadUrl($useCdn = false, $forceProxy = false, $preview =
 		$syncKey = $this->getSyncKey(self::FILE_SYNC_ASSET_SUB_TYPE_ASSET);
 		list($fileSync, $serveRemote) = kFileSyncUtils::getFileSyncByStoragePriority($this->getPartnerId(), $syncKey, true);
 		
-		if($serveRemote && $fileSync && !in_array($fileSync->getDc(), kDataCenterMgr::getSharedStorageProfileIds()))
+		//Shared storage file syncs should not be handled as remote
+		$serveRemote = $serveRemote && !in_array($fileSync->getDc(), kDataCenterMgr::getSharedStorageProfileIds());
```
inbal-ben-david, line 549: i think the original condition was more correct as you access $fileSync->getDc but here only in line 551 you check if the object is not null instead of having it before

alpha/apps/kaltura/modules/extwidget/actions/downloadAction.class.php
```diff
@@ -226,14 +215,9 @@ private function dumpFile($file_path, $file_name, $limit_file_size = 0, $key = n
...
-			$mime_type = null;
-			if(!$allowRemote)
-			{
-				$mime_type = kFile::mimeType($file_path);
-			}
-
-			kFileUtils::dumpFile($file_path, $mime_type, null, $limit_file_size, $key, $iv, $fileSize, $allowRemote, $fileExt);
+			
+			$mime_type = kFile::mimeType($file_path);
+			kFileUtils::dumpFile($file_path, $mime_type, null, $limit_file_size, $key, $iv, $fileSize, false, $fileExt);
```
inbal-ben-david, line 220: if all the references to kFileUtils::dumpFile are now with allowRemote=false maybe worth removing it at all from the function if there is no use for it anymore

alpha/lib/model/asset.php
yossipapi: Done.

PR #10307, opened 2020-12-21
REACH2-981:Dont submit job with new version if an older version exist…

ravitshalem:
… with pending/processing mode


MosheMaorKaltura:
reviewed

PR #10334, opened 2020-12-28
Propus 16.14.0 plat 22548

ZurKaltura:
Changing the code so getFileSyncServeFlavorFields will work as it used to on a non-cloud environment


MosheMaorKaltura:
reviewed

PR #10339, opened 2020-12-30
FOUN-72 - Support writing & reading csv export files directly to & fom 3s

gotlieb:
@yossipapi


alpha/apps/kaltura/lib/batch2/kJobsManager.php
```diff
@@ -1976,7 +1976,16 @@ public static function addExportCsvJob(kExportCsvJobData $jobData, $partnerId, $
 	{
 		$batchJob = new BatchJob();
 		$batchJob->setPartnerId($partnerId);
-		
+		$partner = PartnerPeer::retrieveByPK($partnerId);
```
yossipapi, line 1979: I see this logic is repeating in the code so we should probably create something like getPartnerSharedStoargeBaseDir in myCloudUtils and implement the fetching of $storageProfile->getStorageBaseDir() there.

alpha/apps/kaltura/lib/batch2/model/kExportCsvJobData.php
```diff
@@ -25,7 +25,11 @@ class kExportCsvJobData extends kJobData
 	 * @var string
 	 */
 	private $userMail;
-	
+
+	/**
+	 * @var string
+	 */
+	public $storageDestinationFilePath;
```
yossipapi, line 32: better rename the sharedOutputPath

api_v3/services/ExportCsvService.php
```diff
@@ -94,13 +94,26 @@ public static function generateCsvPath($id, $ks)
...
-		
-		$partner_id = kCurrentContext::getCurrentPartnerId();
-		$folderPath = "/content/exportcsv/$partner_id";
-		$fullPath = myContentStorage::getFSContentRootPath() . $folderPath;
-		$file_path = "$fullPath/$id";
-		
-		return $file_path;
+
+		$partnerId = kCurrentContext::getCurrentPartnerId();
+		$folderPath = "/content/exportcsv/$partnerId";
```
yossipapi, line 99: Try to make this a bit more simple
The only diff here should be how you create the folderPath it will always be 
/content/exportcsv/
and then the diff is if the inner folder name is partnerId or the Scattered

yossipapi (changes requested):

PR #10340, opened 2020-12-31
PLAT-22548 volume and thumbnail fix Propus-16.13.0 merge

ZurKaltura:
Changing the code so getFileSyncServeFlavorFields will work as it used to on a non-cloud environment

### DIFF
--- a/alpha/apps/kaltura/lib/myPackagerUtils.class.php
+++ b/alpha/apps/kaltura/lib/myPackagerUtils.class.php
@@ -456,7 +456,7 @@ class myPackagerUtils
 			$preferredStorageId = myCloudUtils::getCloudPreferredStorage();
 		}
 
-		if(!$preferredStorageId)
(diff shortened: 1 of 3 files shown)


GElkayam:
LGTM

alpha/apps/kaltura/lib/myPackagerUtils.class.php
```diff
@@ -456,7 +456,7 @@ public static function getPreferredStorageId($dcId)
 			$preferredStorageId = myCloudUtils::getCloudPreferredStorage();
 		}
 
-		if(!$preferredStorageId)
+		if(!$preferredStorageId && myCloudUtils::isEnvironmentWithCloudStorage())
```
yossipapi, line 459: Why not use isClodDc?

ZurKaltura: @yossipapi isCloudDc will return if the dc is on cloud not if there is a dc with cloud in the environment setup

PR #10341, opened 2020-12-31
FOUN-52 - support s3 read and write batchFiles

gotlieb:
@yossipapi


alpha/lib/model/BatchJob.php
```diff
@@ -390,12 +390,14 @@ public function generateFilePathArr( $sub_type, $version = null, $externalPath =
 		if($externalPath)
 		{
 			$path = '/batchfiles/';
+			$path .= myContentStorage::getScatteredPathFromIntId($this->getPartnerId()) . '/' . $this->generateFileName($sub_type, $version);
```
yossipapi, line 393: why not use the jobId? 
It will give you a much better distribution of the files.

PR #10360, opened 2021-01-05
SUP-25144: add exception when elastic max matches were reached

avichainoach:


plugins/search/providers/elastic_search/services/ESearchService.php
```diff
@@ -117,7 +118,12 @@ protected static function initSearchActionParams($searchParams, KalturaPager $pa
 		$kPager = null;
 		if ($pager)
 		{
+			/* @var kPager $kPager */
 			$kPager = $pager->toObject();
+			if( ($kPager->getPageSize() * $kPager->getPageIndex()) > self::MAX_RESULT_WINDOW )
```
MosheMaorKaltura, line 123: Inside esearch service you can raise KalturaESearchError, since you are in the API side, no need to raise core exception and translate.

MosheMaorKaltura:
reviewed

PR #10372, opened 2021-01-07
REACH2-990: Add the ability to change the caption's label for machine…

avichainoach:
… type service

### DIFF
--- a/plugins/reach/lib/kReachManager.php
+++ b/plugins/reach/lib/kReachManager.php
@@ -310,7 +310,7 @@ class kReachManager implements kObjectChangedEventConsumer, kObjectCreatedEventC
 			&& $object->getStatus() == EntryVendorTaskStatus::READY
 		)
 		{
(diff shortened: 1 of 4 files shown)


plugins/reach/lib/kReachManager.php
```diff
@@ -304,12 +304,15 @@ public function objectChanged(BaseObject $object, array $modifiedColumns)
...
 		)
 			return $this->handleErrorTask($object);
-		
+
 		if ($object instanceof EntryVendorTask
 			&& in_array(EntryVendorTaskPeer::STATUS, $modifiedColumns)
 			&& $object->getStatus() == EntryVendorTaskStatus::READY
 		)
+		{
+			$this->addLabelAdditionForMachineType($object);
```
MosheMaorKaltura, line 313: ? did not forgot the human? or rename so it will work for both?

MosheMaorKaltura:
reviewed

PR #10375, opened 2021-01-10
PLAT-10788:prevent default symfony page from loading

ravitshalem:


MosheMaorKaltura:
reviewed

vendor/symfony-data/modules/default/templates/indexSuccess.php
```diff
@@ -0,0 +1,27 @@
+<meta http-equiv="refresh" content="0; url=http://corp.kaltura.com" />
+<!--?php decorate_with(sfLoader::getTemplatePath('default', 'defaultLayout.php')) ?>
```
gonenradai, line 2: why not just remove the entire commented part?
there's no need for it, and better not have some picky pen-tester look at the 200 response body and say "it says symfony"

ravitshalem: @gonenradai done

PR #10378, opened 2021-01-10
PLAT-22544: validate xml and html content on bulk

inbal-ben-david:


plugins/bulk_upload/base/services/BulkService.php
```diff
@@ -28,7 +28,14 @@ function addEntriesAction($fileData, KalturaBulkUploadJobData $bulkUploadData =
 	{
 		if(get_class($bulkUploadData) == 'KalturaBulkUploadJobData')
 			throw new KalturaAPIException(KalturaErrors::OBJECT_TYPE_ABSTRACT, 'KalturaBulkUploadJobData');
-			
+
+		$validContent = myXmlUtils::validateXmlFileContent($fileData["tmp_name"]);
```
MosheMaorKaltura, line 32: Use single quote

alpha/apps/kaltura/lib/myXmlUtils.php
```diff
@@ -0,0 +1,30 @@
...
+		if(!$filePath)
+		{
+			return true;
+		}
+
+		$fileType = kFileUtils::getMimeType($filePath);
+		if(strpos($fileType, 'html') !== false || strpos($fileType, 'xml') !== false)
+		{
+			$xmlContent = kFile::getFileContent($filePath);
+			if($xmlContent)
```
MosheMaorKaltura, line 16: No need to check xmlContent, since if file was empty, we would not found mimeType xml or html

PR #10385, opened 2021-01-12
PLAT-22565: simulive without specifying delivery profile

Amirhalfon1:
avoid override the entry & entryId in playManifest in case of Simulive (add servedEntryType instead)
remove the Simulive code from playManifest:serveVodEntry and move the logic to execute (initEventData function)
change DeliveryProfilePeer::getLocalDeliveryByPartner to receive isLive param
add supportsDeliveryDynamicAttributes to DeliveryProfileVod
add "simuliveSupport" to deliveryProfile custom data (and add checkbox to admin-console)
add "isSimuliveFlow" function to playManifestAction to detect if we're on flow of serving simulive


david-winder-kaltura:
When merge need to set  on default simulive DP:
1. Simulive support
2. Default true
3. Set weight to force order as last

PR #10399, opened 2021-01-14
PLAT-22584 - Zoom user is missing when trying to match cms_id

ZurKaltura:
Happens when cms_id is empty


MosheMaorKaltura:
reviewed

PR #10403, opened 2021-01-17
SUP-25018: Allow direct server w/o redirect

GElkayam:
in order to support source flavor limiting via ACL
crafted with @Adirv


GElkayam:
@yossipapi did the changes we discussed.

jessp01 (commented):
Unless I'm missing something here, you're better off with:
```php
 		elseif(strpos($livePackagerUrl, "{mn}") !== false){
...
}
```
Since the two conditions are in XOR relationship. It also makes it more readable.

GElkayam:
Closed as done in #10488

PR #10424, opened 2021-01-24
REACH2-1009:changing logic of duplicate tasks

ravitshalem:


plugins/reach/lib/kReachUtils.php
```diff
@@ -180,24 +180,40 @@ public static function checkPriceAddon($entryVendorTask, $taskPriceDiff)
 		return $remainingCredit >= 0 ? true : false;
 	}
 	
-	public static function isDuplicateTask($entryId, $catalogItemId, $partnerId, $version, $allowResubmission)
+	public static function isDuplicateTask($entryId, $catalogItemId, $partnerId, $version, $allowResubmission, $manualFlow)
```
MosheMaorKaltura, line 183: IsDuplicateTask, should work without any relation if allowResubmission or manualFlow.
The logic due to this params should be managed elsewhere

MosheMaorKaltura:
reviewed

PR #10431, opened 2021-01-27
No-Plat: disable isLive index in conf

david-winder-kaltura:


gotlieb (approved):

gotlieb (commented):
change map to 'elastic' not live

PR #10514, opened 2021-02-18
PLAT-22662: change EntryServerNode's 'isPlayableUser' default to false

Amirhalfon1:
change the default value of EntryServerNode:isPlayableUser to false to solve bug in "Explicit Live" feature


david-winder-kaltura (approved):

MosheMaorKaltura:
review

PR #10526, opened 2021-02-22
SUP-26477 add PENDING_ENTRY_READY status to retrieveActiveTasks() to prevent du…

salma2222:
…plication of entry vendor tasks


MosheMaorKaltura:
reviewed

PR #10527, opened 2021-02-22
SUP-24410 add entitledPusersView in entry.php to support entitledUser…

salma2222:
…sView in entries


MosheMaorKaltura:
reviewed

PR #10529, opened 2021-02-23
SUP-18687: fix SRT line index when clipping the beginning of the entry

amiras89:
@MosheMaorKaltura after reviewing, I believe this is OK
1. createNewCaptionFile is called
2. $unsupported_formats are being set (for scc and cap)
3. if caption is supported we call 'buildFile' at: plugins/content/caption/base/batch/CopyCaptions/KAsyncCopyCaptions.class.php:160
4. build file is abstract and implemented at the respective caption managers
5. **note** only srt and webVtt actually call "createCaptionsFile" which is where the fix for the line index is. Since dfxp does not have the same structure as srt and webvtt - I believe this is fine.

Happy to get your input
Thanks


MosheMaorKaltura:
reviewed

PR #10577, opened 2021-03-11
LIV-389: Added support for LiveNG recording url format

nadav-ravid:
LIV-386 #time 3h


alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -353,12 +353,22 @@ protected function getLivePackagerUrl($entryServerNode, $streamFormat = null)
 		
 		if($this->getDynamicAttributes()->getServeVodFromLive())
 		{
-			$entryId = $this->getDynamicAttributes()->getServeLiveAsVodEntryId();
-			$liveType = "/recording/";
-			$entry = entryPeer::retrieveByPK($entryId);
-			if ($entry && $entry->getFlowType() == EntryFlowType::LIVE_CLIPPING)
-				$liveType = "/clip/";
-			$livePackagerUrl = str_replace("/live/", $liveType, $livePackagerUrl);
+			if(strpos($livePackagerUrl, "front-live-service") !== false)
```
david-winder-kaltura, line 356: This should be on the server node obj and not with strpos

alpha/lib/model/MediaServerNode.php
```diff
@@ -90,4 +92,21 @@ public function getSessionType($entryServerNode)
 		return '';
 	}
 
+	public function getPartnerIdUrl(DeliveryProfileDynamicAttributes $da)
```
MosheMaorKaltura, line 95: either is an instance mathod that can take $this->getDynamicAttributes() from this, or static function that gets the value in the arguments.

alpha/lib/model/MediaServerNode.php
```diff
@@ -90,4 +92,21 @@ public function getSessionType($entryServerNode)
...
+	{
+		$partnerId = $da->getEntry()->getPartnerId();
+		return '/' . self::PARTNER_ID_URL_PARAM . '/' . $partnerId;
+	}
+
+	public function getEntryIdUrl(DeliveryProfileDynamicAttributes $da)
```
MosheMaorKaltura, line 101: same here

plugins/media/wowza/lib/model/WowzaMediaServerNode.php
```diff
@@ -233,4 +235,26 @@ public function getLiveServiceInternalDomain()
...
+		return parent::getEntryIdUrl($da);
+	}
+
+	public function modifyUrlForVodFromLive($liveUrl, DeliveryProfileDynamicAttributes $da)
+	{
+		$entryId = $da->getServeLiveAsVodEntryId();
+		$entry = entryPeer::retrieveByPK($entryId);
+
+		$liveType = self::RECORDING_LIVE_TYPE_URL;
+		if ($entry && $entry->getFlowType() == EntryFlowType::LIVE_CLIPPING)
```
MosheMaorKaltura, line 254: Coding convention - use {}

plugins/media/wowza/lib/model/WowzaMediaServerNode.php
```diff
@@ -233,4 +235,26 @@ public function getLiveServiceInternalDomain()
...
+			$liveType = self::CLIPPING_LIVE_TYPE_URL;
+		$liveUrl = str_replace("/live/", $liveType, $liveUrl);
```
MosheMaorKaltura, line 256: Use single quotes

alpha/lib/model/MediaServerNode.php
nadav-ravid: normally i would agree since the function does not access any instance specific information currently.
however i feel neither option fits well here. I think adding the dynamic attributes to the serverNode instance is wrong since the server node isn't entry or call specific. making the methods static would fit better but i don't like using static methods and params with abstract classes and inheritance due to the way they are evaluated in php

MosheMaorKaltura:
reviewed

PR #10585, opened 2021-03-16
SUP-25014: Fix entitlement validation for open entries when 'disable …

avichainoach:
…category limit' is enabled


MosheMaorKaltura:
reviewed

PR #10597, opened 2021-03-22
Added a new annotation that safely truncates the value of the annotated parameter

danbar0kal:
…ed parameter to a specified size


MosheMaorKaltura:
reviewed

PR #10598, opened 2021-03-24
PLAT-22694: appToken - ability to add privileges to KS if they are al…

avichainoach:
…ready set on the appToken


MosheMaorKaltura:
reviewed

PR #10599, opened 2021-03-24
LIV-472: remove "type" from delivery profile url when serveVodFromLive

Amirhalfon1:
- in case of server node of "LiveClusterMediaServerNode" type (LiveNG) and explicitLive is on and "getServeVodFromLive" is true - we don't want the "type" to exist in url as it contradict with the new recording implementation
- in case of LiveNG - add '/tl/' to the url anyway - regardless of explicitLive (with '/tl/main' by default)

### DIFF
--- a/alpha/lib/model/DeliveryProfileLive.php
+++ b/alpha/lib/model/DeliveryProfileLive.php
@@ -358,10 +358,7 @@ abstract class DeliveryProfileLive extends DeliveryProfile {
 		$livePackagerUrl .= $serverNode->getSessionType($entryServerNode);
 
 		$entry = $this->getDynamicAttributes()->getEntry();
-		if ($entry->getExplicitLive())
-		{
-			$livePackagerUrl .= $serverNode->getExplicitLiveUrl($this->getDynamicAttributes());
-		}
+		$livePackagerUrl .= $serverNode->getExplicitLiveUrl($livePackagerUrl, $this->getDynamicAttributes(), $entry);
 		$secureToken = $this->generateLiveSecuredPackagerToken($livePackagerUrl);
 		$livePackagerUrl .= "t/$secureToken/";
 
--- a/alpha/lib/model/DeliveryProfileLive.php
+++ b/alpha/lib/model/DeliveryProfileLive.php
@@ -1,8 +1,6 @@
 <?php
 
 abstract class DeliveryProfileLive extends DeliveryProfile {
-	const USER_TYPE_ADMIN = 'admin';
-	const USER_TYPE_USER = 'user';
 	const DEFAULT_MAINTENANCE_DC = -1;
 
 	/**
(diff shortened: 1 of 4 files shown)


david-winder-kaltura (approved):

alpha/lib/model/MediaServerNode.php
```diff
@@ -109,4 +112,14 @@ public static function modifyUrlForVodFromLive($liveUrl, DeliveryProfileDynamicA
 		return $liveUrl;
 	}
 
+	public static function getExplicitLiveUrl($liveUrl, LiveStreamEntry $entry)
+	{
+		$userType = static::USER_TYPE_ADMIN;
+		if ($entry->getExplicitLive() && !$entry->canViewExplicitLive())
```
MosheMaorKaltura, line 118: In the original code if $entry->getExplicitLive() was not set, then nothing was added to the URL

alpha/lib/model/MediaServerNode.php
```diff
@@ -109,4 +112,14 @@ public static function modifyUrlForVodFromLive($liveUrl, DeliveryProfileDynamicA
 		return $liveUrl;
 	}
 
+	public static function getExplicitLiveUrl($liveUrl, LiveStreamEntry $entry)
+	{
+		$userType = static::USER_TYPE_ADMIN;
```
MosheMaorKaltura, line 117: use self and not static

alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -360,13 +358,7 @@ protected function getLivePackagerUrl($entryServerNode, $streamFormat = null)
...
 
 		$entry = $this->getDynamicAttributes()->getEntry();
-		if ($entry->getExplicitLive())
-		{
-			$userType = self::USER_TYPE_ADMIN;
-		 	if (!$entry->canViewExplicitLive())
-				$userType = self::USER_TYPE_USER;
-			$livePackagerUrl .= "type/$userType/";
-		}
+		$livePackagerUrl .= $serverNode->getExplicitLiveUrl($livePackagerUrl, $entry);
```
MosheMaorKaltura, line 361: Access static function as instance function

alpha/lib/model/MediaServerNode.php
```diff
@@ -11,6 +11,9 @@ abstract class MediaServerNode extends DeliveryServerNode {
 	const DEFAULT_APPLICATION = 'kLive';
 	const ENTRY_ID_URL_PARAM = 'e';
 	const PARTNER_ID_URL_PARAM = 'p';
+	const EXPLICIT_LIVE_VIEWER_TYPE_URL = 'type';
```
MosheMaorKaltura, line 14: to create interface that all inheritors supply URL + user admin + user user please add functions that return it rather then const.

alpha/lib/model/MediaServerNode.php
Amirhalfon1: this behaviour is kept -  it protected from WowzaMediaServerNode [line 264](https://github.com/kaltura/server/pull/10599/files#diff-5ef80af7d044d33fdd433e5eb4257027948c3b601fb25b8959d6c2f6cc03320cR264) -
for the other implementation of "MediaServerNode" (which is only the LiveNG serverNode) we want to add "/tl/.../" to the url regardless of "$entry->getExplicitLive()"

alpha/lib/model/MediaServerNode.php
Amirhalfon1: We must use static as we want to take this value from the inheritance, self will cause it to take it from the parent (current) class - which isn't the value we want

alpha/lib/model/DeliveryProfileLive.php
Amirhalfon1: I aligned with the above string concatenations that also calling the static function that way.
From what I understand from PHP doc it is allowed 
https://www.php.net/manual/ro/language.oop5.static.php
>  A property declared as static cannot be accessed with an instantiated class object **(though a static method can).**

Do you think it is better to change it to instance function? (calling it the regular static way is impossible as we need to call the relevant implementation according to the serverNode type)

PR #10605, opened 2021-03-30
REACH2-1061 : Added default 'accuracy' value of 85% to Caption Asset

danbar0kal:


plugins/content/caption/base/lib/api/KalturaCaptionAsset.php
```diff
@@ -6,6 +6,9 @@
  */
 class KalturaCaptionAsset extends KalturaAsset
 {
```
MosheMaorKaltura, line 8: This entire logic must be encapsulated into ZoomPlugin, when adding new transcription file

PR #10616, opened 2021-04-05
PLAT-22592: add video/x-m4v as allowed file type

avichainoach:


alpha/apps/kaltura/lib/myUploadUtils.class.php
```diff
@@ -210,6 +210,8 @@ public static function checkIfFileIsAllowed($uploadFilePath)
 	{
 		$fileType = kFileUtils::getMimeType($uploadFilePath);
 
+		KalturaLog::info("File type: $fileType");
```
MosheMaorKaltura, line 213: debug

PR #10623, opened 2021-04-06
SUP-26711: prefer file sync from periodic_storage_ids if exists

amiras89:


amiras89:
Fix a case where postConvert and distributionSubmit (for YouTubeAPI) returned file sync from DC 1 which is not mounted on AWS env.

amiras89:
Closing - was separated into 2 PRs
https://github.com/kaltura/server/pull/10635
https://github.com/kaltura/server/pull/10640

PR #10625, opened 2021-04-06
SUP-27046: get width/height from S3 path

amiras89:


amiras89:
approved by @yossipapi F2F

PR #10630, opened 2021-04-07
ZoomIntegrationBranch-PLAT-22704:adding markup on entry that will be…

ravitshalem:
… used to block auto-transcription


MosheMaorKaltura:
reviewed

PR #10638, opened 2021-04-08
Update AppTokenService.php

hilak:
startSession should have the ksIgnored annotation to mark it as a ks-less action for integrating parties.


MosheMaorKaltura:
reviewed

PR #10639, opened 2021-04-10
SUP-27036 change bumperdata type in KalturaPlaybackContext to fix par…

salma2222:
…sing error in client library (java)


MosheMaorKaltura:
reviewed

PR #10653, opened 2021-04-18
move corpit to https

nivshlomokaltura:


alpha/apps/kaltura/modules/corp/actions/actions.class.php
```diff
@@ -15,17 +15,17 @@ class corpActions extends sfActions
    */
   public function executeIndex()
   {
-	$this->redirect("http://".kConf::get('corp_action_redirect'), 301);
+	$this->redirect("https://".kConf::get('corp_action_redirect'), 301);
```
MosheMaorKaltura, line 18: remove the http/s prefix and lets push it to the saas config with the protocol.

PR #10666, opened 2021-04-21
PLAT-22793:Allowing all partners to access dropFolderPlugin and remov…

ravitshalem:
…ing checkbox from partner configuration


MosheMaorKaltura:
reviewed

yossipapi:
@MosheMaorKaltura  Why are we removing this feature flip? If we dont use it maybe someone else does.

Wouldn't it be better to simply enable the flag on the template partner.

PR #10667, opened 2021-04-21
PLAT-22792:moving shared_screen to be lower in the prioritization of …

ravitshalem:
…recording type order


MosheMaorKaltura:
reviewed

PR #10684, opened 2021-04-28
PLAT-22737: Update every flavor asset status to ready when import fin…

avichainoach:
…ished for image entry


MosheMaorKaltura:
reviewed

PR #10686, opened 2021-04-28
Quasar 17.1.0 plat 22647

danbar0kal:
Zoom On Prem - Allow creating of Zoom Integration without oAuth2 call from Zoom


MosheMaorKaltura:
reviewed

PR #10694, opened 2021-04-29
PLAT-22737: for thumbAssets, when entry is image, create file sync pe…

avichainoach:
…r asset and not per entry


MosheMaorKaltura:
reviewed

PR #10712, opened 2021-05-03
PLAT-22490:adding category from scheduling tracking field

ravitshalem:


MosheMaorKaltura:
reviewed + notice that it points to 17.1.0.

PR #10722, opened 2021-05-05
SUP-26045: in case of rowData longer than what the DB can contain, sa…

avichainoach:
…ve the rest in custom data


alpha/lib/model/BulkUploadResult.php
```diff
@@ -9,7 +9,10 @@
  * @subpackage model
  */ 
 class BulkUploadResult extends BaseBulkUploadResult implements IBaseObject
-{   
+{
+	const ROW_DATA_EXTENSION = "row_data_extension";
```
MosheMaorKaltura, line 13: Use '

PR #10728, opened 2021-05-10
SUP-27631: remove quotation marks on 'user-agent' header when curling causing some servers to throw 403 - forbidden

amiras89:


MosheMaorKaltura:
reviewed

PR #10734, opened 2021-05-12
PLAT-22860:when ZoomTranscription field is disabled, do not pull tran…

ravitshalem:
…scription from zoom


MosheMaorKaltura:
reviewed

PR #10740, opened 2021-05-13
PLAT-22867:add webinar handling on new zoom

ravitshalem:


plugins/vendor/zoom/ZoomDropFolderPlugin/batch/processors/zoomWebinarProcessor.php
```diff
@@ -0,0 +1,61 @@
...
+			{
+				$zoomUser = new kZoomUser();
+				$zoomUser->setOriginalName($panelistEmail);
+				$zoomUser->setProcessedName($this->processZoomUserName($panelistEmail));
+				$result[] = $zoomUser;
+			}
+		}
+		else
+		{
+			$result = null;
```
MosheMaorKaltura, line 33: No need, empty array is as same as null, just init $result = array(); in the begining

plugins/vendor/zoom/lib/model/kZoomEventHanlder.php
```diff
@@ -124,7 +126,7 @@ protected static function createZoomDropFolderFile(kZoomEvent $event, $dropFolde
 				if(!array_key_exists($fileName, $dropFolderFilesMap))
 				{
 					if(!ZoomHelper::shouldHandleFileTypeEnum($recordingFile->recordingFileType) ||
-						($recordingFile->recordingFileType == kRecordingFileType::TRANSCRIPT && $enableZoomTranscription === 0))
+						($recordingFile->recordingFileType == kRecordingFileType::TRANSCRIPT && $zoomVendorIntegration->getEnableZoomTranscription() === 0))
```
MosheMaorKaltura, line 129: getEnableZoomTranscription() returns a bool, better check !=true

PR #10741, opened 2021-05-13
PSVAMB-24095: explicitly set entryId as string

GElkayam:
when binding to a prepared statement.


GElkayam:
@NadavHarnik please review.

PR #10744, opened 2021-05-16
simulive mpeg-ts wraparound fix

erankor:
when there's an mpeg-ts timestamp wraparound, the player fails to sync
the captions to the video, and the captions stop. this change moves the
clipStartTime so that it will be far from the wraparound value (1<<33).
as long as the duration of the simulive video is less than 13.5h, no
wraparound will occur.

### DIFF
--- a/alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
+++ b/alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
@@ -151,7 +151,7 @@ class serveFlavorAction extends kalturaAction
 			$mediaSet['liveWindowDuration'] = $dvrWindow;
 		}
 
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
```diff
@@ -148,6 +151,8 @@ public static function serveLiveMediaSet($durations, $sequences, $playlistStartT
 			$mediaSet['liveWindowDuration'] = $dvrWindow;
 		}
 
+		$mediaSet['timeOffset'] = $offset;
```
david-winder-kaltura, line 154: Maybe better to use the const in kSimuliveUtils::SCHEDULE_TIME_OFFSET_URL_PARAM - or maybe we need to move the const to here

PR #10750, opened 2021-05-18
PLAT 22767: BE doesn't work with formatted JSON

drorsou:


alpha/config/kConf.php
```diff
@@ -14,6 +14,7 @@
 set_include_path($include_path);
 
 require_once __DIR__ . '/../../infra/kEnvironment.php';
+require_once __DIR__ . '/../../infra/general/kString.class.php';
```
MosheMaorKaltura, line 17: remove the kString from here and add it to alpha/apps/kaltura/lib/request/infraRequestUtils.class.php

MosheMaorKaltura:
reviewed

PR #10752, opened 2021-05-18
callOauth: replace action names oauthValidationAction and preOauthVal…

ravitshalem:
…idation, so we wont change zoom app config


MosheMaorKaltura:
reviewed

PR #10753, opened 2021-05-18
LIV-558: fix tl missing in case of entryId with 'tl'

Amirhalfon1:
In case of entryId with 'tl' inside - the playManifest response mistakenly won't include the '/tl/' param
fix it by checking the 'tl' strpos **with slashes**


MosheMaorKaltura:
reviewed

PR #10754, opened 2021-05-18
PLAT-22864:when registration to another partner on the same integrati…

ravitshalem:
…on, first need to disable it


MosheMaorKaltura:
reviewed

PR #10755, opened 2021-05-18
PLAT-22858: searching of terms using starts with phrase is not interpreted correct to sphinx queries

drorsou:


MosheMaorKaltura:
reviewed

PR #10761, opened 2021-05-20
PLAT-22877:ignore createdat updatedat fields and remove set fields fr…

ravitshalem:
…om toObject that arent needed


MosheMaorKaltura:
reviewed

PR #10776, opened 2021-05-25
PLAT-22889:on transcription completed event search the parentEntry fo…

ravitshalem:
…r every transcitipon file


MosheMaorKaltura:
reviewed

PR #10781, opened 2021-05-26
PLAT-22891: change the auto deletion days from sec to days and set default values on integration

ravitshalem:


plugins/vendor/zoom/ZoomDropFolderPlugin/lib/kZoomDropFolderFlowManager.php
```diff
@@ -105,6 +146,23 @@ private static function getDropFolderStatus($v)
...
+		{
+			$zoomVendorIntegrationObject->setEnableMeetingUpload(true);
+		}
+		if ($zoomVendorIntegrationObject->getDeletionPolicy() === null)
+		{
+			$zoomVendorIntegrationObject->setDeletionPolicy(false);
+		}
+		if ($zoomVendorIntegrationObject->getEnableZoomTranscription() === null)
+		{
+			$zoomVendorIntegrationObject->setEnableZoomTranscription(false);
```
MosheMaorKaltura, line 161: set to true

PR #10783, opened 2021-05-26
PSVAMB-25378, PSVAMB-24119: match path only as a fallback

GElkayam:
This to support Pexip server with version greater than v21.
@gotlieb , @MosheMaorKaltura , please review.

### DIFF
--- a/plugins/sip/lib/model/providers/pexip/kPexipUtils.php
+++ b/plugins/sip/lib/model/providers/pexip/kPexipUtils.php
@@ -401,16 +401,12 @@ class kPexipUtils
 	}
 
 	/**
-	 * @param $result
 	 * @param $url
-	 * @param $headerSize
-	 * @return null
+	 * @param $headerData
+	 * @return int|null
 	 */
-	public static function extractIdFromCreatedResult($result,$url ,$headerSize)
+	protected static function getVirtualRoomId($url, $headerData)
 	{
-		$header = substr($result, 0, $headerSize);
-		$headerData = explode('\n', $header);
-		KalturaLog::info('Checking Headers ' . print_r($headerData, true));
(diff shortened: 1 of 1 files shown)


plugins/sip/lib/model/providers/pexip/kPexipUtils.php
```diff
@@ -423,6 +423,22 @@ public static function extractIdFromCreatedResult($result,$url ,$headerSize)
 				return $virtualRoomId;
 			}
 		}
+
```
MosheMaorKaltura, line 426: To avoid code duplication, create protected function getVirtualRoomId($url);
if it does not find return null.
And then call it once with the full URL if not fallback to the path only

GElkayam: @MosheMaorKaltura @gotlieb , Changes made, please review.

MosheMaorKaltura:
reviewed

plugins/sip/lib/model/providers/pexip/kPexipUtils.php
```diff
@@ -423,10 +419,33 @@ public static function extractIdFromCreatedResult($result,$url ,$headerSize)
...
+	 * @param $headerSize
+	 * @return int|null
+	 */
+	public static function extractIdFromCreatedResult($result, $url ,$headerSize)
+	{
+		$header = substr($result, 0, $headerSize);
+		$headerData = explode('\n', $header);
+		KalturaLog::info('Checking Headers ' . print_r($headerData, true));
+		$virtualRoomId = self::getVirtualRoomId($url, $headerData);
+		if (is_null($virtualRoomId))
```
jessp01, line 439: Why not add this attempt to obtain the ID into `getVirtualRoomId()` (in the event the first detection method failed)?

PR #10794, opened 2021-05-31
SUP-27646: resolve to "linked_id" fileSync if exists so that height, width and size be set properly

amiras89:


yossipapi:
Looks ok to me :)

PR #10810, opened 2021-06-02
PLAT-22921

hilak:
Resubmission should not be possible for tasks while an active task exists in the queue.


MosheMaorKaltura:
Reviewed

PR #10817, opened 2021-06-07
SUP-24729 change execute action to go to elastic if no filter is prov…

salma2222:
…ided - fix rule based playlist execution with exact match


MosheMaorKaltura:
reviewed

PR #10821, opened 2021-06-08
SUP-27927: fetch metadata from shared storage

amiras89:


plugins/metadata/lib/kMetadataKavaUtils.php
```diff
@@ -89,8 +93,10 @@ protected static function getMetadataFieldValues($objectIds, $partnerId, $metada
 				}
 				else
 				{
-					$fullPath = realpath($fileSync['FILE_ROOT'] . $fileSync['FILE_PATH']);
+					$fullPath = kFile::realPath($fileSync['FILE_ROOT'] . $fileSync['FILE_PATH']);
```
yossipapi, line 96: Better use kFile::getFileContent
Which supports both cases (file on NFS stoarge or file stored on S3).

plugins/metadata/lib/kMetadataKavaUtils.php
```diff
@@ -17,6 +17,10 @@ class kMetadataKavaUtils
 	
 	protected static function getMetadataReadyFileSyncs($objectIds, $partnerId, $metadataProfileId)
 	{
+		$currentDcId = kDataCenterMgr::getCurrentDcId();
```
yossipapi, line 20: I think its more nice to do:
$dcIds = kDataCenterMgr::getSharedStorageProfileIds();
$dcIds[] = kDataCenterMgr::getCurrentDcId();

Saves one line :)

plugins/metadata/lib/kMetadataKavaUtils.php
amiras89: Thanks for the tip!
Done (and tested)

plugins/metadata/lib/kMetadataKavaUtils.php
amiras89: agreed ;)

yossipapi:
Reviewed

PR #10825, opened 2021-06-09
Add support for OpenCalais API Endpoint configuration in partner leve…

hilak:
…l custom data

Remove unnecessary code

### DIFF
--- a/plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
+++ b/plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
@@ -9,10 +9,9 @@ require_once (__DIR__ . DIRECTORY_SEPARATOR . 'RuleEngine.php');
  */
 class kReachVendorTaskOpenCalaisProcessorEngine extends kReachVendorTaskProcessorEngine
 {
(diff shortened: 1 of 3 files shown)


plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
```diff
@@ -295,15 +294,18 @@ protected function retrieveAttachmentAssetContent ($assetId)
 
     protected function sendOpenCalaisGetTagsRequest ($transcript, $partnerId)
     {
+	    $openCalaisData = $this->getOpenCalaisSimpleXMLElementFromPartnerMetadata($partnerId);
+	    $url = $this->retrieveValueFromXml($openCalaisData, self::OPEN_CALAIS_API_URL_FIELD_NAME);
+
```
TalDubovKaltura, line 299: @hilak My suggestion is to have here condition: if(empty($url)){throw exception/write to log} - will happen if no url was set in KMS

hilak: @TalDubovKaltura I will add it, though the same can be achieved by making the endpoint field mandatory at the XSD level.

MosheMaorKaltura:
reviewed

PR #10826, opened 2021-06-13
SUP-27690-Duplicate-Add-To-Channel-Notification:Prevent sending email notification for live entry

OrelAri:
"Entry Was Added To Category" template is triggered by live entry and VOD entry. 
This update will prevent sending live entry notification and, therefore, prevent duplicate notification.


MosheMaorKaltura:
reviewed

PR #10830, opened 2021-06-14
SUP-27856 Check if asset is not CaptionAsset before updating entry st…

salma2222:
…atus to fix entry stuck in import status on adding captions from urlResource


MosheMaorKaltura:
reviewed

PR #10832, opened 2021-06-15
PLAT-22950: simulive - isLive true only when event is playable

Amirhalfon1:
change the dynamicGetter in case of simuliveFlow (LiveStreamScheduleEvent) to return playable only if we're at least 18 seconds inside the event (and not immediately from the event start as today)


david-winder-kaltura (approved):

MosheMaorKaltura:
reviewed

PR #10859, opened 2021-06-29
SUP-28118 add file extension when creating original flavor asset in a…

salma2222:
…ttachFileSync to fix file extension absence for source flavor asset created from fileSyncResource


MosheMaorKaltura:
reviewed

PR #10879, opened 2021-07-05
Quasar 17.5.0 plat 22973

MosheMaorKaltura:


gotlieb (commented):
make sure alphabetic input doesn't break the research query

PR #10888, opened 2021-07-06
Allow media server to get default conversion profile.

shamamayair:
For some scenario we need to get the default conversion profile in the liveNG.


MosheMaorKaltura:
reviewed

PR #10892, opened 2021-07-07
SUP-28338 add container format qt to kAssetUtils::getFileExtension to…

salma2222:
… support support update and get file ext for assets with qt container format


MosheMaorKaltura:
reviewed

PR #10893, opened 2021-07-07
Added encryptedSeed field to user

danbar0kal:


api_v3/lib/types/KalturaUser.php
```diff
@@ -100,6 +100,11 @@ class KalturaUser extends KalturaBaseUser
 	 * @var string
 	 */
 	public $ksPrivileges;
+	
+	/**
```
MosheMaorKaltura, line 104: Add readonly annotations

MosheMaorKaltura (commented):
1. No need to save the encryptedSeed  on the core object since the seed exist.
2. Set is not needed on core object since its a read only value
3. The getter needs to be able to calculate the key on the spot and since it happens only once no scale issue is relevant.

alpha/lib/model/kuser.php
```diff
@@ -27,6 +27,10 @@ public function __construct()
 	const COMPANY = 'company';
 
 	const CUSTOM_DATA_KS_PRIVILEGES = 'ks_privileges';
+	
+	const CUSTOM_DATA_ENCRYPTED_SEED = 'encryptedSeed';
+	const CIPHER_IV = '0000000000000000';
```
MosheMaorKaltura, line 32: delete

MosheMaorKaltura:
reviewed

gonenradai (approved):

PR #10896, opened 2021-07-12
PLAT-22995: In addLoginData avoid validating password if flag $checkPasswordStructure is false

drorsou:


MosheMaorKaltura:
reviewed

PR #10899, opened 2021-07-12
LIV-529 : upload source when media info changed

shamamayair:
In case we send null value as the flavor param upload with the default conversion profile (unless we have default recording conversion profile)


david-winder-kaltura (approved):

MosheMaorKaltura:
reviewed

PR #10901, opened 2021-07-13
SUP-27609: use zoomIntegration param rather than property which in undefined at kZoomTranscriptProcessor

amiras89:


MosheMaorKaltura:
reviewed

PR #10903, opened 2021-07-14
SUP-28338 add support for container formats mpeg audio, mpeg-ts and q…

salma2222:
…t in getFileExtension function in KAssetUtils to get file extensions for files with these container formats in DownloadAction and support script


MosheMaorKaltura:
reviewed

PR #10915, opened 2021-07-21
When copying template partner, also copy it's Reach profile

danbar0kal:


MosheMaorKaltura (commented):
Move the copy to kCopyPartner batch

danbar0kal:
obsolete

PR #10918, opened 2021-07-21
SUP-27559: fix SRT line index when clipping the beginning of the entry

amiras89:


amiras89:
@MosheMaorKaltura please note that parent method at kCaptionsContentManager->createCaptionsFile is public
Therefore child method also is public (and not private as we usually aim for)

amiras89:
@MosheMaorKaltura changed $newLineIndex to static in order to preserve it's value in case the clip / trim is being done in a non-liner manner, like: cut from 1-5 min + 10-15 min

plugins/content/caption/base/lib/captionManagers/srtCaptionsContentManager.php
```diff
@@ -139,4 +139,40 @@ public function merge($content, $toAppend)
...
+					if($adjustedTimeLine)
+						$currentBlock .=$adjustedTimeLine;
+					else
+						$shouldAddBlockToNewFile = false;
+				}
+				else
+					$currentBlock .= $line . kCaptionsContentManager::UNIX_LINE_ENDING;
+				$line = kCaptionsContentManager::getNextValueFromArray($originalFileContentArray);
+			}
+			if($shouldAddBlockToNewFile)
```
MosheMaorKaltura, line 168: No need to copy the entire code.
Just refactor createCaptionsFile on class kCaptionsContentManager
And only this code-block
AddBlockToNewFile needs to be override in the class of srtCaptionsContentManager.php.

plugins/content/caption/base/lib/captionManagers/srtCaptionsContentManager.php
```diff
@@ -139,4 +139,40 @@ public function merge($content, $toAppend)
 	{
 		return $content . $toAppend;
 	}
-}
\ No newline at end of file
+
+	public function createCaptionsFile($content, $clipStartTime, $clipEndTime, $timeCode, $globalOffset)
+	{
+		static $newLineIndex = 1;
```
MosheMaorKaltura, line 145: Why do we need it as static, since its only being referenced here and only within this function it can be local var or member of the class if you need to keep it between the iterations

amiras89: Tried to set it as a property of the class - but it's being reset every time we build a new 'chunk' of SRT
When load the class at: plugins/content/caption/base/batch/CopyCaptions/KAsyncCopyCaptions.class.php:159

MosheMaorKaltura:
reviewed

PR #10924, opened 2021-07-25
PLAT-22932: Add the ability to replace captions with MRSS

avichainoach:


plugins/content/caption/base/CaptionBulkUploadXmlPlugin.php
```diff
@@ -307,6 +309,33 @@ private function handleCaptionAsset($entryId, $conversionProfileId, SimpleXMLEle
 			$captionAssetPlugin->captionAsset->setContent($captionAssetId, $captionAssetResource);
 	}
 
+	protected function deleteCurrentCaptions($entryId)
+	{
+		KBatchBase::impersonate($this->xmlBulkUploadEngine->getCurrentPartnerId());
+
+		$filter = new KalturaAssetFilter();
+		$filter->entryIdEqual = $entryId;
+		$captionsList = KBatchBase::$kClient->captionAsset->listAction($filter);
```
MosheMaorKaltura, line 318: Does not make since to delete all the captions that were there.
One possible option is to filter also by the language

MosheMaorKaltura:
reviewed

PR #10935, opened 2021-07-28
SUP-28434-Clipping-Not-Ready-Audio-Flavor: Check if the audio flavor is ready for retrieve flavor for clipping

OrelAri:
Checking flavor status for audio flavor in source entry when clipping the entry.
Is the audio flavor do not ready we cannot clip the new audio flavors for the temp entry.

### DIFF
--- a/alpha/lib/model/assetPeer.php
+++ b/alpha/lib/model/assetPeer.php
@@ -825,6 +825,7 @@ class assetPeer extends BaseassetPeer implements IRelatedObjectPeer
 		$c = new Criteria();
 		$c->add(assetPeer::ENTRY_ID, $entryId);
 		$c->add(assetPeer::IS_ORIGINAL, 0);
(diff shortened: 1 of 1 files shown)


alpha/lib/model/assetPeer.php
```diff
@@ -825,6 +825,7 @@ public static function retrieveAudioFlavorsByEntryID($entryId)
 		$c = new Criteria();
 		$c->add(assetPeer::ENTRY_ID, $entryId);
 		$c->add(assetPeer::IS_ORIGINAL, 0);
+		$c->add(assetPeer::STATUS, asset::ASSET_STATUS_READY);
```
MosheMaorKaltura, line 828: Since this function can be use to get audio in all status, add the status as optional in the function input.
retrieveAudioFlavorsByEntryID($entryId, $statusList = array())
and add it as Status in.

The caller can add the ready status.

PR #10951, opened 2021-08-04
SUP-28369 change case sensitive compare to case insensitive to get us…

salma2222:
…er with email containing capital letters


MosheMaorKaltura:
reviewed

PR #10964, opened 2021-08-05
Static-Content: Use disable flag to avoid creating a db record for de…

yossipapi:
…fault behavior + fix bug when deleting original asset


alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2355,22 +2355,24 @@ public static function handleConvertProfileFinished(BatchJob $dbBatchJob, kConve
 
 		if($entry)
 		{
-			if(PermissionPeer::isValidForPartner(PermissionName::FEATURE_STATIC_CONTENT_CONVERSION, $entry->getPartnerId()))
+			KalturaLog::debug("TTT: handleConvertProfileFinished" . kBusinessPreConvertDL::shouldCheckStaticContentFlow($entry));
```
gotlieb, line 2358: Do you want to keep the debug TTT msg?

PR #10302, opened 2020-12-21
PLAT-22530: support captions in Simulive

Amirhalfon1:
get the entry's captionAssets and put it in the serveFlavor's json for the Simulive flow

PR #10968, opened 2021-08-08
PLAT-23050: check that the flavor has "getLanguage" callable function

Amirhalfon1:
"flavor" could be of type "flavorParams" which not have "getLanguage" method (the assumption that "flavor" is of "asset" type was wrong)
revert of https://github.com/kaltura/server/pull/10302/commits/d1bb9ca5785c155606e586aa2b72ecb7f2b6156f

### DIFF
--- a/alpha/lib/model/DeliveryProfile.php
+++ b/alpha/lib/model/DeliveryProfile.php
@@ -376,6 +376,7 @@ abstract class DeliveryProfile extends BaseDeliveryProfile implements IBaseObjec
 		$ext = null;
 		$audioLanguage = null;
 		$audioLanguageName = null;
+		$audioLanguageData = null;
 		$audioLabel = null;
 		$audioCodec = null;
 		$isDefaultAudio = false;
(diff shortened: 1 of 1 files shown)


alpha/lib/model/DeliveryProfile.php
```diff
@@ -397,7 +397,7 @@ protected function getFlavorAssetInfo($url, $urlPrefix = '', $flavor = null)
 					$audioLabel = $flavor->getLabel();
 				}
 				
-				$audioLanguageData = $flavor->getLanguage() ? $this->getAudioLanguage($flavor) : null;
+				$audioLanguageData = is_callable(array($flavor, 'getLanguage')) ? $this->getAudioLanguage($flavor) : null;
```
david-winder-kaltura, line 400: let do the is_callable in upper if - as currently it change the behavior in case getLanguage exist but return null

alpha/lib/model/DeliveryProfile.php
```diff
@@ -397,7 +398,10 @@ protected function getFlavorAssetInfo($url, $urlPrefix = '', $flavor = null)
 					$audioLabel = $flavor->getLabel();
 				}
 				
-				$audioLanguageData = $flavor->getLanguage() ? $this->getAudioLanguage($flavor) : null;
+				if (is_callable(array($flavor, 'getLanguage')))
```
MosheMaorKaltura, line 401: Instead use is typeof asset

Amirhalfon1: done

david-winder-kaltura (approved):

PR #10994, opened 2021-08-17
SUP-28499 delete declaration/definition of  from extending class Kalt…

salma2222:
…uraEntryVendorTaskFilter to prevent duplicate declaration/definition


MosheMaorKaltura:
this is already merged by other task - please close it.

PR #10995, opened 2021-08-17
SUP-28258 add kTimeZoneUtils class with function TimeZoneStrToTime to…

salma2222:
… convert time to unix taking into consideration the timezone


MosheMaorKaltura:
reviewed

PR #11005, opened 2021-08-19
Update 01.kaltura_sphinx_ce_tables.sql

hilak:


willbennett-kaltura:
@MosheMaorKaltura Can this get looked at?

GElkayam (changes requested):
deployment script needs to specify database name (kaltura_sphinx_log) as by default it runs on kaltura databae.
it is also missing the alter on sphinx_log_server

PR #11006, opened 2021-08-21
SUP-28787: fix bug due to PLAT-23020 where no entry result will return due to added conversion profile id in baseEntry filter

amiras89:


plugins/drop_folder/batch/Engine/KDropFolderFileTransferEngine.php
```diff
@@ -513,6 +513,13 @@ private function isEntryMatch(KalturaDropFolderContentProcessorJobData $data)
 			$entryPager->pageSize = 1;
 			$entryPager->pageIndex = 1;
 			$entryList = KBatchBase::$kClient->baseEntry->listAction($entryFilter, $entryPager);
+
+			if (is_array($entryList->objects) && !isset($entryList->objects[0]))
+			{
+				KalturaLog::debug('No matching entry with drop folder conversion profile, searching without conversion profile');
+				$entryFilter->conversionProfileIdEqual = null;
```
MosheMaorKaltura, line 520: This feature will break VE flow where we want only the entry with that REF + that conversion profile to work.
We will reject it for now, just revert it and will discuss it with VE client team.

PR #11016, opened 2021-08-24
PLAT-22929: Webex drop folder be able to scan unlimited history

drorsou:
… migrate all content


plugins/webexDropFolderPlugin/batch/KWebexDropFolderEngine.php
```diff
@@ -46,19 +47,42 @@ public function watchFolder(KalturaDropFolder $dropFolder)
...
+			$this->getFilesFromWebex(null, null);
+		}
+		
+		if ($this->dropFolder->fileDeletePolicy != KalturaDropFolderFileDeletePolicy::MANUAL_DELETE)
+		{
+			$this->purgeFiles();
+		}
+	}
+	
+	private function getFilesFromWebex($startTime, $endTime)
```
inbal-ben-david, line 84: use protected

PR #11022, opened 2021-08-25
FOUN-175 support custom CDN per partner

gotlieb:
@yossipapi

### DIFF
--- a/admin_console/forms/PartnerConfiguration.php
+++ b/admin_console/forms/PartnerConfiguration.php
@@ -150,16 +150,35 @@ class Form_PartnerConfiguration extends Infra_Form
 				'label'			=> 'CDN Host:',
 				'filters'		=> array('StringTrim'),
 		));
-		
+
+
+		$this->addElement('checkbox', 'checkbox_cdn_api_host', array(
+			'decorators' => array('ViewHelper', array('Label', array('placement' => 'append')), array('HtmlTag',  array('tag' => 'dt', 'class' => 'partner_configuration_checkbox_field')))
+		));
+
+		$this->addElement('text', 'cdn_api_host', array(
+			'label'			=> 'CDN Api Host:',
+			'filters'		=> array('StringTrim'),
+		));
+
+		$this->addElement('checkbox', 'checkbox_cdn_api_host_https', array(
+			'decorators' => array('ViewHelper', array('Label', array('placement' => 'append')), array('HtmlTag',  array('tag' => 'dt', 'class' => 'partner_configuration_checkbox_field')))
+		));
+
+		$this->addElement('text', 'cdn_api_host_https', array(
+			'label'			=> 'CDN Api Host Https:',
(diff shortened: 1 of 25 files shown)


admin_console/forms/PartnerConfiguration.php
```diff
@@ -150,16 +150,35 @@ public function init()
...
 				'filters'		=> array('StringTrim'),
 		));
-		
+
+
+		$this->addElement('checkbox', 'checkbox_cdn_api_host', array(
+			'decorators' => array('ViewHelper', array('Label', array('placement' => 'append')), array('HtmlTag',  array('tag' => 'dt', 'class' => 'partner_configuration_checkbox_field')))
+		));
+
+		$this->addElement('text', 'cdn_api_host', array(
```
MosheMaorKaltura, line 159: why do we need this? based on the URL protocol - http or https you can decide on the controller to which property to set this value.

Is there a use case where partner can have both HTTP and HTTPS different cdns?

alpha/apps/kaltura/lib/myPartnerUtils.class.php
```diff
@@ -370,37 +370,131 @@ public static function getCdnHost ( $partner_id, $protocol = null, $hostType = n
...
+
+	/**
+	 * @param null $partnerId
+	 * @param null $defaultValue
+	 * @return bool|mixed|string|null
+	 * @throws Exception
+	 */
+	public static function getApiCdnHostHttpsForPartner($partnerId = null, $defaultValue = null)
+	{
+		$host = kConf::hasParam('cdn_api_host_https') ? kConf::get('cdn_api_host_https') : $defaultValue;
```
MosheMaorKaltura, line 407: this is not needed , why not do it in function call kConf::get('cdn_api_host_https', mapname,$defaultValue);

alpha/apps/kaltura/modules/extwidget/actions/kwidgetAction.class.php
```diff
@@ -380,14 +380,21 @@ public function execute()
...
+					$partner_host = $cdnHost;
+				}
 				else if ($optimizedHost)
+				{
 					$partner_host = $optimizedHost;
+				}
 
-				if ($protocol == "https" && $partner_host = kConf::get("cdn_api_host"))
-					$partner_host = kConf::get("cdn_api_host_https");
+				if ($protocol == "https" && $partner_host = $cdnHost)
```
MosheMaorKaltura, line 394: Use const for https (there are already definitions

alpha/apps/kaltura/modules/extwidget/actions/kwidgetAction.class.php
```diff
@@ -380,14 +380,21 @@ public function execute()
...
+				{
```
MosheMaorKaltura, line 395: $partner_host = $cdnHost in the if, you meant ===>

PR #11032, opened 2021-08-31
See https://github.com/kaltura/clients-generator/issues/653

jessp01:


jessp01:
Already fixed in https://github.com/kaltura/server/pull/11007

PR #11045, opened 2021-09-05
Fix for uiconf crash in batch, 'unknown error' exception and a added log line

danbar0kal:


api_v3/services/PartnerService.php
```diff
@@ -141,6 +141,12 @@ public function registerAction( KalturaPartner $partner , $cmsPassword = "" , $t
 			$exceptionMessage = (str_replace('KalturaPartner::', '', $ex->getMessage()));
 			throw new KalturaAPIException( APIErrors::PARTNER_REGISTRATION_ERROR, $exceptionMessage);
 		}
+		catch (KalturaAPIException $ex)
+		{
+			KalturaLog::CRIT($ex);
+			$exceptionMessage = (str_replace('KalturaPartner::', '', $ex->getMessage()));
```
inbal-ben-david, line 147: i think that the error can come from multiple classes so lets remove this handling from here and from the core exception block and just throw $ex->getMessage()

PR #11052, opened 2021-09-10
Prevent catching the exception at this level

hilak:
Exception will be caught by KAsyncReachInternalQueueHandler.class.php where the job will terminate correctly.


MosheMaorKaltura:
Reviewed

PR #11054, opened 2021-09-12
Adding Zoom Transcript accuracy default value

danbar0kal:


alpha/config/kConfMapNames.php
```diff
@@ -8,4 +8,5 @@ class kConfMapNames
 	const EMBED_PLAYKIT = 'embed_playkit';
 	const REQUIRE_CATEGORY_ENTRY_FILTER_PARTNERS = 'partner_ids_require_category_entry_filter';
 	const COMMON_PASSWORDS = 'common_passwords';
+	const MAP_NAME = 'vendor';
```
inbal-ben-david, line 11: As this file contains map names in general change MAP_NAME to VENDOR

plugins/vendor/zoom/lib/kZoomFlowManager.php
```diff
@@ -0,0 +1,37 @@
...
+	
+	/**
+	 * @param BaseObject $object
+	 * @return bool
+	 */
+	public function objectCreated (BaseObject $object)
+	{
+		/* @var $object CaptionAsset */
+		$zoomConfiguration = kConf::get(self::CONFIGURATION_PARAM_NAME, kConfMapNames::MAP_NAME);
+		$object->setAccuracy($zoomConfiguration['ZoomTranscriptionAccuracy']);
```
inbal-ben-david, line 20: what happens if we dont have zoom map configured?
I would check $zoomConfiguration before trying to get property from it

inbal-ben-david (commented):
add to release notes the instruction for adding zoom plugin and the relevant config for the transcription accuracy

PR #11062, opened 2021-09-14
SUP-22014 compare also types when checking the field if was passed to…

salma2222:
… the filter in canTransformFilter, to fix the case of field is boolean and equals 0


inbal-ben-david:
reviewed 

as an improvement - you can implement a translation between is_root to root_entry_id as a special field so we will also search through elastic on this case

PR #11064, opened 2021-09-14
SUP-28533: Race condition between two zoom access token update

OrelAri:


MosheMaorKaltura:
This will not resolve the mutual exclusion, what you need to lock is the creation and saving of access tokens.

PR #11068, opened 2021-09-15
No-plat: Add beacon service access limits

yossipapi:


yossipapi:
Already handled as part of:
https://github.com/kaltura/server/pull/11072
Closing.

PR #11072, opened 2021-09-20
FOUN-240 apply rate limiter on responseCacher

gotlieb:
@yossipapi


yossipapi:
Looks ok

erankor:
I think we should avoid mixing code that runs without dispatcher with code that expects to run with dispatcher. So, IMHO it would be better to move the impl of the rate limit to KalturaResponseCacher and call it from the dispatcher. 
Other than that, instead of shouldCheckPartner, it would be more elegant to just pass partnerId as param.

PR #11085, opened 2021-09-27
PLAT-23102: updateLoginData action in user service allow saving a new…

drorsou:
… password that is identical to the current password


MosheMaorKaltura:
reviewed

PR #11102, opened 2021-10-05
SUP-29013 Update Status Pending Clip Entries To Error

OrelAri:


alpha/scripts/utils/updateEntrySatutsToError.php
```diff
@@ -0,0 +1,94 @@
+<?php
+
+require_once (dirname(__FILE__) . '/../bootstrap.php');
+
+if ($argc < 2)
+{
+	echo PHP_EOL . ' ---- Update Pending Clipping Entries ---- ' . PHP_EOL;
+	echo ' Execute: php ' . $argv[0] . ' [ /path/to/entries_id_list || entryId_1,entryId_2,.. || entry_id ] [realrun / dryrun]' . PHP_EOL;
+	die(' Error: missing entry_ids file, csv or single entry ' . PHP_EOL . PHP_EOL);
```
inbal-ben-david, line 9: instead of logging then die just change the die message to something like
die("Usage: php updateEntrySatutsToError.php < /path/to/entries_id_list || entryId_1,entryId_2,.. || entry_id > <realrun / dryrun>"."\n");

inbal-ben-david (commented):
- you are sometimes using echo and sometimes use kalturaLog. use only kalturaLog (you will also won't need all those PHP_EOL)
- keep coding convention for { } on new lines

inbal-ben-david:
reviewed

PR #11111, opened 2021-10-07
SUP-28652: fix Zoom registration failing due to deletionPolicy=1 or object does not return from DB + reactivating old "deleted" DF instead of creating a new DF

amiras89:


amiras89:
Reviewed with @yossipapi  and @ravitshalem

PR #11112, opened 2021-10-10
SUP-28258 add kTimeZoneUtils class with function TimeZoneStrToTime to convert time to unix taking into consideration the timezone

salma2222:


inbal-ben-david:
reviewed

PR #11117, opened 2021-10-11
PLAT-23201: when changing to status read only present the child partn…

ravitshalem:
…ers as read only


MosheMaorKaltura:
reviewed

PR #11183, opened 2021-11-03
PLAT-23257-HF: fix incorrect getDetails response

Amirhalfon1:
on "dynamicGetter" of LiveStreamScheduleEvent - where we check if the simulive event is currently playable - check the case that the current time is actually equal to the "startTime" + 18 (change "<" to "<="). otherwise the response will be considered as "not live" and will be saved on cache until the end of the event with this incorrect response.
add log print to "getDetails" of the calculated broadcast status


david-winder-kaltura (approved):

MosheMaorKaltura:
Reviewed

PR #11199, opened 2021-11-08
SUP-20297: support REACH automatic rules when updating External-Media type (KMS current flow)

amiras89:
support REACH automatic rules for External-Media type when updating duration after entry was added (KMS current flow)

* When KMS add External Media (YouTube) they first do externalMedia-->add and only then update with 'msDuration'


amiras89:
Was approved by @yossipapi face to face

PR #11215, opened 2021-11-11
SUP-29888 change zoom call to bring page size and reduce date range o…

salma2222:
…f listing zoom recordings - to prevent missed content

### DIFF
--- a/plugins/vendor/zoom/ZoomDropFolderPlugin/batch/KZoomDropFolderEngine.php
+++ b/plugins/vendor/zoom/ZoomDropFolderPlugin/batch/KZoomDropFolderEngine.php
@@ -102,7 +102,7 @@ class KZoomDropFolderEngine extends KDropFolderFileTransferEngine
 			$fromInSec = time() - self::MAX_DATE_RANGE_DAYS * self::ONE_DAY;
 		}
 
-		$toInSec = min(time(), $fromInSec + self::ONE_DAY * 30);
+		$toInSec = min(time(), $fromInSec + self::ONE_DAY);
 		$from = date('Y-m-d', $fromInSec);
 		$to = date('Y-m-d', $toInSec);
 		$nextPageToken = '';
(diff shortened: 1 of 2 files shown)


plugins/vendor/zoom/infra/kZoomClient.php
```diff
@@ -112,7 +112,7 @@ public function deleteRecordingFile($meetingUUid, $recodingId)
 	public function listRecordings($accountId, $from, $to, $nextPageToken, $pageSize)
 	{
 		$apiPath = str_replace('@accountId@', $accountId, self::API_LIST_RECORDING);
-		$apiPath .= '?from=' . $from . '&to=' . $to . '&next_page_token=' . $nextPageToken . '&pageSize=' . $pageSize;
+		$apiPath .= '&page_size=' . $pageSize . '&next_page_token=' . $nextPageToken . '?from=' . $from . '&to=' . $to;
```
MosheMaorKaltura, line 115: In the api call of Zoom get recording ( https://marketplace.zoom.us/docs/api-reference/zoom-api/cloud-recording/getaccountcloudrecording ) the account id is part of the request parameters and the rest of parameters are query parameters, meaning that they must come after the - ?
This is how a correct call should look like '?from=' . $from . '&to=' . $to . '&next_page_token=' . $nextPageToken . '&page_size=' . $pageSize;

plugins/vendor/zoom/ZoomDropFolderPlugin/batch/KZoomDropFolderEngine.php
```diff
@@ -102,7 +102,7 @@ protected function getMeetingsInStartTimeOrder()
 			$fromInSec = time() - self::MAX_DATE_RANGE_DAYS * self::ONE_DAY;
 		}
 
-		$toInSec = min(time(), $fromInSec + self::ONE_DAY * 30);
+		$toInSec = min(time(), $fromInSec + self::ONE_DAY);
```
MosheMaorKaltura, line 105: This means that we will make a least on call per day, customer that has initial content of 3 years will take to long to get its value.e

PR #11229, opened 2021-11-16
FOUN-288: Support decommissioning dc without deleting all file syncs that reside on it

amiras89:
@yossipapi only handled a single flow
There are 9 other places calling getDcById and some are not taking into consideration that it might return 'null'

We either fix all, or only specific flows (as this PR does)
Thanks


yossipapi:
Reviewed

PR #11239, opened 2021-11-21
SUP-29794 check the conversion profile against null before accessing …

salma2222:
…member function getDefaultAudioLang()


inbal-ben-david:
reviewed

PR #11246, opened 2021-11-24
FOUN-291: when making "kaltura_batch.sh restart" do not kill existing jobs

amiras89:


amiras89:
Opened new PR with slight change
https://github.com/kaltura/server/pull/11326/files

PR #11250, opened 2021-11-29
SUP-29836: added new languages to live init and update

nadav-ravid:


nadav-ravid:
Tested the update script to test the ini params

MosheMaorKaltura:
reviewed

PR #11251, opened 2021-11-30
removed sub-type validation from Virtual Event

danbar0kal:


plugins/virtual_event/services/VirtualEventService.php
```diff
@@ -150,32 +150,25 @@ protected function validateScheduleEvents (KalturaVirtualEvent $virtualEvent)
 		$partnerId = kCurrentContext::getCurrentPartnerId();
 		if ($virtualEvent->agendaScheduleEventId)
 		{
```
inbal-ben-david, line 152: would change it to run in loop on array of virtualScheduleEvents, check if they set and call to validate instead of having 3 if conditions

inbal-ben-david:
reviewed

PR #11261, opened 2021-12-01
FOUN-301: count userCpu usage for convert process

amiras89:
@yossipapi this is not the final pr
will clarify in teams


amiras89:
Approved by @yossipapi f2f

PR #11298, opened 2021-12-20
KMS-24338: Reject strings that contain certain special chars for cert…

avichainoach:
…ain properties on Kuser object


alpha/apps/kaltura/lib/myKuserUtils.class.php
```diff
@@ -2,6 +2,7 @@
 
 class myKuserUtils
 {
+	const SPECIAL_CHARS = array('+', '=', '-', '@', ',');
```
inbal-ben-david, line 5: const with array is not supported in low php versions
better just put it in the function

api_v3/lib/types/KalturaBaseUser.php
```diff
@@ -215,17 +215,37 @@ public function getFilterDocs()
...
+			if(strpos($this->$kalturaProperty, kuser::URL_PATTERN) !== false)
+			{
+				throw new KalturaAPIException(KalturaErrors::INVALID_FIELD_VALUE, $kalturaProperty);
+			}
+			
+			//If the core object already contains a "bad" value- don't apply the new restriction
+			if($coreObject)
+			{
+				$dbVal = $coreObject->$dbGetFunction();
+				if( !is_null($dbVal) && myKuserUtils::startsWithSpecialChar($dbVal) )
```
inbal-ben-david, line 240: indentation

PR #11300, opened 2021-12-20
PLAT-23325

danbar0kal:


api_v3/lib/KalturaResponseCacher.php
```diff
@@ -17,6 +17,10 @@ class KalturaResponseCacher extends kApiCache
 	const RESPONSE_TYPE_PHP = 3;
 
 	const BATCH_PARTNER_ID = -1;
+	
+	const POST_INCREMENT = '_postIncrement';
+	
+	static protected $generatedKey;
```
inbal-ben-david, line 23: change to rateLimitKey

api_v3/lib/KalturaResponseCacher.php
```diff
@@ -672,6 +683,22 @@ public static function rateLimit($service, $action, $params, $partnerId = null,
 
 		return false;
 	}
+	
+	public static function incrementPostCallRateLimit($params, $partnerId = null, $ksPartnerId = null)
```
inbal-ben-david, line 687: change to rateLimitPostProcessing, we also dont need all the params

api_v3/lib/KalturaResponseCacher.php
```diff
@@ -672,6 +683,22 @@ public static function rateLimit($service, $action, $params, $partnerId = null,
...
+	{
+		$rule = self::getRateLimitRule($params, $partnerId, $ksPartnerId);
```
inbal-ben-david, line 689: why do we need to check the rule again? we can just check if we stored something in self::$generatedKey

PR #11318, opened 2021-12-29
PLAT-23259: add "shouldRedirect" flag to live delivery profile

Amirhalfon1:
add "shouldRedirect" flag to KalturaDeliveryProfileLivePackager (configurable from admin console form)

### DIFF
--- a/alpha/lib/model/DeliveryProfileLive.php
+++ b/alpha/lib/model/DeliveryProfileLive.php
@@ -2,6 +2,7 @@
 
 abstract class DeliveryProfileLive extends DeliveryProfile {
 	const DEFAULT_MAINTENANCE_DC = -1;
+	const SHOULD_REDIRECT = "should_redirect";
 
 	/**
 	 * @var kLiveStreamConfiguration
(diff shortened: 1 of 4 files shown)


alpha/lib/model/DeliveryProfileLive.php
```diff
@@ -462,5 +462,18 @@ public function getLivePackagerSigningDomain()
 	{
 		return $this->getFromCustomData("livePackagerSigningDomain");
 	}
+
+	public function setShouldRedirect($v)
+	{
+		$this->shouldRedirect = $v;
```
nadav-ravid, line 468: why add this line. doesn't it raise an error since you removed the parameter from the object?

nadav-ravid (changes requested):

alpha/lib/model/DeliveryProfileLive.php
david-winder-kaltura: As far as I see, the "shouldRedirect" is still on the obj - so no error.
However - I not sure we want to set this when calling "setShouldRedirect" - this function will use for setting the default state and the actual property will be change when it is dynamically done during the playManifest request.
Due to the fact you didn't use this function in this PR I guess that was the original intention - so this line still need to be remove in my opinion
The other way to do it is define new property and use him but I not sure we want to go there

Amirhalfon1: My first intention was to avoid case that someone in the future will call "setShouldRedirect" from the "playManifestAction" for example, and expect that the internal parameter will be changed.
It can be removed but it might cause misunderstandings in the future

Amirhalfon1: anyway - updated the PR, the setter won't affect the internal "shouldRedirect" anymore. added comment to the setter

nadav-ravid (approved):

david-winder-kaltura (approved):

inbal-ben-david (approved):

PR #11322, opened 2022-01-03
SUP-30407: Add a property in partner object & configuration- 'Load th…

avichainoach:
…umbnail with KS'


avichainoach:
will merge a different solution

PR #11323, opened 2022-01-03
SUP-28690: Removing audio flavors from youtubeAPI distribution

OrelAri:


plugins/content_distribution/lib/kContentDistributionManager.php
```diff
@@ -74,6 +74,19 @@ protected static function prepareDistributionJob(EntryDistribution $entryDistrib
 			$entryDistribution->getThumbAssetIds(),
 			$entryDistribution->getFlavorAssetIds()
 		)));
+
+		if($distributionProfile->shouldExcludeAudioFlavors())
+		{
+			$audios = assetPeer::retrieveAudioFlavorsByEntryID($entryDistribution->getEntryId());
```
inbal-ben-david, line 80: change to audioFlavorAssets

plugins/content_distribution/lib/kContentDistributionManager.php
```diff
@@ -74,6 +74,19 @@ protected static function prepareDistributionJob(EntryDistribution $entryDistrib
...
+			$audioIds = array();
+			foreach($audios as $audio)
```
inbal-ben-david, line 82: change audios to audioAssetIds
change audio to audioAssetId

plugins/content_distribution/lib/kContentDistributionManager.php
```diff
@@ -74,6 +74,19 @@ protected static function prepareDistributionJob(EntryDistribution $entryDistrib
...
+			{
+				$id = $audio->getId();
+				if (in_array($id, $assetIds))
```
inbal-ben-david, line 85: i don't think you need to check that id is part of the assetIds
array_diff already returns the values that exist only in the first array

inbal-ben-david (approved):

PR #11324, opened 2022-01-03
Get backup entry-server-node for live thumb

david-winder-kaltura:


david-winder-kaltura:
We had a race where the retrieve function return null so we had array with null object which fail the function later on.
As we want to get fallback in case of no primary (even for recording) we will check for both type and then sort to get the primary esn first

inbal-ben-david (approved):

PR #11326, opened 2022-01-03
FOUN-291: add "reload" to "kaltura_batch.sh" which will not kill existing jobs (workers) causing zombie process

amiras89:


yossipapi:
Reviewed 
Let's push to the next release

PR #11330, opened 2022-01-04
Simplify if statement in live-thumb capture

david-winder-kaltura:


alpha/apps/kaltura/lib/myPackagerUtils.class.php
```diff
@@ -380,10 +380,9 @@ protected static function captureLiveThumb(entry $entry, $destThumbPath, $calc_v
 					continue;
 				}
 
-				$result = self::curlThumbUrlWithOffset('', $calc_vid_sec, $serverNodeUrl, $destThumbPath, $width, $height, '+', '', "-s{$liveParam->getFlavorId()}");
-				if($result)
+				if (self::curlThumbUrlWithOffset('', $calc_vid_sec, $serverNodeUrl, $destThumbPath, $width, $height, '+', '', "-s{$liveParam->getFlavorId()}"))
```
nadav-ravid, line 383: can be further simplified to just
```suggestion
return self::curlThumbUrlWithOffset('', $calc_vid_sec, $serverNodeUrl, $destThumbPath, $width, $height, '+', '', "-s{$liveParam->getFlavorId()}")
```

david-winder-kaltura: Doing this will exit the func even if the curl call if fail. But in this case we want to continue the loop

PR #11366, opened 2022-01-18
Quasar 17.19.0 PLAT 23498

danbar0kal:


api_v3/lib/KalturaErrors.php
```diff
@@ -174,6 +174,8 @@ class KalturaErrors extends APIErrors
 	const VOLUME_MAP_NOT_CONFIGURED = "VOLUME_MAP_NOT_CONFIGURED;;Need to add volume map support to configuration";
 
 	const RESOURCE_ENTRY_ID_MISSING= "RESOURCE_ENTRY_ID_MISSING;;Entry Id on resource object is missing";
+	
+	const CYCLE_IN_PARENTAGE = "CYCLE_IN_PARENTAGE;;Parentage has cycle";
```
inbal-ben-david, line 178: i would change it to something like invalid connection between parent and child or something similar as we throw it in the api level

api_v3/lib/KalturaEntryService.php
```diff
@@ -1551,6 +1551,28 @@ protected function validateEntryScheduleDates(KalturaBaseEntry $entry, entry $db
...
+		{
+			return;
+		}
+		$slowParentEntryPointer = $dbEntry;
+		$slowParentEntryPointer->setParentEntryId($entry->parentEntryId);
+		$fastParentEntryPointer = entryPeer::retrieveByPK($slowParentEntryPointer->getParentEntryId());
+		
+		while ($slowParentEntryPointer !== $fastParentEntryPointer)
+		{
+			if ($fastParentEntryPointer == null || !$fastParentEntryPointer->getParentEntryId())
```
inbal-ben-david, line 1566: shouldn't we make sure that slow pointer is also not null?

api_v3/lib/KalturaEntryService.php
```diff
@@ -1551,6 +1551,28 @@ protected function validateEntryScheduleDates(KalturaBaseEntry $entry, entry $db
...
+			{
+				return;
+			}
+			
+			$slowParentEntryPointer = entryPeer::retrieveByPK($slowParentEntryPointer->getParentEntryId());
+			$fastParentEntryPointer = entryPeer::retrieveByPK(entryPeer::retrieveByPK($fastParentEntryPointer->getId())->getId());
```
inbal-ben-david, line 1572: if nothing returned on the first retrieve you will try to get id on a non object

inbal-ben-david (approved):

PR #11398, opened 2022-02-07
FOUN-350: limit object index size up to 1MB due to mysql current config "max_allowed_packet" = 1MB

amiras89:


plugins/search/providers/elastic_search/lib/kElasticSearchManager.php
```diff
@@ -233,8 +234,16 @@ private function shouldSyncElastic($object)
 
     private function shouldSkipSaveToSphinxLog($object, &$command)
     {
-        // limit the number of large SQLs to 1/min per object, since they load the sphinx log database and elastic servers
-	$commandSize = strlen($command);
+		$commandSize = strlen($command);
```
yossipapi, line 237: We should check size is not passed in the zipped value

plugins/search/providers/elastic_search/lib/kElasticSearchManager.php
```diff
@@ -13,6 +13,7 @@ class kElasticSearchManager implements kObjectReadyForIndexEventConsumer, kObjec
     const CACHE_PREFIX_STICKY_SESSIONS = 'elastic_large_sql_lock_';
     const REPETITIVE_UPDATES_CONFIG_KEY = 'skip_elastic_repetitive_updates';
     const METADATA_MAX_LENGTH = 256;
+	const SQL_MAX_ALLOWED_PACKET = 1048576; // 1MB (1048576 / 1024 / 1024)
```
GElkayam, line 16: indent

plugins/search/providers/elastic_search/lib/kElasticSearchManager.php
```diff
@@ -233,8 +234,16 @@ private function shouldSyncElastic($object)
...
     private function shouldSkipSaveToSphinxLog($object, &$command)
     {
-        // limit the number of large SQLs to 1/min per object, since they load the sphinx log database and elastic servers
-	$commandSize = strlen($command);
+		$commandSize = strlen($command);
+		
+		// limit object index size up to 1MB due to mysql current 'max_allowed_packet' config which is 1MB
+		if ($commandSize > self::SQL_MAX_ALLOWED_PACKET)
+		{
+			KalturaLog::log('skipping saving elastic sphinxLog sql for [' . $object->getId() . '] - command size [' . $commandSize .'] larger than 1MB');
```
GElkayam, line 242: wouldn't that make Elasticsearch clusters inconsistent when using Multi DC?
Also, as this is a tunable value (OnPerms and Regional Clouds set that to 16MB) on the sql side, should this be adjustable as well?

plugins/search/providers/elastic_search/lib/kElasticSearchManager.php
amiras89: Good point - noted.
For now we're not handling it since it's fairly rare and only impact the specific object
Also, per Yossi comment above, when we will proceed with this fix I'll make it after we compress the data

PR #11412, opened 2022-02-14
SUP-28302 get file extension from redirect url to set file extension upon import

salma2222:


batch/batches/Import/KAsyncImport.class.php
```diff
@@ -187,7 +187,18 @@ private function fetchFile(KalturaBatchJob $job, KalturaImportJobData $data)
 			else
 			{
 				// creates a temp file path
-				$data->destFileLocalPath = $this->getTempFilePath($sourceUrl, $fileSize);
+				$curlWrapperForRedirect = new KCurlWrapper(self::$taskConfig->params);
+				$curlWrapperForRedirect->setTimeout(self::HEADERS_TIMEOUT);
+				$curlWrapperForRedirect->exec($sourceUrl);
```
inbal-ben-david, line 192: make sure you get only the headers and not the content

batch/batches/Import/KAsyncImport.class.php
```diff
@@ -187,7 +187,18 @@ private function fetchFile(KalturaBatchJob $job, KalturaImportJobData $data)
...
+				$redirectUrl = $curlWrapperForRedirect->getInfo(CURLINFO_EFFECTIVE_URL);
+				$curlWrapperForRedirect->close();
+
+				if(is_null($redirectUrl))
+				{
+					$redirectUrl = $sourceUrl;
+				}
+
+				$data->destFileLocalPath = $this->getTempFilePath($redirectUrl, $fileSize);
```
inbal-ben-david, line 201: how it affects the rest of the flow?
on line 207 you are using the source url and not redirect

PR #11417, opened 2022-02-16
PLAT-23516: new capability to sso login

danbar0kal:


alpha/lib/model/kuser.php
```diff
@@ -675,6 +675,22 @@ public function getIsAdmin()
 		return parent::getisAdmin() == true;
 	}
 	
+	/**
+	 * @return is_sso_excluded parameter from custom_data
+	 */
+	public function getIsSsoExcluded()
+	{
+		return $this->getFromCustomData('is_sso_excluded') == true;
```
inbal-ben-david, line 683: just call $this->getFromCustomData('is_sso_excluded', null, false)

alpha/lib/model/kuser.php
```diff
@@ -675,6 +675,22 @@ public function getIsAdmin()
...
+	}
+	
+	/**
+	 * @param bool $isSsoExcluded
+	 */
+	public function setIsSsoExcluded(bool $isSsoExcluded)
```
inbal-ben-david, line 689: no need to declare the var type in the signature of the function

alpha/lib/model/kuser.php
```diff
@@ -675,6 +675,22 @@ public function getIsAdmin()
...
+	{
+		$this->putInCustomData('is_sso_excluded', $isSsoExcluded);
```
inbal-ben-david, line 691: move is_sso_excluded to const

api_v3/lib/types/KalturaUser.php
```diff
@@ -187,4 +194,12 @@ public function validateForUpdate($sourceObject, $propertiesToSkip = array())
 		$this->validateNames($sourceObject, self::$names);
 		parent::validateForUpdate($sourceObject, $propertiesToSkip);
 	}
+	
+	protected function validateSsoExclusion()
+	{
+		if(is_null($this->isSsoExcluded))
```
inbal-ben-david, line 200: dont think you will need this function if the default value will return as false from isSsoExcluded

inbal-ben-david (approved):

PR #11420, opened 2022-02-17
PLAT-23522: emails for updating users

danbar0kal:


alpha/lib/model/kuser.php
```diff
@@ -158,6 +158,11 @@ public function postUpdate(PropelPDO $con = null)
 		if($this->isColumnModified(kuserPeer::STATUS) && $this->getStatus() == KuserStatus::DELETED) {
 			$objectDeleted = true;
 		}
+		
+		if ($this->isCustomDataModified(null, 'is_sso_excluded'))
```
inbal-ben-david, line 162: you should pass the const for is_sso_excluded as the first arg, the field name is the name var in the customDataModified function

alpha/lib/model/kuserPeer.php
```diff
@@ -707,6 +707,24 @@ public static function sendNewUserMail(kuser $user, $existingUser)
 			);
 		}
 	}
+	
+	protected static function getAuthenticationType($user, $partner)
+	{
+		/* @var $user kuser*/
+		$authType = $partner->getAuthenticationType();
+		if ($authType == PartnerAuthenticationType::SSO)
```
inbal-ben-david, line 715: if(!sso)
   return authType

alpha/lib/model/kuserPeer.php
```diff
@@ -707,6 +707,24 @@ public static function sendNewUserMail(kuser $user, $existingUser)
...
+		{
+			if ($user->getIsSsoExcluded())
```
inbal-ben-david, line 717: you can just ask if ($user->getIsSsoExcluded() && !$partner->getUseTwoFactorAuthentication())
then return password only
otherwise you can just return the auth type of the partner

alpha/lib/model/kuserPeer.php
```diff
@@ -707,6 +707,24 @@ public static function sendNewUserMail(kuser $user, $existingUser)
 			);
 		}
 	}
+	
+	protected static function getAuthenticationType($user, $partner)
```
inbal-ben-david, line 711: need to use this function also in verifyAuthenticatedPartnerSwitch

alpha/lib/model/kuserPeer.php
danbar0kal: what you are suggesting is missing some use cases

PR #11421, opened 2022-02-17
SUP-30667 change html purifier doctype to HTML 4.01 Transitional

salma2222:


infra/general/kHtmlPurifier.class.php
```diff
@@ -91,6 +92,7 @@ public static function initHTMLPurifier()
 		{
 			$config = HTMLPurifier_Config::createDefault();
 			$config->set('Cache.DefinitionImpl', null);
+			$config->set('HTML.Doctype', self::HTML_DOCTYPE);
```
inbal-ben-david, line 95: take the property from the html purifier dynamic config

inbal-ben-david (approved):

PR #11443, opened 2022-03-01
FOUN-353: skip MR flavor delete actions for "static" content

amiras89:


amiras89:
Approved per Teams meeting with @yossipapi

PR #11471, opened 2022-03-13
SUP-31376: Add and submit requests for entry_distribution with distibution profile of the request partner

OrelAri:


OrelAri:
Adding check when distribution profile been use is not belong to the Pid that created the request.
In such situations the error that return is the distributor profile not found.
Modify the action of add, submitAddAction, submitUpdateAction and submitRetryAction.
For Cross Kaltura Distribution the distribution flow work as expected.

inbal-ben-david (approved):

PR #11477, opened 2022-03-14
SUP-30689: Removing start with special char check for user name validation

OrelAri:


api_v3/lib/types/KalturaBaseUser.php
```diff
@@ -6,6 +6,8 @@
 */
 class KalturaBaseUser extends KalturaObject implements IRelatedFilterable
 {
+	const CHAR_APOSTROPHE 	= "'";
```
inbal-ben-david, line 9: fix spacing

api_v3/lib/types/KalturaBaseUser.php
```diff
@@ -241,10 +243,13 @@ protected function validateNames($coreObject, array $names)
...
 			}
-			
+
 			if(myKuserUtils::startsWithSpecialChar($this->$kalturaProperty))
 			{
-				throw new KalturaAPIException(KalturaErrors::INVALID_FIELD_VALUE, $kalturaProperty);
+				if (strlen($this->$kalturaProperty) != 1) {
+					throw new KalturaAPIException(KalturaErrors::INVALID_FIELD_VALUE, $kalturaProperty);
+				}
+				$this->$kalturaProperty = self::CHAR_APOSTROPHE . $this->$kalturaProperty;
```
inbal-ben-david, line 252: lets do this on all cases where the name starts with invalid char instead of throwing an error (without checking the length

inbal-ben-david (approved):

PR #11479, opened 2022-03-14
FOUN-414: modify "deploy_v2.php" script to work with kFile wrapper

amiras89:


amiras89:
approved by @yossipapi  on teams

PR #11482, opened 2022-03-15
FOUN-394: provision new esearch or beacon indices for elastic host

amiras89:


amiras89:
approved by @yossipapi over Teams

PR #11487, opened 2022-03-17
LIV-917: simulive - add support for playlist source

Amirhalfon1:
### playManifest changes:
in case of simulive with playlist source - override the servedType with MEDIA_CLIP to continue to the regular flow of simulive
### kSimuliveUtils:
in case of source of playlist type - create the json in accordance with all the playlist entries assets
**IMPORTANT:**
this PR will cause behavioral change for simulive. until now - if event total time was configured to be shorter than the sourceEntry - the source entry duration (only) in json was shorten to the event's total time. this PR changing it so if the event duration is shorter than the total time of all sources - the durations array will be transformed s.t the duration that causing the exceeding will be shorten to the remainder, and all the durations after will get the value of '1'.

### DIFF
--- a/alpha/apps/kaltura/lib/kSimuliveUtils.php
+++ b/alpha/apps/kaltura/lib/kSimuliveUtils.php
@@ -313,7 +313,6 @@ class kSimuliveUtils
 			$entriesFlavorAssets = array_merge($entriesFlavorAssets, array($mainFlavorAssets));
 			$entriesCaptionAssets = array_merge($entriesCaptionAssets, array($mainCaptionAssets));
 			$entriesAudioAssets = array_merge($entriesAudioAssets, array($mainAudioAssets));
-			$durations[] = self::roundDuration($srcEntry->getLengthInMsecs());
 		}
 		return array($entriesFlavorAssets, $entriesCaptionAssets, $entriesAudioAssets);
 	}
--- a/alpha/apps/kaltura/lib/kSimuliveUtils.php
+++ b/alpha/apps/kaltura/lib/kSimuliveUtils.php
@@ -37,27 +37,26 @@ class kSimuliveUtils
 		}
 		// all times should be in ms
 		$startTime = $currentEvent->getCalculatedStartTime() * self::SECOND_IN_MILLISECONDS;
-		$durations[] = self::roundDuration(min($sourceEntry->getLengthInMsecs(), ($currentEvent->getCalculatedEndTime() * self::SECOND_IN_MILLISECONDS) - $startTime));
 
-		list($mainFlavorAssets, $mainCaptionAssets, $mainAudioAssets) = myEntryUtils::getEntryAssets($sourceEntry);
+		$sourceEntries = $sourceEntry->getType() == entryType::PLAYLIST ? myPlaylistUtils::retrieveStitchedPlaylistEntries($sourceEntry) : array($sourceEntry);
 
 		// getting the preStart assets (only if the preStartEntry exists)
 		$preStartEntry = kSimuliveUtils::getPreStartEntry($currentEvent);
-		list($preStartFlavorAssets, $preStartCaptionAssets, $preStartAudioAssets) = myEntryUtils::getEntryAssets($preStartEntry);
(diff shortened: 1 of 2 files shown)


alpha/apps/kaltura/lib/kSimuliveUtils.php
```diff
@@ -302,4 +295,54 @@ public static function getClosestPlaybackTransitionTime($event, $time)
...
+		$entriesFlavorAssets = array();
+		$entriesCaptionAssets = array();
+		$entriesAudioAssets = array();
+		foreach ($sourceEntries as $srcEntry)
+		{
+			list($mainFlavorAssets, $mainCaptionAssets, $mainAudioAssets) = myEntryUtils::getEntryAssets($srcEntry);
+			$entriesFlavorAssets = array_merge($entriesFlavorAssets, array($mainFlavorAssets));
+			$entriesCaptionAssets = array_merge($entriesCaptionAssets, array($mainCaptionAssets));
+			$entriesAudioAssets = array_merge($entriesAudioAssets, array($mainAudioAssets));
+			$durations[] = self::roundDuration($srcEntry->getLengthInMsecs());
```
david-winder-kaltura, line 316: Why is this needed here?

alpha/apps/kaltura/lib/kSimuliveUtils.php
```diff
@@ -302,4 +295,54 @@ public static function getClosestPlaybackTransitionTime($event, $time)
...
+		$durations = array();
+		$eventDuration = $event->getCalculatedEndTime() * self::SECOND_IN_MILLISECONDS - $event->getCalculatedStartTime() * self::SECOND_IN_MILLISECONDS;
+		$aggregatedDuration = 0;
+		foreach ($sourceEntries as $srcEntry)
+		{
+			$entryRoundedDuration = self::roundDuration($srcEntry->getLengthInMsecs());
+			$aggregatedDuration += $entryRoundedDuration;
+			if ($aggregatedDuration >= $eventDuration)
+			{
+				$durations[] = max($entryRoundedDuration - ($aggregatedDuration - $eventDuration), 1); // 1 as 0 is not valid for the packager
```
david-winder-kaltura, line 340: No sure I understood this - as if (aggregatedDuration > eventDuration) then: $aggregatedDuration - $eventDuration is negative -> entryRoundedDuration - negative is big num (above 1) and I think we should set 1 in this case

Amirhalfon1: if "$aggregatedDuration > $eventDuration" is positive (the condition inside if) -> the subtraction "$aggregatedDuration - $eventDuration" is positive, and hence "entryRoundedDuration - POSITIVE" yields number smaller than entryRoundedDuration (or '0' or negative number which will yield '1')

david-winder-kaltura: I miss understood - then it is good

nadav-ravid (changes requested):

alpha/apps/kaltura/lib/kSimuliveUtils.php
```diff
@@ -302,4 +295,53 @@ public static function getClosestPlaybackTransitionTime($event, $time)
...
+	{
+		$entriesFlavorAssets = array();
+		$entriesCaptionAssets = array();
+		$entriesAudioAssets = array();
+		foreach ($sourceEntries as $srcEntry)
+		{
+			list($mainFlavorAssets, $mainCaptionAssets, $mainAudioAssets) = myEntryUtils::getEntryAssets($srcEntry);
+			$entriesFlavorAssets = array_merge($entriesFlavorAssets, array($mainFlavorAssets));
+			$entriesCaptionAssets = array_merge($entriesCaptionAssets, array($mainCaptionAssets));
+			$entriesAudioAssets = array_merge($entriesAudioAssets, array($mainAudioAssets));
```
nadav-ravid, line 315: why use array merge instead of just appending the value to the array?

nadav-ravid (approved):

PR #11489, opened 2022-03-20
PLAT-23634: minimizing returned values from db, to prevent memory exh…

danbar0kal:
…austion

### DIFF
--- a/plugins/vendor/zoom/lib/model/kZoomEventHanlder.php
+++ b/plugins/vendor/zoom/lib/model/kZoomEventHanlder.php
@@ -297,7 +297,7 @@ class kZoomEventHanlder
 	protected static function loadDropFolderFiles($dropFolderId, $fileName)
 	{
 		$statuses = array(KalturaDropFolderFileStatus::PARSED.','.KalturaDropFolderFileStatus::DETECTED);
-		$dropFolderFiles = self::retrieveByFolderIdOrderAndStatusesNotIn($dropFolderId, $statuses, $fileName);
+		$dropFolderFiles = self::retrieveByFolderIdOrderAndStatusesNotInAndName($dropFolderId, $statuses, $fileName);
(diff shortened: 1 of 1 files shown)


plugins/vendor/zoom/lib/model/kZoomEventHanlder.php
```diff
@@ -306,12 +307,13 @@ protected static function loadDropFolderFiles($dropFolderId)
 		return $dropFolderFilesMap;
 	}
 	
-	protected static function retrieveByFolderIdOrderAndStatusesNotIn($dropFolderId, $order, $statuses)
+	protected static function retrieveByFolderIdOrderAndStatusesNotIn($dropFolderId, $order, $statuses, $fileName)
 	{
 		$c = new Criteria();
 		$c->addAnd(DropFolderFilePeer::DROP_FOLDER_ID, $dropFolderId, Criteria::EQUAL);
 		$c->addAnd(DropFolderFilePeer::STATUS, $statuses, Criteria::NOT_IN);
 		$c->addAnd(DropFolderFilePeer::CREATED_AT, time() - dateUtils::DAY * 3, Criteria::GREATER_EQUAL);
```
inbal-ben-david, line 315: we dont want to limit to 3 days anymore. if we have the id we can search it without this timeframe

plugins/vendor/zoom/lib/model/kZoomEventHanlder.php
```diff
@@ -306,12 +307,13 @@ protected static function loadDropFolderFiles($dropFolderId)
...
+		$c->addAnd(DropFolderFilePeer::FILE_NAME, $fileName, Criteria::EQUAL);
 		$c->addAscendingOrderByColumn($order);
```
inbal-ben-david, line 317: ordering is less relevant and we can remove it from the function as we will have only one result from now on

PR #11498, opened 2022-03-22
SUP-31443 compare purifier output to decoded input string to consider case of escaping special chars

inbal-ben-david:


infra/general/kHtmlPurifier.class.php
```diff
@@ -13,6 +13,8 @@ class kHtmlPurifier
 	const ALLOWED_FRAME_TARGETS = 'allowedFrameTargets';
 	const ALLOWED_REL = 'allowedRel';
 	const HTML_DOCTYPE = "HTMLDoctype";
+	const ALLOW_ESCAPED = "allowEscaped";
```
inbal-ben-david, line 16: rename to ALLOWED_ESCAPED_SPECIEL_CHARS

PR #11501, opened 2022-03-24
PLAT-23655: Allowed Domain and Excluded Admin Role Are Not Copied Fro…

danbar0kal:
…m Template To New Partner


admin_console/controllers/PartnerController.php
```diff
@@ -135,6 +136,12 @@ public function createAction()
 		$this->view->form = $form;
 	}
 	
+	protected function setFieldsFromTemplate(& $config, $newPartner)
```
inbal-ben-david, line 139: &$config

lets do it only if we have templatePartnerId

PR #11505, opened 2022-03-27
PSVAMB-32925: use latest v2 player version when initial deploying

rgvsilva:
When initial installing/deploying an environment, HTML5 player from PID 99 gets assigned with v2.55.2 instead of {latest}


rgvsilva:
Hi @jessp01, could you please double check and bless my suggested change

PR #11512, opened 2022-03-28
FOUN-441: "static content" flow enhancement in case only a single asset (excluding the source) has been created

amiras89:


anatolkaltura (commented):
Amir
'isAssetDimBiggerThanSource' is miscorrect and misleading name.
the asset dims is not 'bigger' than source dims,
it is either equal to source dims or just slightly smaller, hence the percentage (5).

plz chg to suitable name, like 'isAssetDimsMatchSource'

yossipapi:
Reviwed.

PR #11524, opened 2022-03-30
PLAT-23665: Allow e-search users by company and title

danbar0kal:


alpha/lib/model/index/kuserIndex.php
```diff
@@ -73,6 +73,8 @@ public static function getIndexFieldsMap()
 				'permission_names' => 'indexedPermissionNames',
 				'role_ids' => 'indexedRoleIds',
 				'type' => 'type',
+				'company' => 'company',
```
yossipapi, line 76: Is this data needed to be indexed to sphinx as well? 
If not why was it added here?

PR #11533, opened 2022-04-07
FOUN-440: index "title" and "company" fields to kuser

amiras89:


GElkayam:
@amiras89 would this require re-indexing of the kuser index in elastic? if so please make sure to note that in release notes.

GElkayam:
Thanks @amiras89 , please also consider also some integration with the deployment scripts directory, as OnPrems / Regional Clouds and probably CE do not update the mapping on every deployment.
@jessp01 @john-hasty please keep me honest here.

PR #11534, opened 2022-04-07
docs: update api apptoken sessionprivileges

pnts-se:
Update description in API to be more aligned with
https://developer.kaltura.com/api-docs/VPaaS-API-Getting-Started/application-tokens.html

I've seen user try adding session privileges like so:
BASE_USER_SESSION_PERMISSION,WIDGET_SESSION_PERMISSION etc.

From the current API description I can see why they do that.

### DIFF
--- a/api_v3/lib/types/KalturaAppToken.php
+++ b/api_v3/lib/types/KalturaAppToken.php
@@ -85,7 +85,8 @@ class KalturaAppToken extends KalturaObject implements IFilterable
 	public $sessionDuration;
 
 	/**
-	 * Comma separated privileges to be applied on KS (Kaltura Session) that created using the current token
+	 * Privileges for KS created by this token. User role and privacy like so: setrole:<role id>,privacycontext:<privacy context label>
+	 *
 	 * @var string
 	 */
 	public $sessionPrivileges;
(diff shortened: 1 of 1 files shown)


jessp01:
Hi @pnts-se ,

I agree that we can add a better annotation for the `privileges/sessionPrivileges` members.
Since there are many different privileges and also seeing how there are several actions (of different services) that accept the `privileges` param and, given valid inputs, return a KS, I think we may be better off referencing a URI to where this is documented (for example, https://developer.kaltura.com/api-docs/VPaaS-API-Getting-Started/Kaltura_API_Authentication_and_Security.html). 

And, whatever course we choose, we best use the same annotation cross wide.
@erankor, @inbal-ben-david, @hilak - what do you think?

hilak:
> https://developer.kaltura.com/api-docs/VPaaS-API-Getting-Started/Kaltura_API_Authentication_and_Security.html

@jessp01  that sounds good to me, with the addition that perhaps we could add a hyperlink specifically to the privileges chapter on that page? UX would be better if a user upon landing in that page, didn't have to scroll through a lot of (potentially) irrelevant text to reach the section on privileges specifically.

pnts-se:
Ah I see.. 

setrole:<role id>,privacycontext:<privacy context label> is just an easy way to configure session privileges for appTokens. Not the only way.

The URL with an anchor is a good way to go.

PR #11547, opened 2022-04-12
SUP-29366: Fix restored entry button in admin console

OrelAri:


yossipapi:
@OrelAri why are you doing this change?

The validateEntryForRestoreDelete also checks entry & file sync statuses before allowing the restore and by removing you can potentially do a restore entry operation on an entry that is not deleted (I know UI blocks it but still).

OrelAri:
@yossipapi The button not working consistently on our admin console, this is a result of false return in the validation because at least one flavor/file sync is ready. 
When the button not working for the support, they open ticket for T3 to run restoring script, which in my opinion contain much more risks and no validation (only on entry entry status). 
https://github.com/kaltura/server/blob/Rigel-18.3.0/alpha/scripts/utils/restoreDeletedEntriesByEntryList.php
https://github.com/kaltura/server/blob/Rigel-18.3.0/alpha/scripts/utils/restoreDeletedEntries.php
I don't fully understand why we need the validation in the first place if we only changing the objects statuses. 
I thought about validation in the entry status, but as u said because the button is not accessible when entry is not deleted, I did not add this validation. 
Should I prepare different option for fix with better validation ? If so, I believe the restoring logic could be improved by us, so I will appreciate guidelines.

yossipapi:
So the real issue is that we have ready fileSyncs/assets on a deleted entry. Did you look into why this is happening?
As IIRC the event handling flow should delete all assets & file sync when deleting the entry as can be seen here:
[alpha/apps/kaltura/lib/kObjectDeleteHandler.php:84](https://github.com/kaltura/server/blob/c09d9a7a225257d532aaf58a0c3416dd1d81a8a4/alpha/apps/kaltura/lib/kObjectDeleteHandler.php#L84)

PR #11549, opened 2022-04-13
FOUN-450: add 'dynamic' = 'false' to elastic mappings to prevent dyna…

amiras89:
…mically adding new fields to elastic maps


amiras89:
Approved by @yossipapi

PR #11552, opened 2022-04-18
PLAT-23675: Recording Ingestion By Group - Event Workflow

danbar0kal:


plugins/vendor/zoom/lib/model/kZoomEventHanlder.php
```diff
@@ -55,6 +57,19 @@ public function processEvent($event)
 				if ($zoomDropFolderId)
 				{
 					$zoomClient = $this->initZoomClient($zoomVendorIntegration);
+					$hostId = $event->object->hostId;
```
inbal-ben-david, line 60: move the logic inside createZoomDropFolderFile so it will be inforced on transcripts as well

plugins/vendor/zoom/lib/model/kZoomEventHanlder.php
```diff
@@ -55,6 +57,19 @@ public function processEvent($event)
...
+					$zoomUser = $zoomClient->retrieveZoomUser($hostId);
+					$hostEmail = '';
+					if(isset($zoomUser[self::EMAIL]) && !empty($zoomUser[self::EMAIL]))
+					{
+						$hostEmail = $zoomUser[self::EMAIL];
+					}
+					$userId = $this->getEntryOwnerId($hostEmail);
+					if ($zoomVendorIntegration->shouldNotIngestRecording($userId))
+					{
+						KalturaLog::notice('The user is configured to not save recordings - Not processing');
```
inbal-ben-david, line 70: print the userId in the notice

plugins/vendor/zoom/lib/model/kZoomEventHanlder.php
```diff
@@ -92,6 +107,21 @@ public function processEvent($event)
 		}
 	}
 	
+	protected function getEntryOwnerId($hostEmail)
```
inbal-ben-david, line 110: this logic does not take into consideration the different options that can be set for zoom user
why not using the id on the object returns on getEntryOwner on kZoomProcessor instead?

plugins/vendor/zoom/lib/model/processors/kZoomWebinarProcessor.php
```diff
@@ -39,9 +39,11 @@ protected function parseAdditionalUsers($additionalUsersZoomResponse)
 			$result = array();
 			foreach ($panelistsEmails as $panelistEmail)
 			{
+				$accessToken = kZoomOauth::getValidAccessToken($this->zoomIntegration);
```
inbal-ben-david, line 42: we need to take the user from zoom only on cms case, on other cases we will just have another unneeded call to zoom.
lets create a function on the processor parent that also check the user mode type like it was in the original function

plugins/vendor/zoom/lib/model/ZoomVendorIntegration.php
```diff
@@ -119,11 +125,17 @@ public function addGroupNameToOptInGroups($v)
...
+		if ($v != $this->getGroupParticipationType())
+		{
+			$this->putInCustomData(self::GROUP_PARTICIPATION_TYPE, $v);
+			$this->putInCustomData(self::OPT_IN_GROUP_NAMES, '');
+			$this->putInCustomData(self::OPT_OUT_GROUP_NAMES, '');
+		}
+	}
+	public function getGroupParticipationType()
+	{
+		return $this->getFromCustomData( self::GROUP_PARTICIPATION_TYPE, null, kZoomGroupParticipationType::NO_CLASSIFICATION);
```
inbal-ben-david, line 137: fix space before self

PR #11565, opened 2022-05-02
PLAT-23686: Race condition between partner creation and KMS

drorsou:


api_v3/services/CategoryService.php
```diff
@@ -316,15 +316,19 @@ function indexAction($id, $shouldUpdate = true)
...
 			case kCoreException::DUPLICATE_CATEGORY:
 				throw new KalturaAPIException(KalturaErrors::DUPLICATE_CATEGORY, $categoryDb->getFullName());
 				
 			case kCoreException::PARENT_ID_IS_CHILD:
-				throw new KalturaAPIException(KalturaErrors::PARENT_CATEGORY_IS_CHILD, $category->parentId, $categoryDb->getId());
+				if (!isset($parentId))
+				{
+					$parentId = $category->parentId;
+				}
+				throw new KalturaAPIException(KalturaErrors::PARENT_CATEGORY_IS_CHILD, $parentId, $categoryDb->getId());
```
inbal-ben-david, line 331: this will result in a different category id as you are not passing the category from the template
maybe we can pass the new and original category instead of this change and take the full name from the original

api_v3/services/CategoryService.php
```diff
@@ -316,12 +316,19 @@ function indexAction($id, $shouldUpdate = true)
...
 	}
 	
-	private function handleCoreException(kCoreException $ex, category $categoryDb, KalturaCategory $category)
+	private function handleCoreException(kCoreException $ex, category $categoryDb, KalturaCategory $category, $originalCategoryDb = null)
 	{
 		switch($ex->getCode())
 		{
 			case kCoreException::DUPLICATE_CATEGORY:
-				throw new KalturaAPIException(KalturaErrors::DUPLICATE_CATEGORY, $categoryDb->getFullName());
+				if (isset($originalCategoryDb))
```
inbal-ben-david, line 324: better use something like
$name =  $categoryDb->getFullName();
if isset(originalCategoryDb)
     $name =
and then have one throw

PR #11570, opened 2022-05-06
FOUN-453: HF - offset must be >= 0

amiras89:


amiras89:
This was just HF in case of need, closing due to a more general fix that will be applied
https://github.com/kaltura/server/pull/11571/files

PR #11571, opened 2022-05-08
FOUN-453: prevent kPager to set incorrect values to pageIndex and pageSize

amiras89:


amiras89:
Approved by @yossipapi on a Teams chat

PR #11573, opened 2022-05-09
LIV-956: remove the universal akami live option

david-winder-kaltura:


inbal-ben-david (approved):

yossipapi:
If we are no longer supporting this option why not remove the permission and all classes related to it?

PR #11574, opened 2022-05-09
PLAT-23684: Enable TV Platform Studio (HTML5 Player V7) For every new…

danbar0kal:
… partner on KMC


danbar0kal:
handled via config

PR #11581, opened 2022-05-11
PLAT-23727: Zoom OPT IN/OUT - wrong call to DB for user groups

danbar0kal:


plugins/vendor/zoom/lib/model/ZoomVendorIntegration.php
```diff
@@ -185,7 +185,7 @@ public function shouldExcludeUserRecordingsIngest($userId)
 		{
 			return false;
 		}
-		$userGroupsArray = KuserKgroupPeer::retrievePgroupIdsByKuserIds($userId);
+		$userGroupsArray = KuserKgroupPeer::retrievePgroupIdsByPuserIds($userId);
```
inbal-ben-david, line 188: extract the kuserId from the partnerId + puserId and use the previous function since puserId is not a key on kuser_kgroup table and its less efficient

inbal-ben-david (approved):

PR #11589, opened 2022-05-13
SUP-31890: Adding support of override attribute via categoryUser bulkUpload

OrelAri:


api_v3/lib/types/KalturaBulkUploadResultCategoryUser.php
```diff
@@ -34,15 +34,21 @@ class KalturaBulkUploadResultCategoryUser extends KalturaBulkUploadResult
     * @var int
     */
    public $requiredObjectStatus;
+
+	/**
+	 * @var bool
+	 */
+	public $override;
```
inbal-ben-david, line 41: fix indentation

plugins/bulk_upload/csv/batch/BulkUploadCategoryUserEngineCsv.php
```diff
@@ -257,29 +255,38 @@ protected function createObjects()
...
+			$categoryUser->permissionLevel = $bulkUploadCategoryUserResult->permissionLevel;
+		}
+		
+		if (!is_null($bulkUploadCategoryUserResult->updateMethod))
+		{
+			$categoryUser->updateMethod = $bulkUploadCategoryUserResult->updateMethod;
+		}
+		else
+		{
+			$categoryUser->updateMethod = KalturaUpdateMethodType::MANUAL;
```
inbal-ben-david, line 286: why this logic was changed?
until now the default was automatic in case no value was set up and not manual

PR #11594, opened 2022-05-16
PLAT-23675: Fixing

danbar0kal:


plugins/vendor/zoom/lib/model/ZoomVendorIntegration.php
```diff
@@ -179,24 +179,26 @@ public function getGroupParticipationType()
 		return $this->getFromCustomData(self::GROUP_PARTICIPATION_TYPE, null, kZoomGroupParticipationType::NO_CLASSIFICATION);
 	}
 	
-	public function shouldExcludeUserRecordingsIngest($puserId)
+	public function shouldExcludeUserRecordingsIngest($userId)
```
inbal-ben-david, line 182: your function gets a non consistent value (sometimes string puser id and sometimes inte kuser id).
so your previous logic won't work and the current one won't work for all cases as well.
you need to return back the previous logic and change https://github.com/kaltura/server/blob/e6d76c6cadeaefb59955ba83160fe76035fe700c/plugins/vendor/zoom/lib/model/kZoomEventHanlder.php#L142 to get puser id like we do on zoom processor class

plugins/vendor/zoom/lib/model/kZoomEventHanlder.php
```diff
@@ -100,9 +100,11 @@ public function processEvent($event)
 		}
 	}
 	
-	protected static function shouldExcludeUserFromSavingRecording (kZoomEvent $event, $zoomClient, ZoomVendorIntegration $zoomVendorIntegration)
+	protected static function shouldExcludeUserFromSavingRecording ($event, $zoomClient, ZoomVendorIntegration $zoomVendorIntegration)
 	{
-		$hostEmail = $event->object->host_email;
+		/* @var kZoomRecording $object*/
+		$object = $event->object;
```
inbal-ben-david, line 106: rename to eventObject

PR #11599, opened 2022-05-25
SUP-30452 add the permission to the permissionname array

Adirv:


Adirv:
@inbal-ben-david fixed var name and I removed the block which was there by mitsake

PR #11602, opened 2022-05-25
FOUN-456: extend "country" max allowed letters + add support for "country" in esearch search

amiras89:


amiras89:
Approved by @yossipapi F2F

PR #11631, opened 2022-06-13
Rigel 18.7.0 liv 964 add vendor task

nadav-ravid:


nadav-ravid:
merged accidentally to previous version. previous PR with approval and comments:
https://github.com/kaltura/server/pull/11615

PR #11634, opened 2022-06-14
PLAT-23755: Upload flow is not working as expected (mediaType)

danbar0kal:


alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -176,6 +176,19 @@ public function uploadFileToToken($fileData, $resume = false, $resumeAt = -1)
 		$this->_uploadToken->setDc(kDataCenterMgr::getCurrentDcId());
 		
 		$this->_uploadToken->save();
+		
+		$flavorAsset = assetPeer::retrieveById($this->_uploadToken->getObjectId());
```
inbal-ben-david, line 180: this logic needs to run only if the upload was finished fully and the flavor asset is source and $this->_uploadToken->getObjectId()

api_v3/lib/KalturaEntryService.php
```diff
@@ -148,6 +148,10 @@ protected function replaceResourceByEntry($dbEntry, $resource, $tempMediaEntry)
 		$dbEntry->setReplacementStatus(entryReplacementStatus::NOT_READY_AND_NOT_APPROVED);
 		if(!$partner->getEnabledService(PermissionName::FEATURE_ENTRY_REPLACEMENT_APPROVAL) || $dbEntry->getSourceType() == EntrySourceType::KALTURA_RECORDED_LIVE)
 			$dbEntry->setReplacementStatus(entryReplacementStatus::APPROVED_BUT_NOT_READY);
+		if ($dbEntry->getMediaType() == null && $tempDbEntry->getMediaType() != null)
```
inbal-ben-david, line 151: check if this is already happening on 
api_v3/services/BaseEntryService.php:399

danbar0kal: does not happen. this is relevent only when uploading in the usual way, where the token is already uploaded and the type (field inside KalturaBaseEntry) is automatic. In our case, we asked Atar to preselect KalturaMediaEntry so we don't go in that piece of code.

alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -168,14 +170,56 @@ public function uploadFileToToken($fileData, $resume = false, $resumeAt = -1)
...
 		}
 		
 		$this->_uploadToken->setUploadedFileSize($fileSize);
 		$this->_uploadToken->setDc(kDataCenterMgr::getCurrentDcId());
 		
 		$this->_uploadToken->save();
+		
+		if ($this->_uploadToken->getObjectId() && $this->_uploadToken->getObjectType())
+		{
+			switch ($this->_uploadToken->getObjectType())
```
inbal-ben-david, line 184: move it to protected function getAttachedEntry

alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -168,14 +170,56 @@ public function uploadFileToToken($fileData, $resume = false, $resumeAt = -1)
...
+				{
+					$flavorAsset = assetPeer::retrieveById($this->_uploadToken->getObjectId());
+					if ($flavorAsset && $flavorAsset->getType() == assetType::FLAVOR && $flavorAsset->getEntryId())
+					{
+						$entry = entryPeer::retrieveByPK($flavorAsset->getEntryId());
+					}
+					break;
+				}
+			}
+			if ($entry && $entry->getType() == entryType::MEDIA_CLIP && !$entry->getMediaType())
```
inbal-ben-david, line 201: $entry is not necesseraly defined, declere it before the call to the new function you will create

inbal-ben-david (approved):

PR #11650, opened 2022-06-21
SUP-31486: get shorter expiry than 24 hours in session.impersonate

coralburg12:
Cover the cases : 

1. Parter has expiry null or non positive and expiry from user is null or non positive - return 24 hours as expiry.
2. Parter has expiry null or non positive and expiry from user is less than 24 hours but positive - return expiry from user.
3. Parter has expiry null or non positive and expiry from user is more than 24 hours - return 24 hours as expiry.
4. Partner has positive value for expiry and expiry from user is null or non positive - return partner's expiry.
5. Partner has positive value for expiry and expiry from user is positive - return the minimum between them.

Not handled for now - case 5 when both of expiries are bigger than 24 hours.

### DIFF
--- a/alpha/lib/model/Partner.php
+++ b/alpha/lib/model/Partner.php
@@ -129,7 +129,16 @@ class Partner extends BasePartner
 			in_array($partner_secret, $additionalSecrets, true) ||
 			(!$admin && $partner_secret === $this->getSecret()))
 		{
-			$ks_max_expiry_in_seconds = $this->getKsMaxExpiryInSeconds();
+			$ks_max_expiry_in_seconds_from_partner = $this->getKsMaxExpiryInSeconds();
+			if (!$ks_max_expiry_in_seconds_from_partner|| $ks_max_expiry_in_seconds_from_partner < 0){
+				$ks_max_expiry_in_seconds_from_partner = 86400;
+			}
+			if ($ks_max_expiry_in_seconds && $ks_max_expiry_in_seconds > 0) {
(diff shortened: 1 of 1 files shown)


alpha/lib/model/Partner.php
```diff
@@ -129,7 +129,13 @@ public function validateSecret ( $partner_secret , $partner_key , &$ks_max_expir
 			in_array($partner_secret, $additionalSecrets, true) ||
 			(!$admin && $partner_secret === $this->getSecret()))
 		{
-			$ks_max_expiry_in_seconds = $this->getKsMaxExpiryInSeconds();
+			$ks_max_expiry_in_seconds_from_partner = $this->getKsMaxExpiryInSeconds();
+			if (!$ks_max_expiry_in_seconds_from_partner) {
```
inbal-ben-david, line 133: i think that in that case we need to keep the old behavior and just return the $this->getKsMaxExpiryInSeconds() as it was. we only want to add an upper bound to the expiry in case it was set

alpha/lib/model/Partner.php
```diff
@@ -129,7 +129,13 @@ public function validateSecret ( $partner_secret , $partner_key , &$ks_max_expir
...
+				$ks_max_expiry_in_seconds_from_partner = 86400;
+			}
+			if (!$ks_max_expiry_in_seconds || $ks_max_expiry_in_seconds > $ks_max_expiry_in_seconds_from_partner) {
```
inbal-ben-david, line 136: move { to new lines on all if blocks

inbal-ben-david (approved):

PR #11668, opened 2022-07-03
Rigel-18.9.0-PLAT-23755-1

danbar0kal:


alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -117,6 +117,24 @@ public function getUploadToken()
 	 */
 	public function uploadFileToToken($fileData, $resume = false, $resumeAt = -1)
 	{
+		if (!$this->_uploadToken->getFileName())
```
inbal-ben-david, line 120: there is a function for that on line 148, no need to so it twice
$this->updateFileName($fileData);

alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -117,6 +117,24 @@ public function getUploadToken()
...
+		{
+			$this->_uploadToken->setFileName($fileData['name']);
+		}
+		$entry = $this->getAttachedEntry();
+		if ($entry && $entry->getType() == entryType::MEDIA_CLIP && !$entry->getMediaType())
+		{
+			$entry->setMediaType($this->getMediaType());
```
inbal-ben-david, line 127: change it to 2 phases get and then set so the entry wont indexed if we dont have to,
like we have on:
api_v3/lib/KalturaEntryService.php:444

alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -117,6 +117,24 @@ public function getUploadToken()
 	 */
 	public function uploadFileToToken($fileData, $resume = false, $resumeAt = -1)
 	{
+		if (!$this->_uploadToken->getFileName())
+		{
+			$this->_uploadToken->setFileName($fileData['name']);
+		}
+		$entry = $this->getAttachedEntry();
+		if ($entry && $entry->getType() == entryType::MEDIA_CLIP && !$entry->getMediaType())
```
inbal-ben-david, line 125: i think we can move all the logic back to the end of the function (the place where it was before this change)

alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -126,9 +126,34 @@ public function uploadFileToToken($fileData, $resume = false, $resumeAt = -1)
...
+		{
+			$mediaType = $this->getMediaType();
+			if ($mediaType)
+			{
+				$entry->setMediaType($this->getMediaType());
+				$entryUpdated = true;
+			}
+			
+			if($entry->getMediaType() == KalturaMediaType::IMAGE)
```
inbal-ben-david, line 143: you can use $mediaType var you already initialized and check
if($mediaType && $mediaType ==..)

alpha/apps/kaltura/lib/kUploadTokenMgr.php
```diff
@@ -126,9 +126,34 @@ public function uploadFileToToken($fileData, $resume = false, $resumeAt = -1)
...
+			{
+				$contentResource = new KalturaUploadedFileTokenResource();
+				$contentResource->token = $this->_uploadToken->getId();
+				$dbContentResource = $contentResource->toObject();
+				$dbContentResource->attachCreatedObject($entry);
+			}
+		}
+		
+		if($entryUpdated)
```
inbal-ben-david, line 152: do it before the image entry check like being done on kLocalResource

PR #11678, opened 2022-07-06
PLAT-23805: userScore/list doesnt work with game services created lea…

drorsou:
…derboards


plugins/game/lib/api/filters/KalturaUserScorePropertiesFilter.php
```diff
@@ -8,7 +8,7 @@
 class KalturaUserScorePropertiesFilter extends KalturaUserScorePropertiesBaseFilter
 {
 	/**
-	 * @var int
+	 * @var string
```
inbal-ben-david, line 11: 1. need to also change the api for update and delete
2. need to test with client to make sure we have backword compatibility with int

inbal-ben-david (approved):

PR #11681, opened 2022-07-07
PLAT-23768: Add validation for config.ini file in deployV2 script

drorsou:


deployment/uiconf/deploy_v2.php
```diff
@@ -31,6 +31,11 @@
 
 //error_reporting(0);
 $confObj = uiConfDeployment::init($arguments['ini']); // get and read the config file
+if (!(isset($confObj->general->component->name) && isset($confObj->general->component->version)))
```
inbal-ben-david, line 34: i think we better have it on a different function and check first if $confObj->general->component is set then check the inner fields, and also check that $confObj->player exists in the conf

PR #11693, opened 2022-07-13
SUP-30725: increase moderation count to be more then 1 when flag entry

coralburg12:


alpha/lib/model/entry.php
```diff
@@ -363,11 +363,14 @@ public function setStatusReady ( $force = false )
 	 */
 	public function setModerationStatus($v)
 	{
-		if($v == $this->getModerationStatus())
-			return $this;
 			
-		if($v == entry::ENTRY_MODERATION_STATUS_PENDING_MODERATION || $v == entry::ENTRY_MODERATION_STATUS_FLAGGED_FOR_REVIEW)
+		$moderationStatuses = array(entry::ENTRY_MODERATION_STATUS_PENDING_MODERATION, $v == entry::ENTRY_MODERATION_STATUS_FLAGGED_FOR_REVIEW);
```
inbal-ben-david, line 367: this resolves to 0/1, not to the enum. looks like copy error
$v == entry::ENTRY_MODERATION_STATUS_FLAGGED_FOR_REVIEW

alpha/lib/model/entry.php
```diff
@@ -363,11 +363,14 @@ public function setStatusReady ( $force = false )
...
+		if (in_array($v, $moderationStatuses)) {
```
inbal-ben-david, line 368: { always in new line

alpha/lib/model/entry.php
```diff
@@ -363,11 +363,14 @@ public function setStatusReady ( $force = false )
 	 */
 	public function setModerationStatus($v)
```
inbal-ben-david, line 364: how you decide whether we always need to increase the moderation count if we got status ENTRY_MODERATION_STATUS_PENDING_MODERATION/ENTRY_MODERATION_STATUS_FLAGGED_FOR_REVIEW is it true on all cases?

alpha/lib/model/entry.php
```diff
@@ -363,11 +363,16 @@ public function setStatusReady ( $force = false )
...
-		if($v == $this->getModerationStatus())
-			return $this;
 			
-		if($v == entry::ENTRY_MODERATION_STATUS_PENDING_MODERATION || $v == entry::ENTRY_MODERATION_STATUS_FLAGGED_FOR_REVIEW)
+		$moderationStatuses = array(entry::ENTRY_MODERATION_STATUS_PENDING_MODERATION, entry::ENTRY_MODERATION_STATUS_FLAGGED_FOR_REVIEW);
+		if (in_array($v, $moderationStatuses))
+		{
 			$this->incModerationCount();
+		}
+		elseif($v == $this->getModerationStatus())
```
inbal-ben-david, line 372: should be in a different if since both conditions can be fullfiled

inbal-ben-david (approved):

PR #11699, opened 2022-07-18
PLAT-23808: session->impersonate doesnt allow setting ks expiry longe…

drorsou:
…r than max partner expiry

### DIFF
--- a/api_v3/services/SessionService.php
+++ b/api_v3/services/SessionService.php
@@ -90,8 +90,9 @@ class SessionService extends KalturaBaseService
 	{
 		KalturaResponseCacher::disableCache();
 		
+		$validatedExpiry = $expiry;
 		// verify that partnerId exists and is in correspondence with given secret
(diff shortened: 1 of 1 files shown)


api_v3/services/SessionService.php
```diff
@@ -90,8 +90,9 @@ function impersonateAction($secret, $impersonatedPartnerId, $userId = "", $type
 	{
 		KalturaResponseCacher::disableCache();
 		
+		$validatedExpiry = $expiry;
 		// verify that partnerId exists and is in correspondence with given secret
-		$result = myPartnerUtils::isValidSecret($partnerId, $secret, "", $expiry, $type);
+		$result = myPartnerUtils::isValidSecret($partnerId, $secret, "", $validatedExpiry, $type);
 		if ($result !== true)
```
inbal-ben-david, line 96: i think we also need to add check like we do on
alpha/apps/kaltura/lib/webservices/kSessionUtils.class.php:77

inbal-ben-david (approved):

PR #11705, opened 2022-07-25
LIV-958: Add new SCHEDULE_EVENT_FEATURE_MANAGE for the update live fe…

david-winder-kaltura:
…ature action


david-winder-kaltura:
User role is not update due to insertDefault is only for new object. Update the role should be done manually by alpha-script/sql command

nadav-ravid (approved):

PR #11708, opened 2022-07-27
FOUN-579: add "registrationFormSchema" JSON-Schema value to virtualEvent object

amiras89:


plugins/virtual_event/lib/model/VirtualEvent.php
```diff
@@ -97,4 +98,22 @@ public function setDeletionDueDate($v)
...
+	 
+	 public function getRegistrationFormSchema()
+	 {
+		 return $this->getFromCustomData(self::CUSTOM_DATA_FIELD_REGISTRATION_FORM_SCHEMA);
+	 }
+
+	 public function setRegistrationFormSchema($v)
+	 {
+		 // validate $v is a valid JSON object
+		 json_decode($v);
```
yossipapi, line 110: I would do this validation in the validateForinsert phase and not here.
Once you do this you won't need the kVirtualEventException.php class 

also I think the check should b something lile

$json = json_decode($string);
$isValid =  (is_object($json) && json_last_error() == JSON_ERROR_NONE) ? true : false;

plugins/virtual_event/services/VirtualEventService.php
```diff
@@ -28,15 +28,27 @@ public function initService($serviceId, $serviceName, $actionName)
...
 	 */
 	public function addAction(KalturaVirtualEvent $virtualEvent )
 	{
-		/* @var $dbVirtualEvent VirtualEvent */
-		$this->validateScheduleEvents($virtualEvent);
-		$this->validateGroups($virtualEvent);
-		$dbVirtualEvent = $virtualEvent->toInsertableObject();
-		$dbVirtualEvent->setPartnerId($this->getPartnerId());
-		$dbVirtualEvent->save();
+		try
```
yossipapi, line 36: Once you do the fix I mentioned in setRegistrationFormSchema you won't need this try catch here as well as you will directly throw an API exceptions if the string is not a valid json

plugins/virtual_event/services/VirtualEventService.php
```diff
@@ -89,10 +102,20 @@ public function updateAction($id, KalturaVirtualEvent $virtualEvent)
...
 		
-		// save the object
-		/** @var VirtualEvent $dbVirtualEvent */
-		$dbVirtualEvent = $virtualEvent->toUpdatableObject($dbVirtualEvent);
-		$dbVirtualEvent->save();
+		try
+		{
+			// save the object
+			/** @var VirtualEvent $dbVirtualEvent */
+			$dbVirtualEvent = $virtualEvent->toUpdatableObject($dbVirtualEvent);
```
yossipapi, line 109: Same comment as above 
Validate the json string is valid in validateForUpdate if it was sent on the virtualEvent object.

yossipapi (changes requested):

yossipapi (approved):

PR #11716, opened 2022-07-31
FOUN-596: Move entryVendorTask->add / update to use general "kString::checkIsValidJson()

amiras89:
…alidJson()"


yossipapi (approved):
Approved.

PR #11724, opened 2022-08-02
add check to confirm task data exists when required

nadav-ravid:


plugins/reach/lib/api/data/entryVendorTaks/KalturaEntryVendorTask.php
```diff
@@ -342,6 +343,23 @@ private function validateEntryId()
 			}
 		}
 	}
+
+	private function validateCatalogLimitations()
```
inbal-ben-david, line 347: Use Protected

plugins/reach/lib/api/data/entryVendorTaks/KalturaEntryVendorTask.php
```diff
@@ -342,6 +343,23 @@ private function validateEntryId()
...
+	{
+		$vendorCatalogItem = VendorCatalogItemPeer::retrieveByPK($this->catalogItemId);
+		//currently a param for simplicity should be made a const with more complex requirement options
+		$featureDataRequirements = array(VendorServiceFeature::LIVE_CAPTION);
```
inbal-ben-david, line 351: The name of the var is not clear to me

plugins/reach/lib/api/data/entryVendorTaks/KalturaEntryVendorTask.php
nadav-ravid: used private to be consistent with the other internal validate function

PR #11738, opened 2022-08-08
fixed an edge case that would cause an internal server error

nadav-ravid:


nadav-ravid:
When creating a new entryVendorTask with scheduled vendor task data. an internal server error would be raised if the provided scheduled event id doesn't exist due to one of the checks related to the event running before the scheduled event validation

inbal-ben-david (approved):

PR #11743, opened 2022-08-09
SUP-33280 change loginDataResetPasswordAction to return the user with…

salma2222:
… encryptedSeed instead of returning void

### DIFF
--- a/api_v3/services/UserService.php
+++ b/api_v3/services/UserService.php
@@ -399,7 +399,11 @@ class UserService extends KalturaBaseUserService
 	public function loginDataResetPasswordAction($loginDataId, $newPassword)
 	{
 		self::validateLoginDataParams(array('id' => $loginDataId));
-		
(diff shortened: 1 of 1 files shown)


api_v3/services/UserService.php
```diff
@@ -417,7 +417,9 @@ public function loginDataResetPasswordAction($loginDataId, $newPassword)
 			}
 			throw $e;
 		}
-		return $updateLoginData;
+
+		$user = KuserPeer::getKuserByEmail($loginDataId, kCurrentContext::$partner_id);
```
inbal-ben-david, line 421: i think it will make more sense to extract the user before we update the login data, and also throw exception like we do in enable login in case it is not found

salma2222: Done

inbal-ben-david (approved):

PR #11746, opened 2022-08-10
SUP-30103: get userId from the request only on admin session otherwise…

salma2222:
… get the userId from ks


inbal-ben-david (approved):
reviewed and fixed offline

PR #11751, opened 2022-08-15
LIV-1004: use alternate duration value for live entry vendor tasks

nadav-ravid:
Still needs to be tested

### DIFF
--- a/plugins/reach/lib/kReachManager.php
+++ b/plugins/reach/lib/kReachManager.php
@@ -482,7 +482,8 @@ class kReachManager implements kObjectChangedEventConsumer, kObjectCreatedEventC
 			$pendingEntryReadyTask->setStatus($newStatus);
 			$pendingEntryReadyTask->setAccessKey(kReachUtils::generateReachVendorKs($pendingEntryReadyTask->getEntryId(), $pendingEntryReadyTask->getIsOutputModerated(), $pendingEntryReadyTask->getAccessKeyExpiry()));
 			if($pendingEntryReadyTask->getPrice() == 0)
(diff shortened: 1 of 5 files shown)


plugins/reach/lib/kReachManager.php
```diff
@@ -482,7 +482,8 @@ protected function checkPendingEntryTasks($object)
 			$pendingEntryReadyTask->setStatus($newStatus);
 			$pendingEntryReadyTask->setAccessKey(kReachUtils::generateReachVendorKs($pendingEntryReadyTask->getEntryId(), $pendingEntryReadyTask->getIsOutputModerated(), $pendingEntryReadyTask->getAccessKeyExpiry()));
 			if($pendingEntryReadyTask->getPrice() == 0)
-				$pendingEntryReadyTask->setPrice(kReachUtils::calculateTaskPrice($object, $pendingEntryReadyTask->getCatalogItem()));
+				$taskDuration =  $pendingEntryReadyTask->getTaskJobData()->getEntryDuration();
```
inbal-ben-david, line 485: 1. need to wrap the lines with {}
2. remove extra space after =
3. what if getTaskJobData() is null?

plugins/reach/lib/kReachManager.php
```diff
@@ -602,7 +603,8 @@ private function handleEntryDurationChanged(entry $entry)
 		{
 			/* @var $pendingEntryVendorTask EntryVendorTask */
 			$oldPrice = $pendingEntryVendorTask->getPrice();
-			$newPrice = kReachUtils::calculateTaskPrice($entry, $pendingEntryVendorTask->getCatalogItem());
+			$taskDuration = $pendingEntryVendorTask->getTaskJobData()->getEntryDuration();
```
inbal-ben-david, line 606: you might access null pointer if job data is null

inbal-ben-david (approved):

PR #11777, opened 2022-08-24
FOUN-627: Add logs for sphinx field to detect unbalanced " and uneven \ at EOL

amiras89:


yossipapi (approved):

jessp01:
@amiras89 - better use `KalturaLog::err()` rather than `KalturaLog::log()` for these, IMHO.

PR #11785, opened 2022-08-30
PSVAMB-37517: Change log message from notice to error level

rgvsilva:
In the situation sphinx is not able to update a record, a NOTICE message is throw, even though it is an error so depending on loglevel, the message is not displayed and the issue is missed.


GElkayam (approved):
LGTM, Approved by @yossipapi

PR #11803, opened 2022-09-06
PLAT-23842: Permit partner -16 to use groupUser.list

avichainoach:


release-notes.md
```diff
@@ -1,5 +1,12 @@
 # Rigel-18.14.0
 
+## Add groupUse list permission for game service partner ##
```
inbal-ben-david, line 3: typo - groupUser
also makes more sense to put this section under the add partner -16 section

PR #11813, opened 2022-09-14
No-Plat: Add check for flavor param id

david-winder-kaltura:


nadav-ravid (approved):

david-winder-kaltura:
Currently saw on QA id 35,36,37 were not return for the DB query - will check it

inbal-ben-david (approved):

david-winder-kaltura:
As this is not failing the code, and the one sending the kLiveStreamParam is kaltura liveNG system - we can skip this check and save performance

PR #11832, opened 2022-09-29
Override flavorParams in case of live-caption feature

david-winder-kaltura:


alpha/apps/kaltura/modules/extwidget/actions/playManifestAction.class.php
```diff
@@ -1143,6 +1143,23 @@ private function serveLiveEntry()
 		if(!$this->deliveryProfile)
 			return null;
 
+		// temporary workaround till adding caption as asset for the live entry
+		if ($this->entry->hasCapability(LiveEntry::LIVE_FEATURE_CAPABILITY))
+		{
+			KalturaLog::info("Entry has capability of schedule live feature on : " . $this->entry->getId());
+			$events = $this->entry->getScheduleEvents(time(), time() + 3600);
```
david-winder-kaltura, line 1150: we can consider remove the SQL query and just set redirect and empty array for flavorsParams when the capability of live feature is set (consider the fact it is just workaround)

PR #11833, opened 2022-09-29
PLAT-23924: Fix PHP notice when array string index is not defined

danbar0kal:


alpha/apps/kaltura/lib/multiLingualUtils.php
```diff
@@ -132,7 +132,7 @@ public static function addFieldMappingToMultiLangMapping(&$multiLingualMapping,
 		{
 			foreach ($valueToAdd as $languageKey => $languageValue)
 			{
-				if ($multiLingualMapping[$field][$languageKey] != $languageValue)
+				if (!isset($multiLingualMapping[$field][$languageKey]) || $multiLingualMapping[$field][$languageKey] != $languageValue)
```
inbal-ben-david, line 135: for sure we have $multiLingualMapping[$field]?
if not we will still get notice in the next line

danbar0kal: i think that if it doesnt exist, the next line creates it. the next line doest lookup the value it sets it

PR #11844, opened 2022-10-11
PLAT-23928: Disable password restriction on FirstName/LastName/Email

danbar0kal:


alpha/lib/model/UserLoginDataPeer.php
```diff
@@ -271,6 +274,16 @@ public static function checkPasswordValidation($newPassword, $loginData)
 		}
 	}
 
+	protected static function getPartnerByLoginData($loginData)
+	{
+		if (!$loginData) {
```
inbal-ben-david, line 279: { in new line

PR #11854, opened 2022-10-23
PLAT-23939 - Return the vertical on partner.get

danbar0kal:


api_v3/lib/types/partner/KalturaPartner.php
```diff
@@ -469,6 +469,12 @@ class KalturaPartner extends KalturaObject implements IFilterable
 	 */
 	public $eventPlatformAllowedTemplates;
 	
+	/**
+	 * @var string
+	 * @readonly
+	 */
+	public $verticalClassificationName;
```
inbal-ben-david, line 476: didnt we agreed that this will be a number?
why the name contisn 'Name' and why is it a string?

api_v3/lib/types/partner/KalturaPartner.php
```diff
@@ -475,6 +475,12 @@ class KalturaPartner extends KalturaObject implements IFilterable
 	 */
 	public $eventPlatformAllowedTemplates;
 	
+	/**
+	 * @var int
+	 * @readonly
+	 */
+	public $verticalClassificationId;
```
inbal-ben-david, line 482: remove id and check for the var type in system partner configuration class

inbal-ben-david (approved):

PR #11863, opened 2022-10-26
PLAT-23909 - BE does not respond in case of captions that contain non…

danbar0kal:
…-UTF-8 content


plugins/content/caption/base/lib/captionManagers/srtCaptionsContentManager.php
```diff
@@ -15,7 +15,11 @@ public function parse($content)
 	{
 		if (kString::beginsWith($content, "\xff\xfe"))
 		{
-			$content = iconv('utf-16', 'utf-8', substr($content, 2));
+			$content = iconv('utf-16', 'utf-8//IGNORE', substr($content, 2));
```
inbal-ben-david, line 18: silent all iconv function calls with @
change all utf-8 string to lower

inbal-ben-david (approved):

PR #11865, opened 2022-10-27
PLAT-23950 - Expose flag skip password restriction of first/last name on response

danbar0kal:
…-UTF-8 content

### DIFF
--- a/api_v3/lib/types/partner/KalturaPartner.php
+++ b/api_v3/lib/types/partner/KalturaPartner.php
@@ -469,6 +469,12 @@ class KalturaPartner extends KalturaObject implements IFilterable
 	 */
 	public $eventPlatformAllowedTemplates;
 	
+	/**
+	 * Is password restriction feature on FirstName/LastName/Email enabled
+	 * @var bool
+	 */
+	public $passwordRestrictionDisabled;
+	
 	private static $map_between_objects = array
 	(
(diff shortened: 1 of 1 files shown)


api_v3/lib/types/partner/KalturaPartner.php
```diff
@@ -469,6 +469,12 @@ class KalturaPartner extends KalturaObject implements IFilterable
 	 */
 	public $eventPlatformAllowedTemplates;
 	
+	/**
+	 * Is password restriction feature on FirstName/LastName/Email enabled
+	 * @var bool
+	 */
+	public $passwordRestrictionDisabled;
```
inbal-ben-david, line 476: put this option around line 441 next to the other password options
if no one is going to change it from the api add @readonly
no need for the descriptoin
on api better have it in the oposite direction - name like $allowDeafultPasswordRestrictions set to true by deafult

api_v3/lib/types/partner/KalturaPartner.php
```diff
@@ -506,6 +512,7 @@ public function doFromObject($partner, KalturaDetachedResponseProfile $responseP
 		$this->adminName = kString::stripUtf8InvalidChars($this->adminName);
 		$this->describeYourself = kString::stripUtf8InvalidChars($this->describeYourself);
 		$this->additionalParams = KalturaKeyValueArray::fromKeyValueArray($partner->getAdditionalParams());
+		$this->passwordRestrictionDisabled = $partner->getEnabledService(PermissionName::FEATURE_DISABLE_PASSWORD_RESTRICTION);
```
inbal-ben-david, line 515: opposite logic will be clearer on api

inbal-ben-david (approved):

PR #11872, opened 2022-11-05
PLAT-23957:add support of external links in admin console 

MosheMaorKaltura:


plugins/admin_console/kaltura_application_links/admin/SelfserveAdminAction.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+class SelfserveAdminAction extends KalturaApplicationPlugin
+{
+    public function __construct()
+    {
+        $this->rootLabel = "External apps";
```
inbal-ben-david, line 7: the root label is not something related to all external apps?
why it should be defined in the action explicitly? if someone will add additional app action we will have multiple tabs with the same root name?

PR #11879, opened 2022-11-08
PLAT-23953:"Enforce HTTPS API" is not enforced on session.start

danbar0kal:


alpha/apps/kaltura/lib/request/kSessionBase.class.php
```diff
@@ -338,6 +338,37 @@ static public function getSecretsFromCache($partnerId)
...
+	{
+		$cacheSections = kCacheManager::getCacheSectionNames(kCacheManager::CACHE_TYPE_PARTNER_SECRETS);
+		
+		if(!$cacheSections)
+		{
+			return null;
+		}
+		$cacheKey = self::getSecretsCacheKey($partnerId);
+		KalturaLog::debug("Deleting key [$cacheKey] now [" . date('Y-m-d H:i:s', time()) . "]");
+		$deleted = false;
```
inbal-ben-david, line 352: I don't see any real meaning to this flag, if you were able to delete the value only in one of the stores why should it be true?
also you don't really check the return value of delete so you don't really know if it succeeded or failed
in my opinion you can remove the flag, you also don't use its return value

danbar0kal: you are correct.

danbar0kal: should we use the return statements and issue an error in case we were not able to delete in a store?

alpha/apps/kaltura/lib/webservices/kSessionUtils.class.php
```diff
@@ -223,6 +223,40 @@ public static function killKSession ( $ks )
...
+				continue;
+			}
+			$secrets = $cacheStore->get($cacheKey);
+			if (!$secrets)
+			{
+				continue;
+			}
+			if ($cacheStore->delete($cacheKey))
+			{
+				KalturaLog::debug("query took " . (microtime(true) - $queryStart) . " seconds [$cacheSection, $cacheKey] - Deletion successful");
```
inbal-ben-david, line 252: remove 'Deletion successful'

alpha/apps/kaltura/lib/webservices/kSessionUtils.class.php
```diff
@@ -223,6 +223,40 @@ public static function killKSession ( $ks )
...
+			}
+			else
+			{
+				KalturaLog::warning("query took " . (microtime(true) - $queryStart) . " seconds [$cacheSection, $cacheKey] - Failed to delete");
```
inbal-ben-david, line 256: err level
start with "Failed to delete [$key] from [$section], query took..

PR #11880, opened 2022-11-09
PLAT-23970: add support in studio v7

MosheMaorKaltura:


inbal-ben-david (approved):
need to add release notes

PR #11885, opened 2022-11-10
SUP-34629 change the fromDate and toDate params to bigint instead of …

Adirv:
…time to geenrate correct types in libraries


api_v3/lib/types/report/KalturaReportInputBaseFilter.php
```diff
@@ -8,14 +8,14 @@ class KalturaReportInputBaseFilter extends KalturaObject
 	/**
 	 * Start date as Unix timestamp (In seconds)
 	 * 
-	 * @var time
+	 * @var bigint
```
inbal-ben-david, line 11: this was already changed from int to time on this pull
https://github.com/kaltura/server/commit/0a5d3c3a95ad08b2ddf779c64ae0748564ed01e3
dont see a reason to change again

PR #11887, opened 2022-11-13
Rigel 18.18.0-FOUN-321

amiras89:


amiras89:
TESTS:
1. Upload small file (25 MB)	√
2. Upload large file (143 MB)	√
3. Fail 403 / 404 HTTP Error	√ - consider not getting partialDownload and FileSize when 4xx or 5xx http?
4. Test Axel errors:
	 4.1.url over 1024 bytes	√
	 4.2. state file not found	
5. Resume - better check on test server √
6. Security:
	6.1. "file://" import	√ - Error: Unsupported protocol (entry id: 0_08qyesin - on prem)
	6.2. If user:pass exist on URL - don't use axel	√	
	6.3. If URL redirect - don't use axel √
7. Check multiple huge files √
	Uploaded 3 * 32 GB files + 3 * 18 GB files - memory never got higher than 1GB
	Disk space was at ~80% before importing the files to S3 - but that's not a real use case

amiras89:
Approved by @yossipapi and @erankor on a Teams chat

PR #11898, opened 2022-11-27
FOUN-321: phase II modifications

amiras89:


amiras89:
Approved by Eran on teams chat

PR #11945, opened 2022-12-20
PLAT-24027: Handle pagination on responses from Webex

drorsou:


plugins/vendor/webex/batch/processors/KWebexAPIDropFolderEngine.php
```diff
@@ -34,11 +34,18 @@ class KWebexAPIDropFolderEngine extends KDropFolderFileTransferEngine
 	public function watchFolder(KalturaDropFolder $dropFolder)
 	{
 		$this->initDropFolderEngine($dropFolder);
-		$recordingsList = $this->retrieveRecordingsList();
-		if ($recordingsList)
+		$paginationLink = null;
```
inbal-ben-david, line 37: i would call it nextPageLink (unless its a fixed one which i assume its not the case)

plugins/vendor/webex/batch/processors/KWebexAPIDropFolderEngine.php
```diff
@@ -34,11 +34,18 @@ class KWebexAPIDropFolderEngine extends KDropFolderFileTransferEngine
...
 	{
 		$this->initDropFolderEngine($dropFolder);
-		$recordingsList = $this->retrieveRecordingsList();
-		if ($recordingsList)
+		$paginationLink = null;
+		do
 		{
-			$this->handleRecordingsList($recordingsList);
+			$recordingsList = $this->retrieveRecordingsList($paginationLink);
+			$paginationLink = $this->webexClient->getPaginationLinkFromLastRequest();
```
inbal-ben-david, line 41: makes more sense to flip the order of the lines and first handle the requests then take the next link

plugins/vendor/webex/batch/processors/KWebexAPIDropFolderEngine.php
```diff
@@ -34,11 +34,18 @@ class KWebexAPIDropFolderEngine extends KDropFolderFileTransferEngine
 	public function watchFolder(KalturaDropFolder $dropFolder)
 	{
 		$this->initDropFolderEngine($dropFolder);
```
inbal-ben-david, line 36: i think that we also need to reset the paginationLink on the client class when we init new folder, to make sure there is no case we use a previously set pagainationLink

inbal-ben-david (approved):

PR #11950, opened 2022-12-21
LIV-1087: remove underscore from host in broadcast url

david-winder-kaltura:
Underscore in host can cause problem in dns-resolve.
For entry starting with '1_' - we just remove the prefix
For all other entries we just replace the underscore with '-'

### DIFF
--- a/alpha/apps/kaltura/lib/live/kBroadcastUrlManager.php
+++ b/alpha/apps/kaltura/lib/live/kBroadcastUrlManager.php
@@ -93,7 +93,7 @@ class kBroadcastUrlManager
 		return $partner->getLiveStreamBroadcastUrlConfigurations($dc);
 	}
 
-	private static function getLiveIdForHost(BaseEntry $entry)
+	protected static function getLiveIdForHost($entry)
 	{
 		$entryId = str_replace('1_', '', $entry->getId());
 		return str_replace('_', '-', $entryId); // dns resolve don't handle well underscore in host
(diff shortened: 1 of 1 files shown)


nadav-ravid (approved):

nadav-ravid (approved):

alpha/apps/kaltura/lib/live/kBroadcastUrlManager.php
```diff
@@ -92,13 +92,19 @@ protected function getConfiguration ($dc = null)
 		$partner = PartnerPeer::retrieveByPK($this->partnerId);
 		return $partner->getLiveStreamBroadcastUrlConfigurations($dc);
 	}
+
+	private static function getLiveIdForHost(BaseEntry $entry)
```
inbal-ben-david, line 96: use protected
also no need to declare on BaseEntry

alpha/apps/kaltura/lib/live/kBroadcastUrlManager.php
```diff
@@ -92,13 +92,19 @@ protected function getConfiguration ($dc = null)
...
+	{
+		$entryId = str_replace('1_', '', $entry->getId());
```
inbal-ben-david, line 98: can we have old entries with 0_ on this case?

david-winder-kaltura: Yes - in the '0_' case the underscore will be replaced with '-'

inbal-ben-david (approved):

PR #11970, opened 2023-01-04
PLAT-23990: Add extended audio description as a Reach service

danbar0kal:


plugins/reach/admin/forms/CatalogItemConfigure.php
```diff
@@ -126,6 +126,23 @@ public function init()
 			));
 		}
 		
+		if ($this->catalogItemType == Kaltura_Client_Reach_Enum_VendorServiceFeature::EXTENDED_AUDIO_DESCRIPTION)
```
inbal-ben-david, line 129: if both dubbing, audio, and audio description have the same elements the why not having in array condition and write them once instead of duplicating

plugins/reach/lib/model/enums/VendorServiceFeature.php
```diff
@@ -5,12 +5,13 @@
...
-	const DUBBING               = 7;
-	const LIVE_CAPTION          = 8;
+	const CAPTIONS                      = 1;
+	const TRANSLATION                   = 2;
+	const ALIGNMENT                     = 3;
+	const AUDIO_DESCRIPTION             = 4;
+	const CHAPTERING                    = 5;
+	const INTELLIGENT_TAGGING           = 6;
+	const DUBBING                       = 7;
+	const EXTENDED_AUDIO_DESCRIPTION    = 8;
```
inbal-ben-david, line 15: you cant insert the enum before another one which already in use and recognized in a specific way

inbal-ben-david (approved):

PR #12017, opened 2023-01-22
SUP-34618: ignore not allowed chars when encoding xml

coralburg12:


infra/general/kString.class.php
```diff
@@ -271,6 +271,11 @@ public static function removeHttp($url)
 
 	public static function xmlEncode($str)
 	{
+		$notAllowedChars = kConf::get("not_allowed_chars", "runtime_config", array());
```
inbal-ben-david, line 274: call it chars_not_allowed

inbal-ben-david (approved):

PR #12056, opened 2023-02-15
No-Plat: Add object cache for local thumb capturing

yossipapi:


yossipapi:
No longer relevant, closing!

PR #12064, opened 2023-02-23
LIV-1123: reverted stream name change due to ux issue

nadav-ravid:


nadav-ravid:
opened at david ben shoshans request

inbal-ben-david (approved):

PR #12080, opened 2023-03-12
PLAT-24171: Clear the Recycle Bin

drorsou:


batch/batches/Delete/Engines/KDeletingEntryEngine.php
```diff
@@ -0,0 +1,64 @@
...
+		$entriesList = KBatchBase::$kClient->baseEntry->listAction($filter, $this->pager);
+		if (!$entriesList->objects || !count($entriesList->objects))
+		{
+			return 0;
+		}
+		
+		$entriesToDeletePerRequest = KBatchBase::$taskConfig->params->entriesToDeletePerRequest;
+		$waitBetweenRequestsInSeconds = KBatchBase::$taskConfig->params->waitBetweenRequestsInSeconds;
+		
+		$entriesCount = 0;
```
inbal-ben-david, line 36: worth changing to something more meaningful like handled or deleted entries count
also whats the difference between countResults and entriesCount? looks like they should be similar at the end

batch/batches/Delete/Engines/KDeletingEntryEngine.php
```diff
@@ -0,0 +1,64 @@
...
+	 * @return int the number of deleted entries
+	 */
+	protected function deleteEntries(KalturaBaseEntryFilter $filter)
+	{
+		if (!$filter->orderBy)
+		{
+			$filter->orderBy = KalturaBaseEntryOrderBy::UPDATED_AT_ASC;
+		}
+		
+		$entriesList = KBatchBase::$kClient->baseEntry->listAction($filter, $this->pager);
```
inbal-ben-david, line 27: what if you have multiple pages of results? i don't see this being handled

batch/batches/Delete/Engines/KDeletingEntryEngine.php
```diff
@@ -0,0 +1,64 @@
...
+		
+		$entriesToDeletePerRequest = KBatchBase::$taskConfig->params->entriesToDeletePerRequest;
+		$waitBetweenRequestsInSeconds = KBatchBase::$taskConfig->params->waitBetweenRequestsInSeconds;
+		
+		$entriesCount = 0;
+		$countResults = 0;
+		while ($entriesCount < count($entriesList->objects))
+		{
+			KBatchBase::$kClient->startMultiRequest();
+			for ($i = $entriesCount; $i < min($entriesCount + $entriesToDeletePerRequest, count($entriesList->objects)); $i++)
```
inbal-ben-david, line 41: this is not fully clear and seems like it will work only from one page results of base entry list
maybe worth simplifying it and just do paginations on 30 page size filter multiple times and create multi request for each one of them

batch/batches/Delete/Engines/KDeletingEntryEngine.php
```diff
@@ -0,0 +1,78 @@
...
+			foreach ($results as $index => $result)
+			{
+				if (is_array($result) && isset($result['code']))
+				{
+					unset($results[$index]);
+				}
+			}
+			$numberOfHandledEntries += count($results);
+			
+		} while (count($entriesList->objects) >= $this->pager->pageSize);
```
inbal-ben-david, line 69: shouldnt this be the total count? as the number of objects will always be the page size or less

batch/batches/Delete/Engines/KDeletingEntryEngine.php
```diff
@@ -0,0 +1,78 @@
...
+		
+		if (!$numberOfHandledEntries)
```
inbal-ben-david, line 71: you init it to 0 so not need for this if, just return numberOfHandledEntries value

inbal-ben-david (approved):

PR #12088, opened 2023-03-15
Scorpius 19.5.0 psvamb 42269

TalDubovKaltura:


plugins/event_notification/providers/http/lib/KDispatchHttpNotificationEngine.class.php
```diff
@@ -182,13 +184,18 @@ public function sendHttpRequest(KalturaHttpNotificationTemplate $httpNotificatio
...
+		$headers = substr($results, 0, $header_size);
+		$body = strip_tags(substr($results, $header_size));
+
 		$curlWrapper->close();
 
-		KalturaLog::info("HTTP Request httpCode [" . $httpCode . "] Results [$results]");
+        	KalturaLog::info("HTTP Request httpCode [" . $httpCode . "] Results [$results] Headers [$headers] Body [$body]");
 		if(!$results || $httpCode != 200)
 		{
+            $		body = substr($body, 0, 1000);
```
inbal-ben-david, line 196: you have tons of spaces between $ and body param

inbal-ben-david, line 196: put 1000 in const for max error string size

plugins/event_notification/providers/http/lib/KDispatchHttpNotificationEngine.class.php
```diff
@@ -182,13 +184,18 @@ public function sendHttpRequest(KalturaHttpNotificationTemplate $httpNotificatio
 		$errCode = $curlWrapper->getErrorNumber();
 		$errMessage = $curlWrapper->getError();
 
+		$header_size = curl_getinfo($curlWrapper->ch, CURLINFO_HEADER_SIZE);
+		$headers = substr($results, 0, $header_size);
```
inbal-ben-david, line 188: since you are adding CURLOPT_RETURNTRANSFER if this the headers will already be stored in infra/general/KCurlWrapper.class.php:64

plugins/event_notification/providers/http/lib/KDispatchHttpNotificationEngine.class.php
TalDubovKaltura: Fixed. https://github.com/kaltura/server/pull/12088/commits/bd7bfa77710425308a8cdf27bccf007d16a4d987

TalDubovKaltura: Done. https://github.com/kaltura/server/pull/12088/commits/bd7bfa77710425308a8cdf27bccf007d16a4d987

plugins/event_notification/providers/http/lib/KDispatchHttpNotificationEngine.class.php
TalDubovKaltura: Actually I already checked this in my debugging. In this case, we don't have the $headers because we need to request to store them(using the method here -KCurlWrapper.class.php:847), and the only method using it is getHeader()(KCurlWrapper.class.php:453) but in our case, we need also the body in the result. In order to do so I needed to change some methods in  KCurlWrapper and I decided not to touch so sensitive area which I can do but here it's only info we need to log so logic change will be overkill(I think). what do you think?

TalDubovKaltura:
been merged with a new PR https://github.com/kaltura/server/pull/12110

PR #12136, opened 2023-04-18
Scorpius 19.7.0 recycle bin auto delete

inbal-ben-david:


inbal-ben-david:
all approvals are in the inner pr's

PR #12142, opened 2023-04-24
PLAT-24229: Add new recycle bin auto delete processor

drorsou:


plugins/scheduled_task/lib/processors/KRecycleBinProcessor.php
```diff
@@ -0,0 +1,121 @@
...
+	/**
+	 * @param $partnerId
+	 * @return MultiRequestSubResult
+	 */
+	protected function getPartnerDaysBeforeDelete($partnerId)
+	{
+		KBatchBase::impersonate($partnerId);
+		$partner = KBatchBase::$kClient->partner->get($partnerId);
+		KBatchBase::unimpersonate();
+		return $partner->daysBeforeRecycleBinDeletion;
```
inbal-ben-david, line 66: if partner returned empty (and not with an error which will be caught from the upper try block) you will access null pointer
In this case we also need to throw an error as the number of days can falsely return as 0

plugins/scheduled_task/lib/processors/KRecycleBinProcessor.php
```diff
@@ -0,0 +1,121 @@
...
+	}
+	
+	protected function runSearch($pageIndex, $daysBeforeDelete)
```
inbal-ben-david, line 69: I would use a more indicative name since we use it only for recycle bin, something like ger overdueRecycledEntries

plugins/scheduled_task/lib/processors/KRecycleBinProcessor.php
```diff
@@ -0,0 +1,121 @@
...
+	{
+		$pager = new KalturaPager();
+		$pager->pageIndex = $pageIndex;
+		$pager->pageSize = 50;
+		$range = new KalturaESearchRange();
+		$range->lessThanOrEqual = time() - kTimeConversion::DAYS * $daysBeforeDelete;
```
inbal-ben-david, line 75: make sure this doesn't returns entries that don't have recycle_at at all or have null value

plugins/scheduled_task/lib/processors/KRecycleBinProcessor.php
```diff
@@ -0,0 +1,121 @@
...
+		{
+			return array();
+		}
+		foreach ($results as $index => $result)
+		{
+			if (is_array($result) && isset($result['code']))
+			{
+				unset($results[$index]);
+			}
+		}
```
inbal-ben-david, line 113: need to add a check for rate limit error, and if so sleep a short while

plugins/scheduled_task/batch/ScheduledTaskRunner/KScheduledTaskRunner.class.php
```diff
@@ -143,6 +158,12 @@ protected function getSortedScheduledTaskProfiles($maxProfiles = 500)
 		$filter->lastExecutionStartedAtLessThanOrEqualOrNull = strtotime('today UTC');
 		$pager = new KalturaFilterPager();
 		$pager->pageSize = $maxProfiles;
+		
+		$runnerTypes = $this->getAdditionalParams(KScheduledTaskRunner::runnerTypes);
```
inbal-ben-david, line 162: this is not on the template file
do we need both runnerTypes and params.recyclebin? looks overlapping

inbal-ben-david (commented):
we will also need to have release notes

inbal-ben-david (approved):

PR #12172, opened 2023-05-15
PLAT-24238: Add permission to block new direct login users

drorsou:


admin_console/forms/PartnerConfiguration.php
```diff
@@ -591,6 +592,11 @@ public function init()
 			'decorators' => array('ViewHelper', array('Label', array('placement' => 'append')), array('HtmlTag',  array('tag' => 'dt', 'class' => 'partner_configuration_checkbox_field_only')))
 		));
 		
+		$this->addElement('checkbox', 'ALLOW_SSO_PER_USER', array(
```
inbal-ben-david, line 595: will prefer to have this in lowercase like the rest of the element names

admin_console/forms/PartnerConfiguration.php
```diff
@@ -625,6 +631,10 @@ public function init()
 	
 			foreach($moduls as $name => $modul)
 			{
+				if ($modul->skip)
```
inbal-ben-david, line 634: if you skip you ignore the default decorator?

release-notes.md
```diff
@@ -1,3 +1,19 @@
+# Scorpius-19.10.0
+## Add permission to block new direct login users ##
```
inbal-ben-david, line 2: i would rephrase it to Add permission to allow setting SSO per user
like the permission name you gave

admin_console/forms/PartnerConfiguration.php
drorsou: Because the permission name on admin.ini is all caps, then here it is all caps.
So it will be an exception either be here as caps, or on admin.ini as lowercase.

inbal-ben-david (approved):

PR #12198, opened 2023-05-29
PLAT-24232: Static playlist missing entries after partner cloning

drorsou:


batch/batches/CopyPartner/KAsyncCopyPartner.class.php
```diff
@@ -12,6 +12,8 @@ class KAsyncCopyPartner extends KJobHandlerWorker
 	
 	protected $fromPartnerId;
 	protected $toPartnerId;
+	protected $entryIdsMap = array();
```
inbal-ben-david, line 15: init const array is supported from php 5.6 and on, there can still be on prems using older version so let's init the consts on line 55/58 and only declare them here

PR #12200, opened 2023-05-30
PLAT-24266: Empty the Recycle Bin when disabling it

drorsou:


configurations/plugins.template.ini
```diff
@@ -145,3 +145,4 @@ Interactivity
 ;KalturaApplicationLinks
 ;Room
 VirtualEvent
+;SessionCuePoint
```
inbal-ben-david, line 148: we better put this in a different PR, its unrelated to RB

inbal-ben-david:
dont merge it for now

PR #12214, opened 2023-06-06
PLAT-24294: accessing null pointer in multilingual array

danbar0kal:


alpha/apps/kaltura/lib/multiLingualUtils.php
```diff
@@ -261,7 +261,7 @@ protected static function getMappedLanguages($newMultiLingualMapping, $supported
 			}
 		}
 		
-		return array_unique($mappedLanguages)[0];
+		return empty($mappedLanguages) ? null : array_unique($mappedLanguages, SORT_REGULAR)[0];
```
inbal-ben-david, line 264: since we expect array to return makes more sense to return array() rather than just null if the $mappedLanguages array is empty

inbal-ben-david (approved):

PR #12224, opened 2023-06-11
Support de-registering batch server

yossipapi:


batch/KGenericBatchMgr.class.php
```diff
@@ -24,11 +24,19 @@ function shutDown($signal)
...
+	echo "Got signal [$signal] from terminal";
+	$kscheduler->deRegisterHost($signal);
+}
+
 //Windows machines by default do not have the pcntl installed, so check if function exists before calling it
 if(function_exists("pcntl_signal"))
 {
 	pcntl_signal(SIGINT, 'gracefulShutDown');
 	pcntl_signal(SIGTERM, 'shutDown');
+	pcntl_signal(SIGUSR1, 'deRegisterHost');
```
nivshlomokaltura, line 39: Do we wanna move to SIGUSR2? so we will keep USR1 for logcompressor?

api_v3/services/BatchService.php
```diff
@@ -672,4 +672,26 @@ public function putFileAction($destPath, $data)
...
+	 * @param string $host
+	 * @return bool
+	 */
+	public function deRegisterSchedulerAction($host)
+	{
+		KalturaLog::debug("Purging scheduler data for [$host]");
+		$scheduler = SchedulerPeer::getScheudlerByHostName($host);
+		if(!$scheduler)
+		{
+			KalturaLog::debug("Scheduler not found for host, [$host]");
```
nivshlomokaltura, line 689: Do we wanna extend error handling? for timeout, 5xx?

batch/KGenericBatchMgr.class.php
yossipapi: But logcompressor is a different process so not sure its needed.

PR #12265, opened 2023-07-05
PLAT-24249: clip concat child job

danbar0kal:


alpha/apps/kaltura/lib/clipconcat/kClipManager.php
```diff
@@ -123,13 +123,30 @@ protected function isConcatOfAllChildrenDone(BatchJob $batchJob)
 				KalturaLog::info('Child job id [' . $job->getId() . '] status [' . $job->getStatus() . ']' . '] type ['.$job->getJobType() .']' );
 				if($job->getStatus() != BatchJob::BATCHJOB_STATUS_FINISHED)
 				{
-					return false;
+					return $this->hasGrandChildFinished($job);
```
inbal-ben-david, line 126: we used to return here false if one of the children was not finished (it meant that not all concat done)
now to return true if one of the once that doesnt finished has a son which finished was finished but what if others were not? its a change in the logic

inbal-ben-david (approved):

PR #12275, opened 2023-07-12
PLAT-24353: Add KalturaSessionCuePointFilter

drorsou:


plugins/cue_points/session_cue_point/lib/api/filters/KalturaSessionCuePointFilter.php
```diff
@@ -0,0 +1,33 @@
...
+ 	 */
+	protected function getCoreFilter()
+	{
+		return new SessionCuePointFilter();
+	}
+	
+	/* (non-PHPdoc)
+	 * @see KalturaRelatedFilter::validateForResponseProfile()
+	 */
+	public function validateForResponseProfile()
```
inbal-ben-david, line 20: this should be empty like on other cue point filter types, the only one that uses isPublicEqual is annotation cue point which actually has this field on the object

plugins/cue_points/session_cue_point/lib/model/filters/SessionCuePointFilter.php
```diff
@@ -0,0 +1,22 @@
...
+ * @package plugins.sessionCuePoint
+ * @subpackage model.filters
+ */
+class SessionCuePointFilter extends CuePointFilter
+{
+	public function init()
+	{
+		parent::init();
+		
+		$extendedFields = kArray::makeAssociativeDefaultValue(array(
```
inbal-ben-david, line 13: no need to implement this and override the parent function, all the fields you need exist on CuePointFilter

inbal-ben-david (approved):

PR #12284, opened 2023-07-18
PLAT-24347: Send email when user updates isAdmin 0 to 1

drorsou:


alpha/lib/model/kuser.php
```diff
@@ -201,7 +201,12 @@ public function postUpdate(PropelPDO $con = null)
 		if ($this->isColumnModified(kuserPeer::LOGIN_DATA_ID)) {
 			$oldLoginDataId = $this->oldColumnsValues[kuserPeer::LOGIN_DATA_ID];
 		}
-
+		
+		if ($this->isColumnModified(kuserPeer::IS_ADMIN))
```
inbal-ben-david, line 205: what if the value was changed from true to false?

inbal-ben-david (approved):

PR #12289, opened 2023-07-20
Support stop file for periodic workers to allow quicker pod shutdown

yossipapi:


amiras89 (approved):
looks good 
FYI @inbal-ben-david (if you have any feedback)

PR #12295, opened 2023-07-26
LIV-1104: Add region to ServerNode

zoharsKaltura:


zoharsKaltura:
The new validation will fail only if there is an existing stream in different region, and not in any other case (to be backwards compatible).

david-winder-kaltura (approved):

api_v3/services/LiveStreamService.php
```diff
@@ -192,6 +194,47 @@ function authenticateAction($entryId, $token, $hostname = null, $mediaServerInde
...
+
+			//currently verifying only if already streaming in another environment
+			if ($mediaServerNode && $mediaServerNode->getEnvironment() != $registeredServerNode->getEnvironment()) {
+				$streamAlreadyExsits = true;
+			}
+
+		}
+		catch(Exception $e)
+		{
+			return;
```
inbal-ben-david, line 229: if there was error you are not interested in what it was?

inbal-ben-david (approved):

PR #12322, opened 2023-08-17
SUP-38438: esearch results should be same in kmcng and in exportToCsv

coralburg12:


batch/batches/ExportCsv/Engine/KEntryExportEngine.php
```diff
@@ -13,7 +13,14 @@ protected function getFilterOrderBy()
 
 	protected function getItemList($filter, $pager)
 	{
-		return KBatchBase::$kClient->baseEntry->listAction($filter, $pager);
+		$clientTag = KBatchBase::$kClient->getClientTag();
+		if(is_object($filter->advancedSearch))
+		{
+			KBatchBase::$kClient->setClientTag($clientTag ." useESearch");
+		}
+		$items = KBatchBase::$kClient->baseEntry->listAction($filter, $pager);
```
inbal-ben-david, line 21: need to add try catch and if failed to reset the previous clientTag else we will be stuck with the modified one for the rest of the process

inbal-ben-david (approved):

PR #12333, opened 2023-08-29
modify corpAction redirect from http to https + move schema to config…

amiras89:
… rather than hardcoded

### DIFF
--- a/alpha/apps/kaltura/modules/corp/actions/actions.class.php
+++ b/alpha/apps/kaltura/modules/corp/actions/actions.class.php
@@ -15,17 +15,17 @@ class corpActions extends sfActions
    */
   public function executeIndex()
   {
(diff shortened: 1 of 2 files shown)


GElkayam:
@amiras89 I suggest adding a default to the kConf::get calls, or at least adding to local.ini.template (which will mean more work down the road)

inbal-ben-david:
> @amiras89 I suggest adding a default to the kConf::get calls, or at least adding to local.ini.template (which will mean more work down the road)

+1, i think we better have it in the template as well

amiras89:
Good catch @GElkayam @inbal-ben-david 
With all the 'server-saas-config', 'server-saas-config-nvd1', 'server-saas-config-generic' I forgot about the template 🙁

Also added default value to keep the code 'backward compatible' (thought since we have no more new 'on-prems' I feel it's irrelevant - but, better be on the safe side)

Thanks

inbal-ben-david (approved):

PR #12351, opened 2023-09-07
Add UserAppRole Service - rebased

amiras89:


api_v3/lib/types/KalturaUserAppRole.php
```diff
@@ -0,0 +1,121 @@
...
+	public function toInsertableObject($kuserToUserRole = null, $skip = array())
+	{
+		if (is_null($kuserToUserRole))
+		{
+			$kuserToUserRole = new KuserToUserRole();
+		}
+		
+		$kuserToUserRole = parent::toInsertableObject($kuserToUserRole, $skip);
+		
+		/* @var KuserToUserRole $kuserToUserRole */
```
yossipapi, line 34: IMHO you dont really need this as the parent toInsertableObject already builds the object for you

api_v3/lib/types/KalturaUserAppRole.php
```diff
@@ -0,0 +1,121 @@
...
+		{
+			return false;
+		}
+		
+		$this->userId = kuserPeer::retrieveByPK($kuserToUserRoleObject->getKuserId())->getPuserId();
+		
+		parent::doFromObject($kuserToUserRoleObject, $responseProfile);
+	}
+	
+	private function verifyMandatoryParams($puserId, $appGuid ,$userRoleId)
```
yossipapi, line 99: Maybe try annotating the param with minLength 1 
I think it will remove the need of using this function

api_v3/lib/types/KalturaUserAppRole.php
```diff
@@ -0,0 +1,121 @@
...
+	{
+		if (is_null($kuserToUserRole))
+		{
+			$kuserToUserRole = new KuserToUserRole();
+		}
+		
+		$kuserToUserRole = parent::toInsertableObject($kuserToUserRole, $skip);
+		
+		/* @var KuserToUserRole $kuserToUserRole */
+		$kuserToUserRole->setKuserId($this->kuserId);
```
yossipapi, line 35: You can do this by creating a function on the core object called setKuser id mapping userId to kuserId and the setter will do this for you.

api_v3/lib/types/filters/KalturaUserAppRoleFilter.php
```diff
@@ -0,0 +1,167 @@
...
+	 * @param KalturaFilterPager $pager
+	 * @param KalturaDetachedResponseProfile|null $responseProfile
+	 * @return KalturaListResponse
+	 * @throws PropelException
+	 */
+	public function getListResponse(KalturaFilterPager $pager, KalturaDetachedResponseProfile $responseProfile = null)
+	{
+		$response = new KalturaUserAppRoleListResponse();
+		
+		if (in_array(kCurrentContext::getCurrentSessionType(), array(kSessionBase::SESSION_TYPE_NONE,kSessionBase::SESSION_TYPE_WIDGET)))
```
yossipapi, line 49: Check if this is needed.
If you enforce this in the permissions for the service it should be blocked

api_v3/lib/types/filters/KalturaUserAppRoleFilter.php
```diff
@@ -0,0 +1,167 @@
...
+		
+		if ($srcObj->get('_in_user_id'))
+		{
+			$this->userIdIn = $this->prepareKusersToPusersFilter($srcObj->get('_in_user_id'));
+		}
+	}
+	
+	protected function fixCsvFilterProperties()
+	{
+		$this->userIdIn = !empty($this->userIdIn) ? kString::csvFixWhitespace($this->userIdIn) : null;
```
yossipapi, line 163: If you want to trim the space after the comma better do explode and array map with trim

api_v3/services/UserAppRoleService.php
```diff
@@ -0,0 +1,232 @@
...
+	 * @param $actionName
+	 * @return void
+	 * @throws KalturaAPIException
+	 */
+	public function initService($serviceId, $serviceName, $actionName)
+	{
+		parent::initService($serviceId, $serviceName, $actionName);
+		
+		// Add Partner Ids to criteria for the following peers during execution of that service
+		$this->applyPartnerFilterForClass('UserRole');
```
yossipapi, line 23: Double-check other object that you access as part of this service if you need to apply the filter on them as well

configurations/cache.ini
```diff
@@ -98,3 +98,4 @@ playKitJsSourceMap = filesystemPlayKitJsSourceMap
 uploadToken = memcacheGlobal
 reportscount = memcacheGlobal
 authCachedData = memcacheGlobal
+microservices = memcacheLocal
```
yossipapi, line 101: Should be memcacheGlobal

yossipapi (changes requested):

api_v3/services/UserAppRoleService.php
amiras89: No need
We call it in parent::initService at:
https://github.com/kaltura/server/blob/cc0a987b70fc98d4620f2f3ab73d8c0ae1910e45/api_v3/lib/KalturaBaseService.php#L176

which will get to
https://github.com/kaltura/server/blob/cc0a987b70fc98d4620f2f3ab73d8c0ae1910e45/alpha/apps/kaltura/lib/myPartnerUtils.class.php#L181-L180

Where we have 'kuser' included

api_v3/lib/KalturaErrors.php
```diff
@@ -819,5 +819,15 @@ class KalturaErrors extends APIErrors
 	const HOST_NAME_CONTAINS_ASTRIX = "HOST_NAME_CONTAINS_ASTRIX;HOST_NAME;Host name contains *, use # instead {@HOST_NAME@}";
 	const CHANGE_DESCRIPTION_CANNOT_BE_EMPTY = "CHANGE_DESCRIPTION_CANNOT_BE_EMPTY;;Param changeDescription cannot be empty";
 	const SEARCH_ITEM_TYPE_NOT_FOUND = 'SEARCH_ITEM_TYPE_NOT_FOUND;SEARCH_ITEM_TYPE,ELASTIC_FIELD_NAME; Search item type [@SEARCH_ITEM_TYPE@] not found for field: [@ELASTIC_FIELD_NAME@]';
+	
+	/**
```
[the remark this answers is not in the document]
amiras89: I believe that all the classes that extend the KalturaErrors are plugins
Since this is a core api service, not sure where to place the extended class (unlike plugin which is placed in the correlative plugin folder)

api_v3/lib/types/KalturaUserAppRole.php
amiras89: Was looking into it, not sure it's gonna work
Because for updateAction userId and appGuid should be null (not updatable)

So it's only legible for 'validateForInsert'

Was thinking about 'validateForUsage' but it's being called at "toObject" and will be included for both add and update actions

I think the current function is the more elegant
We can go 'get_class_vars' and make sure non of the properties are null, but don't think it's better.

alpha/lib/model/KuserToUserRolePeer.php
```diff
@@ -46,12 +61,34 @@ public function getRootObjects(IRelatedObject $object)
...
+	 * @param $appGuid
+	 *
+	 * @return KuserToUserRole|null
+	 * @throws PropelException
+	 */
+	public static function getByKuserIdAndAppGuid($kuserId, $appGuid)
+	{
+		$c = new Criteria();
+		$c->addAnd(self::KUSER_ID, $kuserId, Criteria::EQUAL);
+		$c->addAnd(self::APP_GUID, $appGuid, Criteria::EQUAL);
```
yossipapi, line 86: Since you disable the default criteria please add the partnerId to the query as well

api_v3/services/UserAppRoleService.php
```diff
@@ -0,0 +1,196 @@
...
+	 * @param KalturaUserAppRoleFilter $filter
+	 * @param KalturaFilterPager $pager
+	 * @return KalturaUserAppRoleListResponse
+	 *
+	 * @throws KalturaAPIException
+	 * @throws PropelException
+	 */
+	public function listAction(KalturaUserAppRoleFilter $filter = null, KalturaFilterPager $pager = null)
+	{
+		if (!$filter || !($filter->userIdEqual || $filter->userIdIn || $filter->userRoleIdEqual || $filter->userRoleIdIn || $filter->appGuidEqual || $filter->appGuidIn))
```
yossipapi, line 115: I dont follow this limitation 
I think there are a few options here:

1. A request was made with user KS and you can only get a list of items that are specific to your user in any other case a user KS requests this for another user an exception should be thrown.

2. The request is with admin Ks and that you can provide an empty filter to fetch all.

You can say we have a limitation that in this service you can only do list if you either specify appGuid or userId in any other case fail it but this should be agreed with product.

deployment/permissions/service.userapprole.ini
```diff
@@ -0,0 +1,45 @@
+[action_permission_items]
```
yossipapi, line 1: General comments
1. why do we need batch & admin console impersonate allowed here?
2. Why don't we allow EP impersonated calls? do we need this?

alpha/lib/model/KuserToUserRole.php
```diff
@@ -17,7 +17,45 @@ class KuserToUserRole extends BaseKuserToUserRole {
...
+		}
+		
+		parent::setKuserId($kuser->getId());
+	}
+	
+	public function getPuserId()
+	{
+		$kuserId = $this->getKuserId();
+		
+		// todo consult Yossi - when 'validateForUpdate' executes it calls the getter functions before we call setter functions
```
yossipapi, line 39: Remove private comments if you you already handled them

alpha/apps/kaltura/lib/microservice/MicroServiceAppRegistry.php
```diff
@@ -4,13 +4,156 @@
...
+		if (!count($appGuidsToQuery))
+		{
+			return $appGuidsCsv;
+		}
+		
+		$filter = array(
+			'idIn' => $appGuidsToQuery
+		);
+		
+		$pager = array(
```
yossipapi, line 94: Did you check if we allow large APIs if, for example, we get 10000 ids?
And if so how the MS will behave in such a case?

alpha/lib/model/KuserToUserRole.php
amiras89: I did not, wanted your opinion
But I take if you are good with the solution ;)

alpha/lib/model/KuserToUserRolePeer.php
amiras89: We do not have partnerId on this bridge table.
But kuserId + appGuid is an index - so should be OK.

api_v3/services/UserAppRoleService.php
amiras89: 1. If we get a user ks, I override any userEqual or userIn and set the user from the ks on the filter, here:
https://github.com/kaltura/server/blob/3c0061350fcba6ef155e36f57ceb2f2cc14b927c/api_v3/lib/types/filters/KalturaUserAppRoleFilter.php#L94

That way, listAction always return a valid response (either 0 or > 0)

2. We cannot do that because there is not PID on that bridge table, so I must get at least one object_id that I can verify belongs to the PID

deployment/permissions/service.userapprole.ini
amiras89: 1. No need, removed
2. EP can already do session -> impersonate
so they make api requests via the impersonated ks

alpha/apps/kaltura/lib/microservice/MicroServiceAppRegistry.php
amiras89: Generally speaking, EP will send a single appGuid most times (per EP site since appGuid represent an EP instance)
I do not believe that the appGuid is going to be the major bottleneck

amiras89: @yossipapi spoke with Daniel S. the dev working on it
He confirmed the use case is going to be 2 appGuids most time

This we should do something else here?
I can open a separate ticket to do a small 'load' test of sending 1000 appGuids

Not sure we will ever get 10000

PR #12442, opened 2023-11-06
PLAT-24380: Encrypt file names for doc conversion

david-winder-kaltura:


plugins/content/document/lib/operation_engines/KOperationEngineDocument.php
```diff
@@ -56,10 +64,29 @@ protected static function jsonFormat($arr, $basePath)
...
+	}
+
+	protected static function encryptFileName($basePath, $fileName, $key) {
+		if (!$key )
+		{
+			return $fileName;
+		}
+		try {
+			$pathinfo = pathinfo ($fileName);
+			$newName = rtrim(base64_encode(hash_hmac('sha256', $pathinfo['filename'], $key)), "=") . '.' . $pathinfo['extension'];
```
kalturaguy, line 82: for shorter string maybe it's better to do 
rtrim(base64_encode(hash_hmac('sha256', 'docMetadata', "", true)), "=");

(pass true at the last character)

plugins/content/document/lib/operation_engines/KOperationEngineDocument.php
```diff
@@ -56,10 +64,30 @@ protected static function jsonFormat($arr, $basePath)
...
+	}
+
+	protected static function encryptFileName($basePath, $fileName, $key) {
+		if (!$key )
+		{
+			return $fileName;
+		}
+		try {
+			$pathinfo = pathinfo ($fileName);
+			$encryptName = base64_encode(hash_hmac('sha256', $pathinfo['filename'], $key, true));
```
inbal-ben-david, line 82: i think we better make sure is really set before accessing $pathinfo['filename'] just in case, same for extension

plugins/content/document/lib/operation_engines/KOperationEngineDocument.php
```diff
@@ -56,10 +64,30 @@ protected static function jsonFormat($arr, $basePath)
...
+		}
+	}
+
+	protected static function encryptFileName($basePath, $fileName, $key) {
+		if (!$key )
+		{
+			return $fileName;
+		}
+		try {
+			$pathinfo = pathinfo ($fileName);
```
inbal-ben-david, line 81: extra space before (

plugins/content/document/lib/operation_engines/KOperationEngineImageMagick.php
```diff
@@ -124,17 +126,21 @@ protected function operationComplete($rc, $output) {
...
 	// information about the created images.
-	private function createImagesListXML($imagesList){
+	private function createImagesListXML($imagesList, $outDirPath, $key){
 		sort($imagesList);
+		$i = 1;
 		$imagesListXML = new SimpleXMLElement('<'.self::IMAGES_LIST_XML_LABEL_ITEMS.'/>');
 		foreach ($imagesList as $image) {
     		$imageNode = $imagesListXML->addChild(self::IMAGES_LIST_XML_LABEL_ITEM);
-    		$imageNode->addChild(self::IMAGES_LIST_XML_LABEL_NAME, $image);
+		$imageNode->addAttribute(self::IMAGES_LIST_XML_ATTRIBUTE_INDEX, $i++);
```
inbal-ben-david, line 137: why do we need this attribute?

david-winder-kaltura: To make sure we have alignment between the metadata file and the imageList file. As before, we could compare the file by name - but if the name encrypted - we should have indexes to align them

plugins/content/document/lib/operation_engines/KOperationEngineDocument.php
david-winder-kaltura: Well, this wrap in try-catch block, so if we failing I would expect to not encrypt the files name at all.
I have no objection to add protection if you want. Do you want me to check for exist before passing on?

plugins/content/document/lib/operation_engines/KOperationEngineDocument.php
david-winder-kaltura: fixed

inbal-ben-david (approved):

PR #12447, opened 2023-11-07
VCP-15238: add support for template files when running 'insertDefaults.php'

amiras89:


deployment/base/scripts/insertDefaults.php
```diff
@@ -283,5 +293,28 @@ function getVodPackagerUrl()
 	return $vodPackagerUrl;
 }
 
+function isSupportedTemplateFile($dirName)
+{
+	echo("File [$dirName] does not exist - checking if filename support template file\n");
```
yossipapi, line 298: Use KalturaLog not echo

deployment/base/scripts/insertDefaults.php
```diff
@@ -283,5 +293,28 @@ function getVodPackagerUrl()
...
+	$supportedTemplateFilenames = array(
+		'01\.Partner',
+		'07\.DeliveryProfile'
+	);
+	
+	$fileName = basename($dirName);
+	
+	foreach ($supportedTemplateFilenames as $supportedTemplateFilename)
+	{
+		if (preg_match("/^$supportedTemplateFilename\w*\.ini/", $fileName) === 1)
```
yossipapi, line 310: Why not add \w*\.ini to the supportedTemplateFilenames I think it will be much cleaner that way

amiras89: I agree

yossipapi (approved):

PR #12455, opened 2023-11-14
FOUN-1568: prevent race condition where 2 or more concurrent requests are fired

amiras89:


api_v3/services/UserAppRoleService.php
```diff
@@ -41,12 +41,21 @@ protected function partnerGroup($peer = null)
 	 *
 	 * @param KalturaUserAppRole $userAppRole
 	 * @return KalturaUserAppRole
-	 *
+	 */
+	public function addAction(KalturaUserAppRole $userAppRole)
+	{
+		// prevent race condition where 2 or more concurrent requests are fired
+		$lockKey = 'userAppRole_add_' . kCurrentContext::getCurrentPartnerId() . '_' . $userAppRole->appGuid . '_' . $userAppRole->userId;
```
yossipapi, line 48: From what I see the lock now runs inside all the validations for app guide and etc..., this is not really good as this means the lock time will potentially be long.
The only real thing that is important to validate while locked is the duplication of the object.

Based on this you will need to split the logic of this into 2 steps 1 that can run before the lock and the other that runs inside the lock.

amiras89: https://github.com/kaltura/server/pull/12462/files

amiras89:
Closing since we merged the 2nd version
https://github.com/kaltura/server/pull/12462

PR #12462, opened 2023-11-21
FOUN-1568_v2: add lock when doing userAppRole-->addAction

amiras89:


api_v3/services/UserAppRoleService.php
```diff
@@ -43,20 +43,41 @@ protected function partnerGroup($peer = null)
...
 	{
 		$dbUserAppRole = $userAppRole->toInsertableObject();
-		$dbUserAppRole->save();
 		
-		$userAppRole = new KalturaUserAppRole();
+		// prevent race condition where 2 or more concurrent requests are fired
+		$lockKey = 'userAppRole_add_' . kCurrentContext::getCurrentPartnerId() . '_' . $userAppRole->appGuid . '_' . $userAppRole->userId;
+		$dbUserAppRole = kLock::runLocked($lockKey, array($this, 'addUserAppRole'), array($dbUserAppRole));
+		
 		$userAppRole->fromObject($dbUserAppRole, $this->getResponseProfile());
```
yossipapi, line 58: Missing 
$userAppRole = new KalturaUserAppRole();

amiras89: Intentionally removed it
1. Same as in the example from UserEntry:
https://github.com/kaltura/server/blob/41d0d07f52e54a7450d6aea9183d3e8d9321c8ba/api_v3/services/UserEntryService.php#L28 
2. Figured we don't have to run over the object we get
3. Brought it back ;)

PR #12469, opened 2023-11-23
VCP-15400: Remove ImageMagic memory limits

david-winder-kaltura:
Without the specific limitation the default should be use as define in https://imagemagick.org/script/command-line-options.php


plugins/content/document/lib/kdl_transcoders/KDLTranscoderImageMagick.php
```diff
@@ -19,7 +19,7 @@ public function GenerateCommandLine(KDLFlavor $design, KDLFlavor $target, $extra
 			$cmdStr.= $this->getTwoDimensionsParams('density', $target->_image->_densityWidth, $target->_image->_densityHeight);
 			$cmdStr.= $this->getTwoDimensionsParams('geometry', $target->_image->_sizeWidth , $target->_image->_sizeHeight);
 			$cmdStr.= $this->getSimpleParam('depth', $target->_image->_depth);
-			$cmdStr.= '-colorspace RGB -limit memory 1000MB -limit map 200 ';
+			$cmdStr.= '-colorspace RGB ';
```
nivshlomokaltura, line 22: let's see how we take the value dynamically as % of the machine memory

PR #12470, opened 2023-11-26
PLAT-24475 support chapter naming policy for multi clip concat

salma2222:


plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -97,20 +97,43 @@ protected static function handleConcatAfterMultiClipJobFinished($dbBatchJob, $da
...
+			$entry = entryPeer::retrieveByPK($sourceEntryId);
+		}
+		switch ($chapterNamePolicy)
+		{
+			case ChapterNamePolicy::NUMERICAL:
+				$chapterIndex++;
+				return "Chapter $chapterIndex";
+
+			case ChapterNamePolicy::BY_ENTRY_ID:
+				return $entry->getId();
```
inbal-ben-david, line 129: need to validate the entry returned before accessing its property

plugins/cue_points/base/lib/kCuePointManager.php
```diff
@@ -97,20 +97,43 @@ protected static function handleConcatAfterMultiClipJobFinished($dbBatchJob, $da
...
 	}
 
+	protected static function getChapterNameByPolicy($sourceEntryId, $chapterIndex, $chapterNamePolicy)
+	{
+		$entry = null;
+		if($chapterNamePolicy != ChapterNamePolicy::NUMERICAL)
+		{
+			$entry = entryPeer::retrieveByPK($sourceEntryId);
+		}
+		switch ($chapterNamePolicy)
```
inbal-ben-david, line 122: if not set it gets as null and falls to the default case?

salma2222: yes

plugins/cue_points/base/lib/kCuePointManager.php
salma2222: added

inbal-ben-david (approved):

PR #12474, opened 2023-11-27
delete tmp files if exception is thrown

amiras89:


plugins/content_distribution/providers/cortex/lib/CortexApiDistributionEngine.php
```diff
@@ -467,6 +467,16 @@ protected function uploadVideo(KalturaCortexApiDistributionJobProviderData $apiD
 		}
 		catch(Exception $e)
 		{
+			if ($needDel)
```
TalDubovKaltura, line 470: @amiras89 You should add :
`if ($needDel && kFile::checkFileExists($videoFilePath))`
 If an exception been thrown in line 430, the unlink will be failed because there is still no file here

plugins/content_distribution/providers/cortex/lib/CortexApiDistributionEngine.php
```diff
@@ -467,6 +467,16 @@ protected function uploadVideo(KalturaCortexApiDistributionJobProviderData $apiD
...
+			{
+				kFile::unlink($videoFilePath);
+			}
+			
+			if($isRemote)
```
TalDubovKaltura, line 475: @amiras89 If an exception has been thrown before line 437, the var $isRemote still has not been set and it will print  a PHP notice, and also I would check if the file exists like that:
`if(isset($isRemote) && $isRemote && kFile::checkFileExists($tempVideoFilePath))`

TalDubovKaltura (changes requested):

plugins/content_distribution/providers/cortex/lib/CortexApiDistributionEngine.php
amiras89: IF the file is shared, and failed to unlink, it will return false.
If it's local and failed, we suppress the php notice.

If exception is thrown at 430, $needDel is false - so we won't try to unlink

Because $needDel is only set to true at 447 when we copy from s3 to the .dme file - it's sufficient IMHO
Only problem is that if $videoFilePathNew is an existing .dme - then $needDel won't be true and we won't delete
But for that we need a re-design in the flow, so need to better understand the dme as we talked over in Teams

I wanted to only check if file exists --> if yes --> delete
But then we might end up deleting the file from s3, so stayed with checking the $needDel
Better to be safe than sorry.

plugins/content_distribution/providers/cortex/lib/CortexApiDistributionEngine.php
amiras89: I suggest to simplify it to:
isset($tempVideoFilePath)

If exception is thrown before L438 - it won't be set == false.
If the file is not remote, it will be null --> which isset(null) is false

TalDubovKaltura (approved):

PR #12481, opened 2023-12-02
VCP-15458: fix setCdnHostWhiteList and admin console partner config

amiras89:


alpha/lib/model/Partner.php
```diff
@@ -2006,8 +2006,18 @@ public function getCdnHostWhiteList()
...
+		
+		// in case white space is given, store null
+		$whiteListRegEx = trim($whiteListRegEx);
+		if (!empty($whiteListRegEx))
+		{
+			$whiteListArr = explode(',', rtrim($whiteListRegEx, ','));
+			$value = serialize($whiteListArr);
+		}
+		
+		$this->putInCustomData(self::CDN_HOST_WHITE_LIST, $value);
```
amiras89, line 2020: @yossipapi - this alone is enough to fix the issue
So far, this function always stored an array
Issue is when we tried to 'clear' this value which can be:
1. Via admin console partner config (see explanation below KalturaSystemPartnerConfiguration.php)
2. Via alpha - if we do $partner->setCdnHostWhiteList(''); in attempt to clear the value, we ended up storing an empty regex '' which preg_match always translats to true - leading to us building the domain+port at:
https://github.com/kaltura/server/blob/d0cca63de262f88bf308fee6fc8c932b110ae15c/alpha/apps/kaltura/lib/myPartnerUtils.class.php#L365-L364

plugins/admin_console/system_partner/lib/api/KalturaSystemPartnerConfiguration.php
```diff
@@ -871,6 +871,11 @@ public function toObject ( $object_to_fill = null , $props_to_skip = array() )
 			$object_to_fill->setAllowedDomains('');
 		}
 		
+		if (is_null($this->cdnHostWhiteList))
+		{
+			$object_to_fill->setCdnHostWhiteList(null);
+		}
+
```
amiras89, line 878: @yossipapi this is more of a 'quality of life' thing, and I think it's required
If we omit this part, it means in order to clear the cdnHostWhiteList value using admin console partner config we need to pass ' '
Because if we just clear the checkbox "CDN Host white list regex (comma seperated)" from partner config, toObject will skip it since it's null and will not modify the value
To bypass it, we can pass ' ' - which 'setCdnHostWhiteList' will now correctly identify it as empty and set null

However, I think it is confusing and incorrect way of saying "I want to remove this for the partner"
Therefore, now if that field is_null - we will set it.

PR #12484, opened 2023-12-07
VCP-15342: add lock for groupUser->add to avoid race condition result…

amiras89:
… in duplicate objects


inbal-ben-david (commented):
code seems ok, but not convinced that locking is really needed here

PR #12512, opened 2023-12-25
SUP-40674: Fix code to support logic that was implemented in aws sdk v2

yossipapi:


infra/storage/file_transfer_managers/s3Mgr.class.php
```diff
@@ -527,7 +527,7 @@ public function registerStreamWrapper()
 		$this->s3->registerStreamWrapper();
 	}
 	
-	public function getRemoteUrl($remote_file)
+	public function getRemoteUrl($remote_file, $expiry = null)
```
inbal-ben-david, line 530: the function overrides getRemoteUrl($remote_file) on kFileTransferMgr, i think we better align there the signature as well

amiras89 (approved):

PR #12528, opened 2024-01-08
VCP-15967: add safeLog to parent class and change s3mgr to use safelog

amiras89:


amiras89:
Following https://github.com/kaltura/server/pull/12476

PR #12532, opened 2024-01-10
Tucana 20.4.0 plat 24562

shamamayair:


david-winder-kaltura:
Currently in this PR only printing to the log to gather data of the frequency of that - afterward we remove the comment from the actual hash

david-winder-kaltura (approved):

PR #12533, opened 2024-01-10
VCP-16011: Add backoff period for delete workers to not have internal operation over utilize the system

yossipapi:


batch/batches/Delete/Engines/KDeletingEngine.php
```diff
@@ -74,8 +95,22 @@ public function configure($partnerId, $jobData)
 		$this->pager = new KalturaFilterPager();
 		$this->pager->pageSize = 100;
 		
+		$this->deleteOperationBulkSize = self::DELETE_BACKOFF_BULK_SIZE;
+		if(KBatchBase::$taskConfig->params && KBatchBase::$taskConfig->params->bulkSize)
```
amiras89, line 99: We will get php notice if not exist
maybe do
$this->deleteOperationBulkSize = isset(KBatchBase::$taskConfig->params->bulkSize) ? KBatchBase::$taskConfig->params->bulkSize : self::DELETE_BACKOFF_BULK_SIZE;

batch/batches/Delete/Engines/KDeletingEngine.php
```diff
@@ -74,8 +95,22 @@ public function configure($partnerId, $jobData)
...
+		{
+			$this->deleteOperationBulkSize = KBatchBase::$taskConfig->params->bulkSize;
+		}
+		
+		$this->deleteOperationBackOffInterval = self::DELETE_BACKOFF_INTERVAL;
+		if(KBatchBase::$taskConfig->params && KBatchBase::$taskConfig->params->backOffInterval)
```
amiras89, line 105: same as above

batch/batches/Delete/Engines/KDeletingEngine.php
```diff
@@ -88,6 +123,12 @@ public function run(KalturaFilter $filter)
 	{
 		KBatchBase::impersonate($this->partnerId);
 		$ret = $this->delete($filter);
+		self::$deletedRecordCount += $ret;
+		if((self::$deletedRecordCount % $this->deleteOperationBulkSize) == 0)
```
amiras89, line 127: Looking at the filters, we don't set pageSize = 500 explicitly
Also, I think if someone change a filter or if not all 500 return due to index, we will not sleep

Maybe we should cycle every time we passed 500
like:
if(((self::$deletedRecordCount / $this->deleteOperationBulkSize) % $this->deleteOperationBulkSize) >= 1)
{
	self::$deletedRecordCount = 0;
	sleep($this->deleteOperationBackOffInterval);
}

That way, no matter the pageSize or if not all objects returned we sleep every 500 objects

batch/batches/Delete/Engines/KDeletingEngine.php
yossipapi: This remains aligned with how other flag checks in batch work so this is why I wanted to keep the code style the same.

batch/batches/Delete/Engines/KDeletingEngine.php
yossipapi: I am not sure I follow this comment.
No one should set page size which is big it wont make sense as it might die on memory.
I dont really want to be aligned to page size to do the sleep necessarily, and if I do than I will be able to control it using the worker config.

batch/batches/Delete/Engines/KDeletingEngine.php
yossipapi: same as above

amiras89 (approved):

PR #12544, opened 2024-01-15
LIV-1153-Signaling-to-separate-pre-and-post-from-content-isset-validation

zoharsKaltura:


david-winder-kaltura (approved):

david-winder-kaltura:
In the vod packager - if we get empty string as id we use the path (mappedUri) as the id

salma2222 (approved):

PR #12569, opened 2024-02-08
SUP-40534: get admins count without excluded role users when adding l…

coralburg12:
…ogin data


alpha/lib/model/UserLoginDataPeer.php
```diff
@@ -805,7 +805,24 @@ protected static function isExcludedAdminRole($partner, $userRoleNames)
...
+		{
+			return 0;
+		}
+		$excludedUserRole = UserRolePeer::getByNameAndPartnerId($excludedRoleName, $partner->getId());
+		if (!$excludedUserRole)
+		{
+			return 0;
+		}
+		$c = new Criteria();
+		$c->add(KuserToUserRolePeer::USER_ROLE_ID, $excludedUserRole->getId());
```
salma2222, line 822: try to check 
$excludedUserRole->countKuserToUserRoles();
instead, it seems to do what you need

drorsou (approved):

PR #12574, opened 2024-02-12
SUP-40795 adding live stream entry type to taggedObjectType interface

Tova-a:


inbal-ben-david (approved):
rebase to tucana 20.10.0

PR #12595, opened 2024-03-03
FOUN-1370: add kafka event notification for roomEntry changes

Amirhalfon1:
add kafka event notification template for roomEntry changes
raise event also if it doesn't require any permission


deployment/updates/scripts/2024_03_03_deploy_kafka_room_events.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * @package deployment
+ */
+require_once (__DIR__ . '/../../bootstrap.php');
+
+$script = realpath(dirname(__FILE__) . "/../../../tests/standAloneClient/exec.php");
+$newTemplateUpdate = realpath(dirname(__FILE__) . "/../../updates/scripts/xml/notifications/2024_03_03_add_kafka_room_entry_notifications.template.xml");
+
+print_r($newTemplateUpdate);
```
yossipapi, line 10: Why the print_r here?

plugins/event_notification/providers/kafka/lib/model/KafkaNotificationTemplate.php
```diff
@@ -114,7 +114,7 @@ public function dispatch(kScope $scope)
 		}
 
 		$requiredPermissions = explode(",", $this->getRequiresPermissions());
-		if(count($requiredPermissions))
+		if(count(array_filter($requiredPermissions)))
```
yossipapi, line 117: If the purpose is to remove all empty values that return true on the empty() function please state the function when calling array_filter. To avoid case where this logic will change in the future.

From PHP doc:
callback
The callback function to use

If no callback is supplied, all empty entries of array will be removed. See [empty()](https://www.php.net/manual/en/function.empty.php) for how PHP defines empty in this case.

deployment/updates/scripts/xml/notifications/2024_03_03_add_kafka_room_entry_notifications.template.xml
```diff
@@ -0,0 +1,51 @@
...
+            <template objectType="KalturaKafkaNotificationTemplate">
+                <name>Room Entry Updated</name>
+                <systemName>KAFKA_ROOM_ENTRY_UPDATED</systemName>
+                <description>Kafka notification template to be sent when room entry updated</description>
+                <automaticDispatchEnabled>1</automaticDispatchEnabled>
+                <eventType>3</eventType>
+                <!-- EventNotificationEventType::OBJECT_CHANGED -->
+                <eventObjectType>1</eventObjectType>
+                <!-- EventNotificationEventObjectType::ENTRY -->
+                <topicName>@TOPIC_NAME@</topicName>
```
yossipapi, line 27: I would pur @TOPIC_NAME@ statically to be entry.
I know that previously we did this templating but I am not sure this is the right way.

yossipapi (approved):

PR #12598, opened 2024-03-05
Tucana 20.7.0 plat 24535

danbar0kal:


api_v3/services/LiveStreamService.php
```diff
@@ -11,6 +11,9 @@ class LiveStreamService extends KalturaLiveEntryService
 {
 	const ISLIVE_ACTION_CACHE_EXPIRY_WHEN_NOT_LIVE = 5;
 	const ISLIVE_ACTION_CACHE_EXPIRY_WHEN_LIVE = 30;
+
+	const GET_LIVE_STATS = 5;
```
inbal-ben-david, line 15: let's take this value from the live configuration map

api_v3/services/LiveStreamService.php
```diff
@@ -617,6 +620,38 @@ public function archiveAction($liveEntryId, $vodEntryId)
         return true;
     }
 
+	private function responseHandlingGetStats($isLive)
```
inbal-ben-david, line 623: protected and not private

api_v3/services/LiveStreamService.php
```diff
@@ -617,6 +620,38 @@ public function archiveAction($liveEntryId, $vodEntryId)
...
+	{
+		if ($isLive)
```
inbal-ben-david, line 625: i think that even if we are not live we need to have cache expiry

api_v3/services/LiveStreamService.php
```diff
@@ -649,6 +684,18 @@ public function getDetailsAction($id)
 		return $liveStreamDetails;
 	}
 
+	protected function getNumberOfViewers($entryId)
+	{
+		$cache = kCacheManager::getSingleLayerCache(kCacheManager::CACHE_TYPE_PLAYS_VIEWS);
+		if (!$cache)
+		{
+			throw new kUploadTokenException("Cache instance required for Live Viewers Caching Could not initiated", kCoreException::PLAYS_VIEWS_CACHE_NOT_INITIALIZED);
```
inbal-ben-david, line 692: we usually dont throw such type of errors

api_v3/services/LiveStreamService.php
```diff
@@ -649,6 +684,18 @@ public function getDetailsAction($id)
...
+		}
+		$numberOfViewers = $cache->get("plays_views_" . $entryId);
```
inbal-ben-david, line 694: why you are using plays_views key? you should have you own format of a key for the live viewers like "current_viewers_{entry_id}

api_v3/services/LiveStreamService.php
```diff
@@ -649,6 +684,18 @@ public function getDetailsAction($id)
...
+
+		return $numberOfViewers ?: 0;
```
inbal-ben-david, line 696: not clear

api_v3/services/LiveStreamService.php
```diff
@@ -617,6 +618,46 @@ public function archiveAction($liveEntryId, $vodEntryId)
         return true;
     }
 
+	protected function responseHandlingGetStats($isLive)
+	{
+		$liveStatsInterval = kConf::get('liveStatsInterval','live');
```
inbal-ben-david, line 623: take live mape name from kConfMapNames class
also add default value of 5 for the interval

api_v3/services/LiveStreamService.php
```diff
@@ -617,6 +618,46 @@ public function archiveAction($liveEntryId, $vodEntryId)
...
+	 * @ksIgnored
+	 *
+	 * @throws KalturaErrors::INVALID_ENTRY_ID
+	 */
+	public function getLiveStreamStatsAction($entryId)
+	{
+		$liveStreamEntry = $this->fetchLiveEntry($entryId);
+		$isLive = $liveStreamEntry->isCurrentlyLive();
+		$liveStreamInfo = new KalturaLiveStreamStats();
+		$liveStreamInfo->liveViewers = $isLive ? $this->getNumberOfViewers($entryId) : 0;
```
inbal-ben-david, line 643: lets avoid extracting the entry and just return what we have in cache, if its not live the value will be null

api_v3/services/LiveStreamService.php
danbar0kal: didnt nadav asked for 20?

inbal-ben-david: yes it was before, put 20

inbal-ben-david (approved):

PR #12599, opened 2024-03-06
PSVAMB-48998

hilak:
add option to search and order users by fullname using elastic


plugins/search/providers/elastic_search/lib/api/enum/KalturaESearchBaseUserFieldName.php
```diff
@@ -6,6 +6,7 @@
 class KalturaESearchBaseUserFieldName extends KalturaStringEnum
 {
 	const SCREEN_NAME = 'screen_name';
+	const FULL_NAME = 'full_name';
```
inbal-ben-david, line 9: searching of first and last name which already exist cant replace the need for full name on that case?

hilak: @inbal-ben-david not for the user-case in question. Additionally, we have the same ability in the sphinx and should strive for parity.

amiras89 (approved):

PR #12616, opened 2024-03-18
LIV-1153: change clipId structure

zoharsKaltura:


david-winder-kaltura (approved):

alpha/apps/kaltura/lib/kSimuliveUtils.php
```diff
@@ -396,10 +395,18 @@ public static function isOffsetPlaybackAllowed($ksString)
 		}
 	}
 
-	public static function addParamToId($key, $value, $existId=null)
+	private static function generateSourceLabel($type, $srcId)
```
inbal-ben-david, line 398: code convention is to use protected and not private

inbal-ben-david (approved):

PR #12617, opened 2024-03-18
Plat 24606: Multi entry packager clipping API

zoharsKaltura:


alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
```diff
@@ -164,7 +164,28 @@ public static function serveLiveMediaSet($durations, $sequences, $playlistStartT
...
+		if ($entry->getMediaType() == PlaylistType::STATIC_LIST && $entry->getOperationAttributes())
+		{
+			$clipAttributes = array_filter($entry->getOperationAttributes(), function($obj) {
+				return $obj instanceof kClipAttributes;
+			});
+			for ($i = 0; $i < min(count($durations), count($clipAttributes)); $i++)
+			{
+				$offset = $clipAttributes[$i]->getOffset();
+				$duration = $clipAttributes[$i]->getDuration();
+				if ($offset >= 0 && $duration > 0 && $durations[$i] - $offset > $duration)
```
inbal-ben-david, line 180: offset =0, duration = 30, $durations[$i] = 30 is not a valid case?

alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
```diff
@@ -186,14 +207,16 @@ protected function servePlaylist($entry, $captionLanguages)
 		}
 		
 		list($entryIds, $durations, $referenceEntry, $captionFiles) = myPlaylistUtils::executeStitchedPlaylist($entry, $captionLanguages);
+		list($offsets, $durations) = $this->retrievePlaylistAttributesData($entry, $durations);
```
inbal-ben-david, line 210: is it intended to support also the existing stitched playlist? else i would have preffer not to have th possibility to override duration for existing use case with the new function

alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
```diff
@@ -186,14 +207,16 @@ protected function servePlaylist($entry, $captionLanguages)
...
-			$captionFiles, $captionLanguages, $isLive, 0, 0, 0, 0);
+			$captionFiles, $captionLanguages, $isLive, 0, 0, 0, 0, $offsets);
 	}
 
 	protected function serveEntriesAsPlaylist($entryIds, $durations, $referenceEntry, $origEntry, $flavorParamIds,
 	                                          $captionFiles, $captionLanguages, $isLive,
-	                                          $playlistStartTime, $firstClipStartTime, $initialClipIndex, $initialSegmentIndex)
+	                                          $playlistStartTime, $firstClipStartTime, $initialClipIndex, $initialSegmentIndex, $offsets=null)
 	{
+		$offsets = $offsets ?? array_fill(0, count($entryIds), 0);
```
inbal-ben-david, line 219: ?? works only from php 7, there are on prems will older php version so please write the full condition

alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
zoharsKaltura: yes, it's intended to support stitched playlist

alpha/apps/kaltura/modules/extwidget/actions/serveFlavorAction.class.php
```diff
@@ -164,7 +164,28 @@ public static function serveLiveMediaSet($durations, $sequences, $playlistStartT
...
 		return $mediaSet;
 	}
-	
+	// if the entry has clip attributes returns the offsets and updated durations, otherwise returns zeros and original durations
+	protected function retrievePlaylistAttributesData($entry, $durations)
+	{
+		$offsets = array_fill(0, count($durations), 0);
+		if ($entry->getMediaType() == PlaylistType::STATIC_LIST && $entry->getOperationAttributes())
+		{
+			$clipAttributes = array_filter($entry->getOperationAttributes(), function($obj) {
```
yossipapi, line 173: Dynamic function declaration is not supported in php8 so please fix it already as we are moving to php8 shortly.
So to fix it you need declare the function and pass it to the calls.
So something like this:

$func = function($obj) { return $obj instanceof kClipAttributes; };
$clipAttributes = array_filter($entry->getOperationAttributes(), $func);

zoharsKaltura: fixed

inbal-ben-david (approved):

PR #12622, opened 2024-03-24
LIV-937: Add HD/1080 flavor param.

shamamayair:


deployment/updates/scripts/ini_files/2024_03_24_live_1080.liveParams.ini
```diff
@@ -0,0 +1,63 @@
+[HD-1080-WEB-MBL-H264-4000]
+id=44
```
yossipapi, line 2: You can't static ID's in the update file only in the new env creation file.
In some ENV's these IDs are probably already used.

Please remove them so they will be inserted with random IDs.
If you want certain ENV"S to have specific IDs you can always edit the file before running.

deployment/updates/scripts/2024_03_24_add_1080_live_flavour_param.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * @package deployment
+ *
+ * Deploy new Azerbaijani and Urdu language flavors for live
```
inbal-ben-david, line 5: not related

inbal-ben-david (approved):

PR #12631, opened 2024-03-31
Adding missing microservice_url

rgvsilva:


amiras89 (approved):

configurations/local.template.ini
```diff
@@ -201,6 +201,9 @@ analytics_internal_API_url = @SERVICE_URL@:9000/query
 druid_url = @DRUID_BROKER_URL@/druid/v2/
 analytics_sync_secret = @ANALYTICS_SYNC_SECRET@
 
+;microservice_url
+microservice_url = @MICROSERVICE_URL@/api/v1/
```
yossipapi, line 205: This should be different
microservice_url = https://[micro-url-prefix].@MICROSERVICE_URL@/api/v1/

After replacement it should be something like:
https://[micro-url-prefix].nvp1.ovp.kltura.com/api/v1/

yossipapi (changes requested):

inbal-ben-david (approved):

PR #12657, opened 2024-04-17
PLAT-24811: enable set segment duration for specific flavor asset

zoharsKaltura:


david-winder-kaltura (approved):

alpha/lib/model/flavorAsset.php
```diff
@@ -405,4 +405,21 @@ public function getTypeFolderName()
...
+	 * value is in milliseconds
+	 */
+	public function getSegmentDuration()
+	{
+		$tags = explode(',', $this->getTags());
+		foreach ($tags as $tag)
+		{
+			if (strpos($tag, 'segment_duration:') !== false)
+			{
+				return (int)explode(':', $tag)[1];
```
drorsou, line 419: Better be safe and verify there is indeed a value after the colon ':'
Else it would throw a warning of array key missing

drorsou (approved):

PR #12658, opened 2024-04-21
LIV-1253: Force fmp4 in mix container

shamamayair:
If the is a flavor with H265 put add to the URL container with fmp4 value that will tell the live to use fmp4 and not mpegts


david-winder-kaltura:
Basically, getAdditionalUrlParam get get only the entryServreNode param and extract the entry from it. Currently we sending both as we want to do minimum code changes

david-winder-kaltura (approved):

drorsou (approved):

PR #12687, opened 2024-05-09
VCP-17469: php8 TypeError round #1

amiras89:


api_v3/lib/KalturaEntryService.php
```diff
@@ -1397,7 +1397,7 @@ protected function checkAndSetValidUserInsert(KalturaBaseEntry $entry, entry $db
 		}
 		else
 		{
-			$creatorId = trim($entry->creatorId);
+			$creatorId = $entry->creatorId instanceof KalturaNullField ? null : trim($entry->creatorId);
```
amiras89, line 1400: @yossipapi this would be the way to keep behavior

if we want to 'correct' the behavior, we can place the condition at line 1391 like:
if (is_null($entry->creatorId) || $entry->creatorId instanceof KalturaNullField)

yossipapi: I do agree its a better solution but this will change the existing behavior.
Please check the results of the API in both cases, when the check is where you put it and when the check is in the if on line 1391.

Once you have the API call results lets compare and see the diff and decide based on that.

PR #12694, opened 2024-05-11
No ticket - change the date update script to work with standard CSV

hilak:


yossipapi (approved):

yossipapi:
@hilak why not combine these 2 scripts into 1 to support updating both createdAt and updatedAt in a single run?
I assume you do have use cases like this no?

PR #12709, opened 2024-05-19
PLAT-24755

danbar0kal:


api_v3/services/SessionService.php
```diff
@@ -128,7 +128,11 @@ function impersonateAction($secret, $impersonatedPartnerId, $userId = "", $type
 		{
 			$impersonatedSecret = $impersonatedPartner->getSecret();
 		}
-		
+
+		if ($partnerId < 0)
```
yossipapi, line 132: Why is this added here?
There is already a place in the code where we manipulate the KS max expiry levels and it's here:
https://github.com/kaltura/server/blob/38a1a0ec7270ad98048eb29ca1c7927bae2a96f4/alpha/lib/model/Partner.php#L135

danbar0kal: from what i saw, that place only validates the secret and expiry, but doesn't use it actually when starting the session.

yossipapi: It depends on what you pass when calling startKSession if you don't pass $enforcePartnerKsMaxExpiry it wont actually do anything 
Either way if your touching the ks expiry the more correct place to do it is where all this logic is and not in the service call itself

danbar0kal: i see, ok. i'll look at it

alpha/apps/kaltura/lib/webservices/kSessionUtils.class.php
```diff
@@ -94,6 +93,22 @@ public static function startKSession ( $partner_id , $partner_secret , $puser_id
...
 	}
 
+	protected static function calculateExpiry($ks_max_expiry_in_seconds, &$desired_expiry_in_seconds, $enforcePartnerKsMaxExpiry, $master_partner_id)
+	{
+		if ($master_partner_id && $master_partner_id < 0)
+		{
+			$masterPartner = PartnerPeer::retrieveByPK($master_partner_id);
+			if ($masterPartner && $desired_expiry_in_seconds)
+			{
+				$ks_max_expiry_in_seconds = $desired_expiry_in_seconds;
```
inbal-ben-david, line 103: just return instead of updating ks_max_expiry_in_seconds

inbal-ben-david (approved):

PR #12710, opened 2024-05-19
PLAT-24302: Add first and last partner range

drorsou:


alpha/scripts/utils/enableVirtualEventPermissionToAllPartners.php
```diff
@@ -4,16 +4,20 @@
...
+ * php enableVirtualEventPermissionToAllPartners.php 99 1000 realrun
  *
  * @package Deployment
  * @subpackage updates
  */
 
-$dryRun = true;
-if (in_array('realrun', $argv))
-	$dryRun = false;
+$dryRun = in_array('realrun', $argv) ? false : true;
```
inbal-ben-david, line 14: better move this line after the  ($argc < 3) check and check the exact argv value and not that one of the args is realrun

alpha/scripts/utils/enableVirtualEventPermissionToAllPartners.php
```diff
@@ -30,17 +34,23 @@
...
-
 $partners = PartnerPeer::doSelect($c, $con);
 
+$isLastPartner = false;
 while (count($partners))
 {
 	foreach ($partners as $partner)
 	{
 		/* @var $partner Partner */
+		if ($partner->getId() > $lastPartner)
```
inbal-ben-david, line 48: instead of adding all the conditions with the breaks you can add to the db query id lessThenEqual

alpha/scripts/utils/enableVirtualEventPermissionToAllPartners.php
```diff
@@ -53,10 +63,15 @@
...
-		print("Set permission [" . VIRTUALEVENT_PLUGIN_PERMISSION . "] for partner id [". $partner->getId() ."]");
+		print("Set permission [" . VIRTUALEVENT_PLUGIN_PERMISSION . "] for partner id [". $partner->getId() ."]\n");
 		$virtualEventPermission->setStatus(PermissionStatus::ACTIVE);
 		$virtualEventPermission->save();
 	}
+
+	if ($isLastPartner)
+	{
+		break;
+	}
```
inbal-ben-david, line 74: on line 79 you explicitly use 99 and not $firstPartner

PR #12715, opened 2024-05-20
PLAT-24816: Increase groupUser limit to 2048

drorsou:


api_v3/services/GroupUserService.php
```diff
@@ -434,4 +428,24 @@ protected function validateKuserkGroupCoExistence(kuser $kGroup, $kuserId)
...
 	}
+
+	protected function validateMaxGroupsPerUser($partnerId, $kuserId)
+	{
+		$criteria = new Criteria();
+		$criteria->add(KuserKgroupPeer::KUSER_ID, $kuserId);
+		$criteria->add(KuserKgroupPeer::STATUS, KuserKgroupStatus::ACTIVE);
+
+		$maxGroupsPerUser = GroupUserService::DEFAULT_MAX_NUMBER_OF_GROUPS_PER_USER;
+		$groupUserLimitMap = kConf::get('groupUserLimit');
```
inbal-ben-david, line 439: if there will be no such param in one of the envs you will get an execption here, make sure you pass local,null in the function params to avoid it

release-notes.md
```diff
@@ -1,12 +1,25 @@
+# Tucana-20.12.0
+## Configure partner limit on max groups per user ##
+- Issue Type: Task
+- Issue ID: PLAT-24816
+
+### Configuration ###
+Add the following to local.ini:
+```
+[groupUserLimit]
```
inbal-ben-david, line 9: change this to groupUserCountLimit (all on all places that reffer to it)

inbal-ben-david (approved):

api_v3/services/GroupUserService.php
yossipapi: I would probably set default value not to null but to an array to have both default and actual value types to be from the same type.

drorsou: The merged version is indeed with a default empty array. 😃

PR #12718, opened 2024-05-21
Support cross kaltura target url manipulation based on worker config

yossipapi:


plugins/content_distribution/providers/cross_kaltura/lib/batch/CrossKalturaDistributionEngine.php
```diff
@@ -142,12 +166,28 @@ protected function initClients(KalturaCrossKalturaDistributionProfile $distribut
...
 	}
+	
+	private function getTargetServiceUrl($serviceUrl)
+	{
+		if(!$this->targetServiceUrlRegex)
+		{
+			return $serviceUrl;
+		}
+		
+		$matches = array();
```
amiras89, line 183: Do we need it? it is not being used as far as I can see.

amiras89 (approved):

plugins/content_distribution/providers/cross_kaltura/lib/batch/CrossKalturaDistributionEngine.php
yossipapi: I was using it before and forgot to remove it.
Removed now.

plugins/content_distribution/providers/cross_kaltura/lib/batch/CrossKalturaDistributionEngine.php
```diff
@@ -142,12 +166,27 @@ protected function initClients(KalturaCrossKalturaDistributionProfile $distribut
...
 	}
+	
+	private function getTargetServiceUrl($serviceUrl)
+	{
+		if(!$this->targetServiceUrlRegex)
+		{
+			return $serviceUrl;
+		}
+		
+		if (!preg_match($this->targetServiceUrlRegex, $serviceUrl, $matches))
```
amiras89, line 183: need to remove the $matches from here too :)

yossipapi: Done.

PR #12733, opened 2024-05-28
PSVAMB-48298 -> Change entry properties using input of CSV

TalDubovKaltura:
@hilak I'm combining the createdAt/updatedAt updating into single script file, supporting all other entry properties.

### DIFF
--- a/alpha/scripts/utils/setEntryProperties.php
+++ b/alpha/scripts/utils/setEntryProperties.php
@@ -1,0 +1,73 @@
+<?php
+
+if($argc < 3)
+{
+    echo "Arguments missing.\n\n";
+    echo "Usage: php " . __FILE__ . " {mapping} {isDateString} <realrun / dryrun> \n";
+    exit;
+}
+$mapping = $argv[1];
+$isDateString = ($argv[2] === "true");
+$dryRun = ($argv[3] === "dryrun");
+
+require_once(__DIR__ . '/../bootstrap.php');
+
+
+KalturaStatement::setDryRun($dryRun);
+
+$entryMappings = file($mapping, FILE_IGNORE_NEW_LINES);
+
+$properties = [""];
+$counter = 0;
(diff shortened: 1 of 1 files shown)


alpha/scripts/utils/setEntryProperties.php
```diff
@@ -0,0 +1,75 @@
...
+                continue;
+            }
+            $properties[] = $property;
+        }
+        continue;
+    }
+    $entryValues= explode(",", $entryMapping);
+
+    $z = 0;
+    foreach($entryValues as $entryValue){
```
hilak, line 42: @TalDubovKaltura  basically you're calling entry::save() after each property set instead of just once per entry ID. Please re-write this code block

alpha/scripts/utils/setEntryProperties.php
```diff
@@ -0,0 +1,75 @@
...
+        if($z === 0){
```
hilak, line 43: @TalDubovKaltura  This approach is very hard to follow. Instead I suggest that doing
`entryValues = array_combine($properties,$entryValues)` 
and then you can just call 
`$entry = entryPeer::retrieveByPK($entryValues['id']);`

alpha/scripts/utils/setEntryProperties.php
```diff
@@ -0,0 +1,75 @@
...
+KalturaStatement::setDryRun($dryRun);
+
+$entryMappings = file($mapping, FILE_IGNORE_NEW_LINES);
+
+$properties = [""];
+$counter = 0;
+entry::setAllowOverrideReadOnlyFields(true);
+foreach ($entryMappings as $entryMapping)
+{
+    if($counter++ == 0){
```
hilak, line 25: @TalDubovKaltura  I'd suggest just taking this if statement out of the loop altogether

hilak (changes requested):

PR #12760, opened 2024-06-18
SUP-39912 Allow specific partners to use API in blocked countries

bw-kaltura:


alpha/lib/model/Partner.php
```diff
@@ -1890,7 +1890,8 @@ protected function validateGlobalApiAccessLimitations()
 		if ($globalAccessLimitationsConfiguration)
 		{
 			$blockedCountriesList = $globalAccessLimitationsConfiguration['blockedCountries'];
-			if ($blockedCountriesList)
+            		$allowedPartnersInBlockedCountries = $globalAccessLimitationsConfiguration['allowedPartnersInBlockedCountries'];
+			if ($blockedCountriesList && (!$allowedPartnersInBlockedCountries || !in_array($this->id, explode(",", $allowedPartnersInBlockedCountries))))
```
inbal-ben-david, line 1894: i would preffer the condition to be:
$allowedPartnersInBlockedCountries = $globalAccessLimitationsConfiguration['allowedPartnersInBlockedCountries'];
if(allowedPartnersInBlockedCountries && in_array($this->id, explode(",", $allowedPartnersInBlockedCountries)))
return true
$blockedCountriesList = $globalAccessLimitationsConfiguration['blockedCountries'];
if ($blockedCountriesList)
... as it was

currently on your code you will do expload and in_array potentially on null

inbal-ben-david (approved):

PR #12766, opened 2024-06-20
Tucana 20.14.0 vcp 17453 propel

amiras89:


plugins/search/providers/sphinx_search/scripts/setSphinxLogServerPopulateActive.php
```diff
@@ -0,0 +1,150 @@
...
+KAutoloader::addClassPath(KAutoloader::buildPath(KALTURA_ROOT_PATH, "vendor", "propel", "*"));
+KAutoloader::addClassPath(KAutoloader::buildPath(KALTURA_ROOT_PATH, "plugins", "*"));
+KAutoloader::setClassMapFilePath(kConf::get("cache_root_path") . '/sphinx/' . basename(__FILE__) . '.cache');
+KAutoloader::register();
+
+error_reporting(E_ALL);
+KalturaLog::setLogger(new KalturaStdoutLogger());
+
+$populateActiveValue = intval($argv[1]);
+$hostname = $argv[2] ?? ($_SERVER["HOSTNAME"] ?? gethostname());
```
inbal-ben-david, line 38: why if  $argv[2] is not set you take the running server?
also this condition like that is less readable in my opinion

plugins/search/providers/sphinx_search/scripts/setSphinxLogServerPopulateActive.php
```diff
@@ -0,0 +1,150 @@
...
+	}
+}
+
+KalturaLog::log("Successfully updated all sphinx_log_server.server [$sphinxServer] sphinx_log_server.populate_active to [$populateActiveValue]");
+KalturaLog::log("Sleeping for 5 seconds to wait for 'populateFromLog.php' to identify the change to sphinx_log_server for [$sphinxServer]");
+sleep(5);
+
+// verify last_log_id progress or not progress
+SphinxLogServerPeer::clearInstancePool();
+$sphinxLogServers = SphinxLogServerPeer::retrieveByServer($sphinxServer, $sphinxReadConn);
```
inbal-ben-david, line 102: why you need to call it again if you already have it from line 72?

plugins/search/lib/model/om/BaseSphinxLogServer.php
```diff
@@ -429,6 +490,10 @@ public function setUpdatedAt($v)
 	 */
 	public function hasOnlyDefaultValues()
 	{
+			if ($this->populate_active !== 1) {
```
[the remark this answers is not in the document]
amiras89: fixed
BTW - this was generated by propel, didn't touch it
But yeah, seems indentation was off. thanks.

plugins/search/providers/sphinx_search/scripts/setSphinxLogServerPopulateActive.php
amiras89: In populateFromLog we take `$_SERVER["HOSTNAME"] if set or gethostname();`
https://github.com/kaltura/server/blob/b69eb6bd3f9f570058378d62ed415aff85017e34/plugins/search/providers/sphinx_search/scripts/populateFromLog.php#L32-L31 

This works when both sphinx and sphinxPopulate are running on the same hostname
In world of k8s - it is likely that the name of the sphinx and the populate will be different
like "sphinx-1" and "sphinx-populate-1"

at the end of the day, this whole 'hostname' game is to find the record of "kaltura_sphinx_log" database for "sphinx_log_server.server"

**truthfully, made it as a 'preparation' - but it is likely we will not use it since we will adjust config to match relevant 'server' name (like in nvq2, irp2 and frp2 index-populate pods today)**

------------------------------------------------------------------
Hmmm . . is more readable? :)
`$hostname = isset($argv[2]) ? $argv[2] : (isset($_SERVER["HOSTNAME"]) ? $_SERVER["HOSTNAME"] : gethostname());`


I think once we get used to it, it's nice (like in JS)
"if is set take it" ?? "otherwise, take this value"

amiras89: @inbal-ben-david correcting myself (seems I forgot)
Once we detach sphinxPopulate from sphinx - we don't need /opt/kaltura/app/configurations on these machines
I think this is the scenario I planned for when support passing hostname manually

plugins/search/providers/sphinx_search/scripts/setSphinxLogServerPopulateActive.php
amiras89: I want to get the 'last_log_id' value after the 'populate_active' set to 0 and the 5 sec sleep at line 98
Which should be static now and not progress.

amiras89: @inbal-ben-david I moved the sleep inside the "if ($populateActiveValue === 0)"
since if we set it to 1 there is no need to sleep and wait.

PR #12781, opened 2024-07-02
PLAT-24888: Player bundle caching headers should depend on ACL

MosheMaorKaltura:


alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -444,6 +447,34 @@ private function sendHeaders($content, $lastModified)
 		infraRequestUtils::sendCachingHeaders($max_age, false, $lastModified);
 	}
 
+	private function shouldDisableCache() : bool
+	{
+		$entry_id = $this->getRequestParameter(self::ENTRY_ID_PARAM_NAME);
```
yossipapi, line 452: What if its a playlist?

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -544,7 +575,7 @@ private function getIfarmEmbedCode($bundleContent)
                         <body >
                         	<div id="player_container"></div>
                             	<script type = "text/javascript" > ' . $bundleContent . '</script >
-				<script type = "text/javascript" > window.originalRequestReferrer = "' . $_SERVER['HTTP_REFERER'] . '"</script >
+				<script type = "text/javascript" > window.originalRequestReferrer = "' . @$_SERVER['HTTP_REFERER'] . '"</script >
```
yossipapi, line 578: instead of suppressing this shouldn't this added only if $_SERVER['HTTP_REFERER'] is set?

MosheMaorKaltura: this is the common approach used over the code

yossipapi: That now why i am asking 

What will happen if the player gets something like this:
<script type = "text/javascript" > window.originalRequestReferrer = ""</script >

MosheMaorKaltura: The player handles it, if the response is empty "", it goes to fallback

PR #12793, opened 2024-07-08
VCP-17585: Remove PartnerId which is not defined to avoid failure in …

yossipapi:
…removing vendor catalog items


amiras89 (approved):
looks good to me

PR #12805, opened 2024-07-16
Tucana 20.15.0 plat 24859 headers

danbar0kal:


alpha/apps/kaltura/modules/kmcng/actions/kmcngAction.class.php
```diff
@@ -9,6 +9,19 @@ class kmcngAction extends kalturaAction
 	const PLAYER_V3_VERSIONS_TAG = 'playerV3Versions';
 	const PLAYER_V3_OVP_VERSIONS_TAG = 'playerV3OvpVersions';
 
+	public static function getDirectivesAndAllowlists()
+	{
+		$directivesToDisable = kConf::get('kmcng_permissions_policy_directives');
+		$permissionPolicyHeader = "Permissions-Policy: ";
+		$directives = [];
+		foreach ($directivesToDisable as $directive => $allowList)
```
inbal-ben-david, line 17: why you cakk the var name to disable then pass allow list?

danbar0kal: allowList is the technical name of the values to set to the feature/directive. allow list can contain () which means 'dont allow', and this value can be edited in local.ini

inbal-ben-david (approved):

PR #12818, opened 2024-07-23
REACH2-1245: Add additional fields to catalog items

drorsou:


plugins/reach/lib/model/VendorCatalogItem.php
```diff
@@ -229,5 +244,43 @@ public function requiresEntryReady()
...
+	}
+
+	public function setContract($v)
+	{
+		$this->putInCustomData(self::CUSTOM_DATA_CONTRACT, $v);
+	}
+
+	public function getContract()
+	{
+		return $this->getFromCustomData(self::CUSTOM_DATA_CONTRACT, null, VendorCatalogItemStage::PRODUCTION);
```
inbal-ben-david, line 264: why contract, createdBy and notes going by deafult to production?
looks like copy past mistake from stage

plugins/reach/lib/api/data/vendorCatalogItem/KalturaVendorCatalogItem.php
```diff
@@ -97,6 +97,31 @@ abstract class KalturaVendorCatalogItem extends KalturaObject implements IRelate
...
 
+	/**
+	 * @var KalturaVendorCatalogItemStage
+	 */
+	public $stage;
+
+	/**
+	 * @var int
+	 */
+	public $bulkUpdateId;
```
inbal-ben-david, line 108: this field is not fully clear to me, its the one that originally updated it? the last one?
why its interesting to expose it if there can be multiple updates?

drorsou:
Merged with this PR:
https://github.com/kaltura/server/pull/12824

PR #12824, opened 2024-07-29
REACH2-1265: Catalog item management using bulk update

drorsou:


plugins/bulk_upload/csv/batch/BulkUploadVendorCatalogItemEngineCsv.php
```diff
@@ -108,27 +118,7 @@ protected static function handlePriceColumns(&$pricing, $bulkUploadResult, $colu
...
-				break;
-
-			case 'outputFormat':
-				$enumValue = self::getEnumValue('KalturaVendorCatalogItemOutputFormat', $value);
-				break;
-
-			default:
-				$enumValue = null;
-		}
+		$enumValue = self::getEnumValue($this->columnToEnum[$column] ,$value);
```
inbal-ben-david, line 121: you have here an assomption that we will get a valid column name that will have enum value

plugins/bulk_upload/csv/batch/BulkUploadVendorCatalogItemEngineCsv.php
```diff
@@ -257,6 +247,7 @@ protected function createObjects()
 				case KalturaBulkUploadAction::UPDATE:
 					$bulkUploadResultChunk[] = $bulkUploadResult;
 					$vendorCatalogItem = $this->createVendorCatalogItemFromResult($bulkUploadResult);
+					$vendorCatalogItem->bulkUpdateId = $this->job->id;
```
inbal-ben-david, line 250: i think we should change it to lastBulkUpdateId on the second pr to make it more clear whats this field reflects

inbal-ben-david (approved):

PR #12835, opened 2024-08-08
ONPREM-911 block running script in parallel

GElkayam:


amiras89:
lgtm
@inbal-ben-david @NadavHarnik never seen this script and don't know who is using it
If it is used by everyone - I told Guy I think we should pass a param like "lock" when calling it and only then we implement the locking feature

But if it is only used by On-Prem, I do not mind.

inbal-ben-david (approved):

PR #12845, opened 2024-08-19
REACH2-1236: Require source language captions via API

drorsou:


plugins/reach/services/EntryVendorTaskService.php
```diff
@@ -63,11 +63,19 @@ public function addAction(KalturaEntryVendorTask $entryVendorTask)
 		$taskVersion = $dbVendorCatalogItem->getTaskVersion($dbEntry->getId(), $dbTaskData);
 		$taskDuration = $dbTaskData ? $dbTaskData->getEntryDuration() : null;
 		$featureType = $dbVendorCatalogItem->getServiceFeature();
-		if(!kReachUtils::isFeatureTypeSupportedForEntry($dbEntry, $featureType))
+		if (!kReachUtils::isFeatureTypeSupportedForEntry($dbEntry, $featureType))
 		{
 			throw new KalturaAPIException(KalturaReachErrors::FEATURE_TYPE_NOT_SUPPORTED_FOR_ENTRY, $featureType, $entryVendorTask->entryId);
 		}
 
+		if ($dbVendorCatalogItem instanceof VendorTranslationCatalogItem && $dbVendorCatalogItem->getRequireSource())
```
inbal-ben-david, line 71: i think we better have this in a function in reachUtils (isSourceCaptionRequired or similar) which will run the checks and return true/false and throw the error based on it. like we have in the validations in the lines above.

inbal-ben-david (approved):

PR #12850, opened 2024-08-22
SUP-43987: fix the retry when uploading via s3 multipart upload

amiras89:


infra/storage/shared_file_system_managers/kS3SharedFileSystemMgr.php
```diff
@@ -251,6 +252,20 @@ private function doPutFileHelper($filePath , $fileContent, $params)
 		catch (Exception $e)
 		{
 			KalturaLog::warning("Failed to uploaded to s3, info with message: " . $e->getMessage());
+			
+			if ($e instanceof S3MultipartUploadException)
```
yossipapi, line 256: In this case of a resume do we still need the external retry which runs in the outer loop?

yossipapi (changes requested):

infra/storage/shared_file_system_managers/kS3SharedFileSystemMgr.php
amiras89: Due to the way we wrote the S3 and OCI wrappers - we handle retries on the 'parent' function "doPutFileContent" or "doMoveLocalToShared" in that case.

If you call "doPutFileHelper" directly it will not handle the resume.
It must be called from the 'parent' function which will re-call "doPutFileHelper" after the exception with the added 'state' value.

To make this function 'independent' I needed to either implement an internal retry + a recursive call
Move the retries from the parents to "doPutFileHelper"
I feel it is not needed as the way we call "doPutFileHelper" is always from a parent that implement a retry.

infra/storage/shared_file_system_managers/kS3SharedFileSystemMgr.php
```diff
@@ -76,7 +78,6 @@ public function __construct(array $options = null)
 			$this->endPoint = isset($options['endPoint']) ? $options['endPoint'] : null;
 			$this->accessKeySecret = isset($options['accessKeySecret']) ? $options['accessKeySecret'] : null;
 			$this->accessKeyId = isset($options['accessKeyId']) ? $options['accessKeyId'] : null;
-			$this->endPoint = isset($options['endPoint']) ? $options['endPoint'] : null;
```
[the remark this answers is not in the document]
amiras89: It duplicates line 76

yossipapi (approved):

PR #12857, opened 2024-08-28
FOUN-1910: doSelectStmt does not use query cache so use doSelect and …

yossipapi:
…iterate over object to return the ids

### DIFF
--- a/alpha/lib/model/UserEntryPeer.php
+++ b/alpha/lib/model/UserEntryPeer.php
@@ -70,11 +70,8 @@ class UserEntryPeer extends BaseUserEntryPeer {
 	public static function getEntryIdsByFilter($limit, $offset = 0, $filter = null, $entryIds = array())
 	{
 		$userEntryCriteria = new Criteria();
-		
-		$userEntryCriteria->addSelectColumn(self::ENTRY_ID);
 		if($filter)
 			$filter->attachToCriteria($userEntryCriteria);
-		
 		if(count ($entryIds))
 		{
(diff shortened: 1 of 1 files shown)


alpha/lib/model/UserEntryPeer.php
```diff
@@ -84,8 +81,12 @@ public static function getEntryIdsByFilter($limit, $offset = 0, $filter = null,
 		$userEntryCriteria->setLimit($limit);
 		$userEntryCriteria->setOffset($offset);
 		
-		$stmt = self::doSelectStmt($userEntryCriteria);
-		$ids = $stmt->fetchAll(PDO::FETCH_COLUMN);
+		$result = UserEntryPeer::doSelect($userEntryCriteria);
+		
+		$ids = array();
+		foreach ($result as $res) {
```
amiras89, line 87: We probably want to put the curly brackets at the next line

Other then that looks OK

amiras89 (approved):

alpha/lib/model/UserEntryPeer.php
```diff
@@ -70,11 +70,8 @@ public static function getOMClass($row, $colnum)
 	public static function getEntryIdsByFilter($limit, $offset = 0, $filter = null, $entryIds = array())
 	{
 		$userEntryCriteria = new Criteria();
-		
-		$userEntryCriteria->addSelectColumn(self::ENTRY_ID);
 		if($filter)
 			$filter->attachToCriteria($userEntryCriteria);
-
```
amiras89, line 77: I prefer the extra line break, just for esthetics

PR #12872, opened 2024-09-05
KME-2117: allow setting of templateRoomEntry from "global" partner

Amirhalfon1:


inbal-ben-david (approved):

plugins/room/lib/api/KalturaRoomEntry.php
```diff
@@ -81,6 +81,20 @@ public function validateTemplateRoomEntry()
 		}
 	}
 
+	private function retrieveTemplateRoomEntry($entryId)
```
yossipapi, line 84: Please move this to the appropriate peer class
We shouldn't add DB access methods in the Kaltura class

PR #12900, opened 2024-09-22
REACH2-1246: Search for partner IDs that a catalog item is configured on

drorsou:
image


yossipapi:
Why do we need a new screen for this?
Why not add this filter to the already existing "partner catalog items" page? we currently have there filter by partnerId but I believe we can easily add the option to filter by catalogItemId?

PR #12928, opened 2024-10-09
PLAT-24985: Add caching headers to redirect response

MosheMaorKaltura:


alpha/apps/kaltura/modules/extwidget/actions/embedIframeJsAction.class.php
```diff
@@ -178,6 +178,8 @@ private function redirectToV7($v7Id, $v2UiConfId, $partnerId, $shouldTranslatePl
 			return;
 		}
 
+		requestUtils::sendCachingHeaders(600, false, time());
```
yossipapi, line 181: Do you think it would be a good idea to also add
kFile::cacheRedirect($url);

inbal-ben-david (approved):

PR #12936, opened 2024-10-20
Ursa 21.1.0 plat 24980 live translation

danbar0kal:


plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -453,49 +453,68 @@ public function postSave(PropelPDO $con = null)
 
 	public function addSchedulingData()
 	{
+		$language = '';
```
inbal-ben-david, line 456: dont really see a reason to define it here, you already allocate it in both places you use it

plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -453,49 +453,68 @@ public function postSave(PropelPDO $con = null)
...
+		$feature = null;
```
inbal-ben-david, line 457: this var is not in use

plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -453,49 +453,68 @@ public function postSave(PropelPDO $con = null)
...
+		/* @var $connectedEvent LiveStreamScheduleEvent */
+		$connectedEvent = $taskData->getScheduleEvent();
+		/* @var $vendorCatalogItem VendorCatalogItem */
+		$vendorCatalogItem = VendorCatalogItemPeer::retrieveByPK($this->getCatalogItemId());
+
 		switch ($this->getServiceFeature())
 		{
+			case KalturaVendorServiceFeature::LIVE_TRANSLATION:
+			{
+				$language = languageCodeManager::getThreeCodeFromKalturaName($vendorCatalogItem->getTargetLanguage());
```
inbal-ben-david, line 469: since either way we set the language in addLanguageToFeature i think we can move the language extraction to inside the function and just pass to it the source/target language

plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -453,49 +453,68 @@ public function postSave(PropelPDO $con = null)
...
-
-				$feature = new LiveCaptionFeature();
-				$feature->setPreStartTime($connectedEvent->getStartDate(null) - $taskData->getStartDate());
-				$feature->setPostEndTime($taskData->getEndDate() - $connectedEvent->getEndDate(null));
-				$feature->setSystemName(LiveCaptionFeature::defaultName(LiveFeature::REACH_FEATURE_PREFIX . "-{$this->getId()}"));
-				$feature->setLanguage($language);
-
-				$connectedEvent->addFeature($feature, true);
-				$connectedEvent->save();
+				$this->addLanguageToFeature('LiveCaptionFeature', $connectedEvent, $taskData, $language);
```
inbal-ben-david, line 475: at this point i dont see a reason to pass LiveCaptionFeature as param since on both cases we use it

plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -453,49 +453,68 @@ public function postSave(PropelPDO $con = null)
...
 				break;
 		}
 	}
 
+	protected function addLanguageToFeature($featureType, &$connectedEvent, $taskData, $language,)
+	{
+		$feature = new $featureType();
+		$feature->setPreStartTime($connectedEvent->getStartDate(null) - $taskData->getStartDate());
+		$feature->setPostEndTime($taskData->getEndDate() - $connectedEvent->getEndDate(null));
+		$feature->setSystemName(LiveCaptionFeature::defaultName(LiveFeature::REACH_FEATURE_PREFIX . "-{$this->getId()}"));
```
inbal-ben-david, line 485: just like my comment in line 475, if you already explicitly call to live caption feature here there is no point providing the type from outside

plugins/reach/lib/model/EntryVendorTask.php
danbar0kal: in that case, the switch moves to the helper function, while in the future we might need to separate between other features. The division, in my eyes, will be blurry again.

plugins/reach/lib/model/EntryVendorTask.php
danbar0kal: you are right, but i see it as preparation to maybe someday in the future we will want to pass something different

plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -453,49 +453,68 @@ public function postSave(PropelPDO $con = null)
...
+		$feature->setLanguage($language);
+		$connectedEvent->addFeature($feature, true);
+		$connectedEvent->save();
+	}
+
 	public function removeSchedulingData()
 	{
+		$liveFeatureName = '';
```
drorsou, line 493: Unused var

drorsou (approved):

david-winder-kaltura (approved):

PR #12937, opened 2024-10-20
No-Plat: Source map will not return by default + serve latest bundle …

yossipapi:
…if failed to create new one

### DIFF
--- a/alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
+++ b/alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
@@ -16,6 +16,7 @@ class embedPlaykitJsAction extends sfAction
 	const REGENERATE_PARAM_NAME = "regenerate";
 	const IFRAME_EMBED_PARAM_NAME = "iframeembed";
 	const AUTO_EMBED_PARAM_NAME = "autoembed";
+	const INCLUDE_SOURCE_MAP_PARAM_NAME = 'includeSourceMap';
 	const LATEST = "{latest}";
 	const BETA = "{beta}";
 	const CANARY = "{canary}";
@@ -47,6 +48,7 @@ class embedPlaykitJsAction extends sfAction
 	private $playerConfig = null;
 	private $uiConfUpdatedAt = null;
 	private $regenerate = false;
+	private $includeSourceMap = "false";
 	private $uiConfTags = array(self::PLAYER_V3_VERSIONS_TAG);
 
 	public function execute()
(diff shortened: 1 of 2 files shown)


alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -56,16 +58,41 @@ public function execute()
...
-			list($bundleContent, $i18nContent, $extraModulesNames) = kLock::runLocked($this->bundle_name, array("embedPlaykitJsAction", "buildBundleLocked"), array($this), 2, 30);
+			try
+			{
+				list($bundleContent, $i18nContent, $extraModulesNames, $updateUiConfBundleCacheKey) = kLock::runLocked($this->bundle_name, array("embedPlaykitJsAction", "buildBundleLocked"), array($this), 2, 30);
+			}
+			catch (kCoreException $ex)
+			{
+				switch ($ex->getCode())
+				{
+					case kCoreException::LOCK_TIMED_OUT:
```
MosheMaorKaltura, line 76: Does this case cover the 2 cases? case 1 - Bundle is in action and new requests are rejected - 5xx error, 
case 2 - Bundler failed / reset and returns 4xx error?

The only case that should return error is when the bundled content is missing some of the pieces and user gets explicit error in the response

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -865,4 +901,54 @@ public function getRequestParameter($name, $default = null)
...
+			{
+				KalturaLog::debug("Skipping ui conf update to avoid unnecessary system load");
+			}
+		}
+	}
+	
+	protected function tryServingExistingCacheVersion($context, $errCode, $message = null)
+	{
+		$existingCacheVersion = $context->uiConf->getCurrentCacheKey();
+		KalturaLog::debug("TTT: existingCacheVersion [$existingCacheVersion]");
```
MosheMaorKaltura, line 932: TTT: ???

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -865,4 +901,54 @@ public function getRequestParameter($name, $default = null)
...
+		{
+			KExternalErrors::dieError($errCode, $message);
+		}
+		
+		$i18nContent = null;
+		$extraModulesNames = null;
+		$bundleContent = $context->bundleCache->get($existingCacheVersion);
+		if ($bundleContent)
+		{
+			$i18nContent = $context->bundleCache->get($existingCacheVersion . "_i18n");
```
MosheMaorKaltura, line 943: use '

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -47,6 +48,7 @@ class embedPlaykitJsAction extends sfAction
 	private $playerConfig = null;
 	private $uiConfUpdatedAt = null;
 	private $regenerate = false;
+	private $includeSourceMap = "false";
```
MosheMaorKaltura, line 51: I would suggest to load this default value from config, so in case we are catastrophy mode, we can change this value fast, but in general - hold it with value **true**

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -865,4 +901,54 @@ public function getRequestParameter($name, $default = null)
 		$returnValue = parent::getRequestParameter($name, $default);
 		return $returnValue ? $returnValue : $default;
 	}
+	
+	protected function saveCurrentBundleCacheKey($context)
+	{
+		//Avoid updating current cache hash on the ui conf if dynamic params affecting the result are being sent
+		$versions = $this->getRequestParameter(self::VERSIONS_PARAM_NAME);
+		if($versions)
```
MosheMaorKaltura, line 909: Most KMS related calls are using versions, so not sure why to remove this, assume that all the case coming from a website to load UI conf, may constantly come with versions

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
```diff
@@ -865,4 +901,54 @@ public function getRequestParameter($name, $default = null)
...
+			else
+			{
+				KalturaLog::debug("Skipping ui conf update to avoid unnecessary system load");
+			}
+		}
+	}
+	
+	protected function tryServingExistingCacheVersion($context, $errCode, $message = null)
+	{
+		$existingCacheVersion = $context->uiConf->getCurrentCacheKey();
```
MosheMaorKaltura, line 931: general note - 
Saving current cache key on UI conf is not essential, as the current cache key can be calculated 'on the spot' from the versions that are currently required by it ( hash( ui confi versions ) while the action of saving new version on UICONF in DB, will create load on DB any time new version will be released.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
yossipapi: It doesn't make sense the default will be true as the entire purpose of this change is that the source map won't be returned by default.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
yossipapi: As the list of versions sent by KMS is dynamic and can change I want to avoid constantly changing the hash values associated with the player. When sending versions this means there is no constant bundle hash and this is a use case I don't wont to handle.

alpha/apps/kaltura/modules/extwidget/actions/embedPlaykitJsAction.class.php
yossipapi: The entire point is that the current hash of versions is not in the cache, this is why we got to this point in the first place. I need the old hash of the UI conf versions, which I currently don't have a way to get. If in the future we will have the old version pam also stored we can re-visit this.

PR #12959, opened 2024-11-05
ONPREM-1076 Fix PDOException for unknown column in sphinx

GElkayam:
This pull request fixes an issue where a PDOException was being thrown when encountering an unknown column in dynamic_attributes on kaltura_metadata. The code now catches the exception and provides appropriate error messages when updating the affected items.

### DIFF
--- a/alpha/scripts/utils/updatePartnerEntries2Sphinx.php
+++ b/alpha/scripts/utils/updatePartnerEntries2Sphinx.php
@@ -51,8 +51,16 @@ for (;;)
 		if ($mode == 'execute')
 		{
 			echo "Updating [". $item->getId() ."] ... "; 
-			$sphinx->saveToSphinx($item, false, true);
-			echo "Done\n";
+			try {
+				$sphinx->saveToSphinx($item, false, true);
+				echo 'Done' . PHP_EOL;
+			} catch (PDOException $e) {
+				if (strpos($e->getMessage(), 'unknown column') !== false) {
+					echo '[unknown column] error occurred while updating [' . $item->getId() . ']: ' . $e->getMessage() . PHP_EOL;
+				} else {
+					echo 'Failed to update [' . $item->getId() . ']: ' . $e->getMessage() . PHP_EOL;
+				}
+			}
 		}
 		else
 		{
--- a/alpha/scripts/utils/updatePartnerEntries2Sphinx.php
+++ b/alpha/scripts/utils/updatePartnerEntries2Sphinx.php
@@ -50,14 +50,20 @@ for (;;)
 		usleep(100);
 		if ($mode == 'execute')
 		{
-			echo "Updating [". $item->getId() ."] ... "; 
-			try {
+			try 
+			{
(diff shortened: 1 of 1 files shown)


navarro967 (approved):
Looks good to me

alpha/scripts/utils/updatePartnerEntries2Sphinx.php
```diff
@@ -51,8 +51,16 @@
 		if ($mode == 'execute')
 		{
 			echo "Updating [". $item->getId() ."] ... "; 
-			$sphinx->saveToSphinx($item, false, true);
-			echo "Done\n";
+			try {
```
amiras89, line 54: Coding conventions: we set curly brackets at new line
(I would also use the opportunity to align other if statements that don't have (single if line) with curly brackets
at 2011 when script was written it was before we set conventions

alpha/scripts/utils/updatePartnerEntries2Sphinx.php
```diff
@@ -51,8 +51,16 @@
 		if ($mode == 'execute')
 		{
 			echo "Updating [". $item->getId() ."] ... ";
```
amiras89, line 53: I would place the echo inside of try stmnt

alpha/scripts/utils/updatePartnerEntries2Sphinx.php
```diff
@@ -51,8 +51,16 @@
 		if ($mode == 'execute')
 		{
 			echo "Updating [". $item->getId() ."] ... "; 
-			$sphinx->saveToSphinx($item, false, true);
-			echo "Done\n";
+			try {
+				$sphinx->saveToSphinx($item, false, true);
+				echo 'Done' . PHP_EOL;
+			} catch (PDOException $e) {
```
amiras89, line 57: why PDOException specifically?
I think we would be better catch Exception and inside we can check, like
			catch (Exception $e)
			{
				if ($e instanceof PDOException && strpos($e->getMessage(), 'unknown column') !== false)
.
.

alpha/scripts/utils/updatePartnerEntries2Sphinx.php
```diff
@@ -51,8 +51,16 @@
...
+				if (strpos($e->getMessage(), 'unknown column') !== false) {
```
amiras89, line 58: php8 introduced 'str_contains' which is more readable:
str_contains($e->getMessage(), 'unknown column')

alpha/scripts/utils/updatePartnerEntries2Sphinx.php
```diff
@@ -51,8 +51,16 @@
...
+					echo '[unknown column] error occurred while updating [' . $item->getId() . ']: ' . $e->getMessage() . PHP_EOL;
```
amiras89, line 59: I think we can take the opportunity and change all 'echo' to KalturaLog
The script was written in 2011, KalturaLog was added at 2013 iirc

amiras89 (approved):

PR #12964, opened 2024-11-07
PLAT-24975: Validate user when editing Annotation Cue Points on entries

drorsou:


danbar0kal (approved):

plugins/cue_points/annotation/lib/api/KalturaAnnotation.php
```diff
@@ -197,37 +197,47 @@ public function updateEndTimeAndDuration($cuePoint)
...
-            else
-            {
-                if ($dbParentCuePoint->getEntryId() != $this->entryId)
-                    throw new KalturaAPIException(KalturaCuePointErrors::PARENT_ANNOTATION_DO_NOT_BELONG_TO_THE_SAME_ENTRY);
-            }
-        }
-    }
+	protected function validateEntryEntitlement(entry $dbEntry, $tags)
+	{
+		if (!kEntitlementUtils::isEntryEntitled($dbEntry) && str_contains($tags, 'hotspots'))
```
muli, line 237: If I have a cue-point with the following tags:
- `coolhotspots`
- `tag2`
- `tag3`

which translates into `coolhotspots,tag2,tag3` in the DB (and in the API entry object) - this restriction will apply there, which is unexpected.
I think that you should do something like

```php
$tags = explode(',', $tags);
if (in_array('hotspots', $tags) && !kEntitlementUtils::isEntryEntitled($dbEntry)) {
    // throw error
}
```

Also notice the order of the conditions - `isEntryEntitled` will not execute if `hotspots` isn't one of the tags, which might be performance improvement.

muli (commented):
@inbal-ben-david, @drorsou - please see my comment on this one.

plugins/cue_points/annotation/lib/api/KalturaAnnotation.php
drorsou: Opened a new PR #12969

PR #12966, opened 2024-11-07
allow breakout room with blocked parent entry user

danbar0kal:


alpha/lib/model/entry.php
```diff
@@ -2132,14 +2132,22 @@ public function setEntitledPusersEdit($v)
...
+				$this->blockedUserFlow($kuser->getId(), $parentEntry);
 			}
 
 			$entitledUserPuserEdit[$kuser->getId()] = $kuser->getPuserId();
 		}
 
 		$this->putInCustomData ( "entitledUserPuserEdit" , serialize($entitledUserPuserEdit) );
 	}
+
+	protected function blockedUserFlow($kuserId, $parentEntry = null)
```
inbal-ben-david, line 2144: i would make this a boolean function and use it in the set entitled view/edit.. then throw the error from there.
its less common to create a function that if -> throws
then you can also give it a more meanningfull name like isBlockedUserAllowed

alpha/lib/model/entry.php
```diff
@@ -2274,9 +2286,11 @@ public function setEntitledPusersPublish($v)
 			if (!$kuser)
 				throw new kCoreException('Invalid user id', kCoreException::INVALID_USER_ID);
 
-			if ($kuser->getStatus() === KuserStatus::BLOCKED && !$this->isEntitledKuserPublish($kuser->getId()))
+			if ($kuser->getStatus() === KuserStatus::BLOCKED &&
+				!$this->isEntitledKuserEdit($kuser->getId()) &&
+				!$this->isBlockedUserAllowed($kuser->getId(), $parentEntry))
```
drorsou, line 2291: remove 'not' !

alpha/lib/model/entry.php
```diff
@@ -2128,18 +2128,27 @@ public function setEntitledPusersEdit($v)
...
+				throw new kCoreException('Cannot add a blocked user', kCoreException::USER_BLOCKED);
 			}
 
 			$entitledUserPuserEdit[$kuser->getId()] = $kuser->getPuserId();
 		}
 
 		$this->putInCustomData ( "entitledUserPuserEdit" , serialize($entitledUserPuserEdit) );
 	}
+
+	protected function isBlockedUserAllowed($kuserId, $parentEntry = null)
```
drorsou, line 2148: Change function name to 'isOwnerOfEntry'

drorsou (approved):

PR #12971, opened 2024-11-12
plat-25016: Add checklist microservice

MosheMaorKaltura:


alpha/apps/kaltura/lib/microservice/MicroServiceChecklist.php
```diff
@@ -0,0 +1,26 @@
...
+	public static $host = 'checklist';
+	public function __construct()
+	{
+		$this->hostName = self::$host;
+		parent::__construct();
+	}
+
+	public static function buildServiceUrl($hostName, $serviceName , $isApi = true)
+	{
+		$url = parent::buildServiceUrl($hostName, false, false);
```
drorsou, line 17: Better to pass serviceName as null, since buildServiceUrl expects the second param to be string

drorsou (approved):

PR #12985, opened 2024-11-19
SUP-45464: When indexing entry to elastic disable entitlement when ge…

yossipapi:
…tting categoris


amiras89 (approved):
tested it - looks good

PR #12992, opened 2024-11-24
SUP-43907 Use multiple pages to fetch results, up to MAX_ESEARCH_RESU…

bw-kaltura:
…LTS, instead of just one page

### DIFF
--- a/alpha/apps/kaltura/lib/reports/kKavaReportsMgr.class.php
+++ b/alpha/apps/kaltura/lib/reports/kKavaReportsMgr.class.php
@@ -3074,14 +3074,22 @@ class kKavaReportsMgr extends kKavaBase
 			$entry_search = new kEntrySearch();
 			$entry_search->setFilterOnlyContext();
 			$pager = new kPager();
-			$pager->setPageSize(self::MAX_ESEARCH_RESULTS);
-			$elastic_results = $entry_search->doSearch($input_filter->entry_operator, $pager);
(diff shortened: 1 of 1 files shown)


alpha/apps/kaltura/lib/reports/kKavaReportsMgr.class.php
```diff
@@ -3074,14 +3074,23 @@ protected static function getDruidFilter($partner_id, $report_def, $input_filter
...
-			{
-				throw new kCoreException('Search is to general', kCoreException::SEARCH_TOO_GENERAL);
-			}
+            		$page_index = 0;
+            		$elastic_entry_ids = array();
+            		do 
+			{
+                		$page_index++;
+                	    	$pager->setPageIndex($page_index);
+                    	    	$pager->setPageSize(min(self::MAX_ESEARCH_RESULTS - count($elastic_entry_ids), kPager::MAX_PAGE_SIZE));
```
inbal-ben-david, line 3083: why not just taking max page size?
if you going page by page you will just get the additional entries on the second page

bw-kaltura: This handles cases where MAX_ESEARCH_RESULTS isn’t divisible by MAX_PAGE_SIZE (500), so we don’t fetch more than the limit on the last run.

inbal-ben-david (approved):

PR #12996, opened 2024-11-25
SUP-45718: Add partner configuration for disabling publicity of uncategorized entries

drorsou:


configurations/admin.template.ini
```diff
@@ -1251,6 +1251,12 @@ moduls.embedShareStream.label = "Enable embed & social stream in EP"
 moduls.embedShareStream.permissionName = FEATURE_EMBED_SHARE_SOCIAL_STREAM
 moduls.embedShareStream.group = GROUP_ENABLE_DISABLE_FEATURES
 
+moduls.uncategorizedEntriesValidation.enabled = true
+moduls.uncategorizedEntriesValidation.permissionType = 2
+moduls.uncategorizedEntriesValidation.label = "Do not auto allow uncategorized entries"
```
inbal-ben-david, line 1256: i think it doesnt fully represent what it does, as it relevant only to entitlements.
i would change it to something like "Disable category entry entitlement check"

alpha/apps/kaltura/lib/kEntitlementUtils.php
```diff
@@ -626,7 +626,7 @@ protected static function validateEntryEntitlements(entry $entry, $kuserId, $ks,
 		// entry that doesn't belong to any category is public
 		//when ks is not provided - the entry is still public (for example - download action)
 		$categoryEntry = categoryEntryPeer::retrieveOneActiveByEntryId($entry->getId());
-		if (!$categoryEntry)
+		if (!$categoryEntry && !PermissionPeer::isValidForPartner(PermissionName::FEATURE_UNCATEGORIZED_ENTRIES_VALIDATION, kCurrentContext::getCurrentPartnerId()))
```
inbal-ben-david, line 629: if the check is disabled and we have no category entries why even continue to the next checks and not return immediately?

PR #13007, opened 2024-12-03
SUP-45225: Prevent automatic captions when we clone an entry that has captions.

MaimanaGhnem:


alpha/apps/kaltura/lib/myEntryUtils.class.php
```diff
@@ -1551,6 +1551,10 @@ public static function copyEntryData(entry $entry, entry $targetEntry, $copyFlav
 		{
 			if (self::shouldCopyAsset($sourceAsset, $copyFlavors, $copyCaptions, $entry->getId()))
 			{
+                if($copyCaptions && $sourceAsset instanceof captionAsset)
```
drorsou, line 1554: - $copyCaptions is verified already inside _self::shouldCopyAsset_, so no need to check it again
- Add $targetEntry->save so change will take effect
- Change indentation to tabs

drorsou (approved):

PR #13008, opened 2024-12-04
PSVAMB-74281 - abandon tas execution for entries that have PreventAut…

hilak:
…oCategorization set to True

### DIFF
--- a/plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
+++ b/plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
@@ -1,7 +1,5 @@
 <?php
 
-use utils\scripts\client\KalturaPlugins\KalturaTranscriptAsset;
-
 require_once (__DIR__ . DIRECTORY_SEPARATOR . 'OpenCalaisConstants.php');
 require_once (__DIR__ . DIRECTORY_SEPARATOR . 'RuleEngine.php');
 
--- a/plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
+++ b/plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
@@ -1,5 +1,7 @@
 <?php
 
+use utils\scripts\client\KalturaPlugins\KalturaTranscriptAsset;
+
 require_once (__DIR__ . DIRECTORY_SEPARATOR . 'OpenCalaisConstants.php');
 require_once (__DIR__ . DIRECTORY_SEPARATOR . 'RuleEngine.php');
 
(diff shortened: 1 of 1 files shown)


TalDubovKaltura (approved):
@hilak If I understand the new logic correctly, you are checking if the entry has customMetadataProfile (systemName=ShowDataExtractionWorkflow), and the field PreventAutomatedCategorization includes the value True and then you are skipping this entry.
Not sure where this field has been populated but the logic makes sense to me

plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
```diff
@@ -59,7 +63,20 @@ function handleTask(KalturaEntryVendorTask $vendorTask)
         KBatchBase::impersonate($vendorTask->partnerId);
 
         try {
-            $mappingProfileId = $this->getMappingMetadataProfileId();
+            // If entry has PreventAutomatedCategorization set to True --> no tags should be created for the entry.
+            $autoCategorizationMetadataProfileId = $this->getMetaDataProfileId(self::SHOW_DATA_EXTRACTION_WORKFLOW_SYSTEM_NAME);
+            $preventAutoCategorizationMetadata = $this->retrieveMetadataObjectsByMetadataProfileAndObjectId($autoCategorizationMetadataProfileId, KalturaMetadataObjectType::ENTRY ,$vendorTask->entryId);
```
inbal-ben-david, line 68: i would check that this profile actually returned before passing to retrieveMetadataObjectsByMetadataProfileAndObjectId, to avoid the list call for a null metadata profile afterwards

plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
```diff
@@ -59,7 +63,20 @@ function handleTask(KalturaEntryVendorTask $vendorTask)
...
+            if ($preventAutoCategorizationMetadata->totalCount)
```
inbal-ben-david, line 69: here as well i would check the object is not null before accessing its fields

plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
hilak: @inbal-ben-david  list requests never return null objects. Unless you're using a invalid KS, you'll just get an empty list response.

plugins/reach/plugins/OpenCalais/batch/kReachVendorTaskOpenCalaisProcessorEngine.php
hilak: @inbal-ben-david  see above.

inbal-ben-david (approved):

PR #13026, opened 2024-12-19
KAI-121: add duration limit and analysis type to videoAnalysis catalogItem

Amirhalfon1:
KAI-121: duration limit - the limit (in sec) of a video to be analyzed 
KAI-117: type - the analysis type (currently only OCR supported)

### DIFF
--- a/plugins/reach/lib/model/data/VendorVideoAnalysisCatalogItem.php
+++ b/plugins/reach/lib/model/data/VendorVideoAnalysisCatalogItem.php
@@ -5,6 +5,9 @@
  */
 class VendorVideoAnalysisCatalogItem extends VendorCatalogItem
 {
+	const CUSTOM_DATA_VIDEO_ANALYSIS_TYPE = 'video_analysis_type';
+	const CUSTOM_DATA_MAX_VIDEO_DURATION = 'max_video_duration';
+
 	public function applyDefaultValues(): void
 	{
 		$this->setServiceFeature(VendorServiceFeature::VIDEO_ANALYSIS);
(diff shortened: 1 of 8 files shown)


NadavHarnik (changes requested):

plugins/reach/lib/model/data/VendorVideoAnalysisCatalogItem.php
```diff
@@ -20,4 +23,33 @@ public function isEntryTypeSupported($type, $mediaType = null): bool
...
 	}
+
+	public function setVideoAnalysisType($v)
+	{
+		$this->putInCustomData(self::CUSTOM_DATA_VIDEO_ANALYSIS_TYPE, $v);
+	}
+
+	public function getVideoAnalysisType()
+	{
+		return $this->getFromCustomData(self::CUSTOM_DATA_VIDEO_ANALYSIS_TYPE, null, true);
```
inbal-ben-david, line 34: why putting true here as the default values?
i assume you want it to be parsed to VendorVideoAnalysisType::OCR so better just mention it with the enum

plugins/reach/lib/model/data/VendorVideoAnalysisCatalogItem.php
```diff
@@ -20,4 +23,33 @@ public function isEntryTypeSupported($type, $mediaType = null): bool
...
+	}
+
+	public function setMaxVideoDuration($v)
+	{
+		$this->putInCustomData(self::CUSTOM_DATA_MAX_VIDEO_DURATION, $v);
+	}
+
+	public function getMaxVideoDuration()
+	{
+		return $this->getFromCustomData(self::CUSTOM_DATA_MAX_VIDEO_DURATION, null, true);
```
inbal-ben-david, line 44: the default values doesnt makes sense to me here as well

Amirhalfon1: fixed

NadavHarnik (approved):

PR #13057, opened 2025-01-09
PSVAMB-69071 - add support for automated reach flows on live schedule…

hilak:
… events

### DIFF
--- a/plugins/reach/lib/kReachManager.php
+++ b/plugins/reach/lib/kReachManager.php
@@ -34,6 +34,7 @@ class kReachManager implements kObjectChangedEventConsumer, kObjectCreatedEventC
 			"categoryEntry" => objectType::CATEGORY_ENTRY,
 			"CaptionAsset" => CaptionAssetEventNotificationsPlugin::getEventNotificationEventObjectTypeCoreValue(CaptionAssetEventNotificationEventObjectType::CAPTION_ASSET),
 			"TranscriptAsset" => TranscriptAssetEventNotificationsPlugin::getEventNotificationEventObjectTypeCoreValue(TranscriptAssetEventNotificationEventObjectType::TRANSCRIPT_ASSET),
+			"LiveStreamScheduleEvent" => ScheduleEventNotificationsPlugin::getEventNotificationEventObjectTypeCoreValue(ScheduleEventNotificationEventObjectType::SCHEDULE_EVENT),
 			);
 
 		if (isset($mapObjectType[$eventObjectClassName]))
--- a/plugins/reach/lib/model/data/VendorLiveCaptionCatalogItem.php
+++ b/plugins/reach/lib/model/data/VendorLiveCaptionCatalogItem.php
@@ -12,19 +12,27 @@ class VendorLiveCaptionCatalogItem extends VendorLiveCatalogItem
 
 	public function getTaskJobData($object)
 	{
+		$latestEvent = null;
 		if($object instanceof LiveEntry && $object->hasCapability(LiveEntry::LIVE_SCHEDULE_CAPABILITY))
 		{
 			$events = ScheduleEventPeer::retrieveByTemplateEntryIdAndTypes($object->getId(), [ScheduleEventType::LIVE_STREAM]);
 
-			$data = new kScheduledVendorTaskData();
-			if(count($events))
-			{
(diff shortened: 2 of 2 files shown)


plugins/reach/lib/kReachManager.php
```diff
@@ -34,6 +34,7 @@ protected function getObjectType($eventObjectClassName)
 			"categoryEntry" => objectType::CATEGORY_ENTRY,
 			"CaptionAsset" => CaptionAssetEventNotificationsPlugin::getEventNotificationEventObjectTypeCoreValue(CaptionAssetEventNotificationEventObjectType::CAPTION_ASSET),
 			"TranscriptAsset" => TranscriptAssetEventNotificationsPlugin::getEventNotificationEventObjectTypeCoreValue(TranscriptAssetEventNotificationEventObjectType::TRANSCRIPT_ASSET),
+			"LiveStreamScheduleEvent" => ScheduleEventNotificationsPlugin::getEventNotificationEventObjectTypeCoreValue(ScheduleEventNotificationEventObjectType::SCHEDULE_EVENT),
```
david-winder-kaltura, line 37: LiveStreamScheduleEvent -> ScheduleEvent (as that the type you take the enum from)

plugins/reach/lib/model/data/VendorLiveCaptionCatalogItem.php
```diff
@@ -12,19 +12,27 @@ public function applyDefaultValues()
...
-			$data = new kScheduledVendorTaskData();
-			if(count($events))
-			{
-				$latestEvent = $events[count($events)-1];
-				$data->setEntryDuration($latestEvent->getDuration()*1000);
-				$data->setStartDate($latestEvent->getStartDate());
-				$data->setEndDate($latestEvent->getEndDate());
-				$data->setScheduledEventId($latestEvent->getId());
+			if (count($events)) {
+				$latestEvent = $events[count($events) - 1];
```
david-winder-kaltura, line 21: I don't think this is good (know it was here before and still) - as you assume the last event is the one you want to add the data from but that may not be the case. I never review that original PR that insert it.
Anyway, if you move to use event that sent the LiveStreamScheduleEvent why not remove all this part of the object been the live entry?

plugins/reach/lib/kReachManager.php
hilak: @david-winder-kaltura  I've tested this. The objectType that the flow get is LiveStreamScheduleEvent and if I put the generic calssname here, the flow will ignore the event

plugins/reach/lib/model/data/VendorLiveCaptionCatalogItem.php
hilak: @david-winder-kaltura  if you re-use the same live event for additional events later on, why wouldn't you use the latest scheduled event?
I would be inclined to leave the code in case we would need to support live captioning on entry update events in the future, but will defer to @salma2222

plugins/reach/lib/kReachManager.php
david-winder-kaltura: I think it can be confuse but because this map also contain LiveStreamEntry -> entry I guess we can live with that as well

plugins/reach/lib/model/data/VendorLiveCaptionCatalogItem.php
```diff
@@ -12,19 +12,27 @@ public function applyDefaultValues()
 
 	public function getTaskJobData($object)
 	{
+		$latestEvent = null;
 		if($object instanceof LiveEntry && $object->hasCapability(LiveEntry::LIVE_SCHEDULE_CAPABILITY))
 		{
 			$events = ScheduleEventPeer::retrieveByTemplateEntryIdAndTypes($object->getId(), [ScheduleEventType::LIVE_STREAM]);
 
-			$data = new kScheduledVendorTaskData();
```
salma2222, line 19: before, we returned empty kScheduledVendorTaskData when there were no events on the entry
now, we only return filled kScheduledVendorTaskData when we find events and null otherwise

was this intended?

hilak: @salma2222  fixed.

salma2222 (approved):

PR #13087, opened 2025-01-21
Add option to delay notifications when entry added to channel if entr…

danbar0kal:
…y is not ready


plugins/event_notification/lib/model/BatchEventNotificationTemplate.php
```diff
@@ -117,7 +117,29 @@ protected function addEventNotificationDispatchJob($eventNotificationType, kEven
 		$batchJob = kJobsManager::addJob($batchJob, $jobData, $jobType, $eventNotificationType);
 		$jobData->setJobId($batchJob->getId());
 		$batchJob->setData($jobData);
-		
+
+		if ($this->getEventDelayedConditions() && $entryId && $this->areDelayedEventConditionsMet($entryId))
```
danbar0kal, line 121: insert all this logic into a function that return true/false

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,16 @@ public static function addDeliveryTagToEntry(entry $entry)
 			$entry->save();
 		}
 	}
+
+	public static function handleDelayedNotification(entry $entry)
+	{
+		$delayedLockedJob = BatchJobLockPeer::retrieveByEntryIdAndStatus($entry->getEntryId(), BatchJob::BATCHJOB_STATUS_DELAYED);
+		if ($delayedLockedJob)
```
danbar0kal, line 2974: check in the template what is the condition. cant release the job when entry is ready, while the conditioon is something else

plugins/event_notification/lib/model/BatchEventNotificationTemplate.php
salma2222, line 121: put all conditions in one function

alpha/lib/model/BatchJobLockPeer.php
```diff
@@ -243,4 +244,12 @@ public static function retrieveByEntryId($entryId, array $jobTypes = null, array
 	
 	    return self::doSelect($c);
 	}
+
+	public static function retrieveByEntryIdAndStatus($obj_id, $status)
+	{
+		$c = new Criteria();
+		$c->add ( self::ENTRY_ID , $obj_id );
+		$c->add ( self::STATUS , $status );
+		return self::doSelectOne( $c );
```
salma2222, line 253: 1. is default criteria applied here ? 
2. what is we have multiple batchJobs locks ?

alpha/apps/kaltura/lib/batch2/kFlowManager.php
```diff
@@ -686,6 +686,7 @@ public function objectChanged(BaseObject $object, array $modifiedColumns)
 			&& $object->getStatus() == entryStatus::READY)
 		{
 			kFlowHelper::addDeliveryTagToEntry($object);
+			kFlowHelper::handleDelayedNotification($object);
```
salma2222, line 689: when entry becomes ready and there is a delayed job we release the job regardless of the notification template configuration 
1. what if we define different conditions later for notification templates - need to check if the notification template condition if it's indeed entry ready
2. what if we have multiple notification templates with delayed condition - need to distinguish between them

alpha/apps/kaltura/lib/batch2/kFlowManager.php
```diff
@@ -686,6 +686,7 @@ public function objectChanged(BaseObject $object, array $modifiedColumns)
 			&& $object->getStatus() == entryStatus::READY)
```
salma2222, line 686: if condition never applies: 
check if we have cleanup for both batchJob and batchJobLock tables for the delayed jobs

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,39 @@ public static function addDeliveryTagToEntry(entry $entry)
 			$entry->save();
 		}
 	}
+
+	public static function handleDelayedNotification($object)
+	{
+
+		$delayedLockedJobs = match (true)
+		{
+			$object instanceof entry => BatchJobLockPeer::retrieveByEntryIdAndStatus($object->getEntryId(), BatchJob::BATCHJOB_STATUS_DELAYED)
```
gonenradai, line 2976: this will throw Fatal error if the function is called with an object that is not of type `entry`
while right now this is not possible, because it is only called from within a block that only handles entries - it might in the future if someone will call it from another context.
worth either adding a fallback like:
```
true => []
```
write the code differently without using `match`

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,39 @@ public static function addDeliveryTagToEntry(entry $entry)
...
+				/* @var $job BatchJobLock */
+				$delayedJob = BatchJobPeer::retrieveByPK($job->getId());
+				/* @var $job BatchJob */
+				if ($delayedJob)
+				{
+					$jobData = $delayedJob->getData();
+					/* @var $jobData kEventNotificationDispatchJobData */
+					match (true)
+					{
+						$jobData->getEventDelayedConditions() == EventNotificationDelayedConditions::PENDING_ENTRY_READY =>
```
gonenradai, line 2992: same comment here - if something else would ever be returned (or nothing would be returned by the condition) the result will be a Fatal error (`Uncaught UnhandledMatchError: Unhandled match case true`)

plugins/event_notification/lib/model/BatchEventNotificationTemplate.php
```diff
@@ -116,8 +116,39 @@ protected function addEventNotificationDispatchJob($eventNotificationType, kEven
...
+	public function areDelayedEventConditionsMet(&$jobData, $objectId = null)
+	{
+		$delayedEventConditions = $this->getEventDelayedConditions();
+		if ($delayedEventConditions)
+		{
+			switch ($delayedEventConditions)
+			{
+				case EventNotificationDelayedConditions::PENDING_ENTRY_READY:
+				{
+					if ($objectId)
```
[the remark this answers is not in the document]
danbar0kal: i dont know how this might be used in the future. there might be a use case where the batch is not dependent of another object to be delayed or be released in the future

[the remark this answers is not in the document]
gonenradai: OK. I don't see much use of dependency or delaying by things that are not related to some objects (for example - delay to an absolute time in the future) but I am OK with leaving the condition here too for any future case.

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,49 @@ public static function addDeliveryTagToEntry(entry $entry)
 			$entry->save();
 		}
 	}
+
+	public static function handleDelayedNotification($object)
+	{
+		$delayedLockedJobs = [];
+		switch (true)
+		{
+			case ($object instanceof entry):
```
salma2222, line 2976: why not just if condition ?

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,49 @@ public static function addDeliveryTagToEntry(entry $entry)
...
+			{
+				$delayedLockedJobs = BatchJobLockPeer::retrieveByEntryIdAndStatus($object->getEntryId(), BatchJob::BATCHJOB_STATUS_DELAYED);
```
salma2222, line 2978: let's change the naming to delayedJobLocks so it is more clear that the objects are locks and not the jobs themselves

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,49 @@ public static function addDeliveryTagToEntry(entry $entry)
...
+				break;
+			}
+		}
+
+		if (count($delayedLockedJobs) > 0)
```
salma2222, line 2983: not needed, we will not enter the for loop in this case

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,49 @@ public static function addDeliveryTagToEntry(entry $entry)
...
+		{
+			foreach ($delayedLockedJobs as $job)
```
salma2222, line 2985: the name $job is confusing since the object is a jobLock

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,49 @@ public static function addDeliveryTagToEntry(entry $entry)
...
+			{
+				/* @var $job BatchJobLock */
+				$delayedJob = BatchJobPeer::retrieveByPK($job->getId());
+				/* @var $job BatchJob */
```
salma2222, line 2989: $delayedJob not $job

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,49 @@ public static function addDeliveryTagToEntry(entry $entry)
...
+				if ($delayedJob)
+				{
+					$jobData = $delayedJob->getData();
+					if ($jobData)
```
salma2222, line 2993: let make sure that this job is of type notification since we need the jobData to be kEventNotificationDispatchJobData and we will access $jobData->getEventDelayedConditions()

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
```diff
@@ -2967,6 +2967,49 @@ public static function addDeliveryTagToEntry(entry $entry)
...
+					{
+						/* @var $jobData kEventNotificationDispatchJobData */
+						switch (true)
+						{
+							case ($jobData->getEventDelayedConditions() == EventNotificationDelayedConditions::PENDING_ENTRY_READY):
```
salma2222, line 2998: simple "if" condition is more clear :)

plugins/event_notification/lib/model/BatchEventNotificationTemplate.php
```diff
@@ -116,8 +116,39 @@ protected function addEventNotificationDispatchJob($eventNotificationType, kEven
...
+
 		$batchJob->setData($jobData);
-		
-		return kJobsManager::updateBatchJob($batchJob, BatchJob::BATCHJOB_STATUS_PENDING);
+		return kJobsManager::updateBatchJob($batchJob, $batchJobStatus);
+	}
+
+	public function areDelayedEventConditionsMet(&$jobData, $objectId = null)
+	{
+		$delayedEventConditions = $this->getEventDelayedConditions();
```
salma2222, line 132: is delayed conditions (getEventDelayedConditions) meant to be an array of condition types / enum ? 
if it's meant to be just one condition then better give it a singular name, otherwise it's confusing

plugins/event_notification/lib/model/data/kEventNotificationDispatchJobData.php
```diff
@@ -14,14 +14,35 @@ class kEventNotificationDispatchJobData extends kJobData
 	 * @var int
 	 */
 	protected $jobId;
+
+	/**
+	 * @var KalturaEventNotificationDelayedConditions
```
salma2222, line 19: I believe here we need the core object of EventNotificationDelayedConditions since we are in a a core class

plugins/event_notification/lib/model/data/kEventNotificationDispatchJobData.php
```diff
@@ -14,14 +14,35 @@ class kEventNotificationDispatchJobData extends kJobData
...
 	/**
 	 * Define the content dynamic parameters
 	 * 
 	 * @var array<key,value>
 	 */
 	protected $contentParameters;
-	
+
+	/**
+	 * @return KalturaEventNotificationDelayedConditions $eventDelayedConditions
```
salma2222, line 31: same here for core obj

plugins/event_notification/lib/model/data/kEventNotificationDispatchJobData.php
```diff
@@ -14,14 +14,35 @@ class kEventNotificationDispatchJobData extends kJobData
...
+	 */
+	public function getEventDelayedConditions()
+	{
+		return $this->eventDelayedConditions;
+	}
+
+	/**
+	 * @param KalturaEventNotificationDelayedConditions $eventDelayedConditions
```
salma2222, line 39: same here for core obj

release-notes.md
```diff
@@ -1,3 +1,9 @@
+# Ursa-21.8.0
```
salma2222, line 1: Ursa-21.9.0

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
danbar0kal: maybe in the future we will have more conditions concerning other objects, not just entries

alpha/apps/kaltura/lib/batch2/kFlowHelper.php
danbar0kal: not sure i understand

danbar0kal: you men doing this?
if ($jobData && $jobData instanceof kEventNotificationDispatchJobData)

MosheMaorKaltura (approved):

PR #13098, opened 2025-01-25
SUP-46601: v2 to v7 autoembed, add targetId to the call

MosheMaorKaltura:


drorsou (approved):

alpha/apps/kaltura/modules/extwidget/actions/embedIframeJsAction.class.php
```diff
@@ -187,13 +187,20 @@ private function redirectToV7($v7Id, $v2UiConfId, $partnerId, $shouldTranslatePl
 			return;
 		}
 
+		$targetId='';
+		//In v2 'playerId' turns to 'targetId' in v7
+		if($this->getRequestParameter('autoembed'))
+		{
+			$targetId = '&targetId=' . $this->getRequestParameter('playerId');
```
salma2222, line 194: can $this->getRequestParameter('playerId') be empty ? if so then we want $targetId=''

MosheMaorKaltura: if its empty then we do not what targetId to be anything, no key nor value

salma2222: if it's empty the code will do
$targetId = '&targetId='
and not 
$targetId=''
as needed

PR #13099, opened 2025-01-26
PLAT-25073: Move extendedStatus field to viewHistoryUserEntry

drorsou:


MosheMaorKaltura (approved):

plugins/content/entry_permission_level/lib/api/filters/KalturaPermissionLevelUserEntryFilter.php
```diff
@@ -1,14 +1,64 @@
...
+	/**
+	 * @var KalturaPermissionLevelArray
+	 */
+	public $permissionLevels;
+
+	/**
+	 * @var int
+	 * @readonly
+	 */
+	public $permissionLevelsBitmask;
```
yossipapi, line 17: 2 comments:
1. Why id permissionLevelsBitmask public or even exposed in the API? It should be private an internal only. The only thing the client should send is the array and we should calc the bitmask internally only.
2. I don't get why isn't the filters permissionLevels exposed in the map_between_objects? how does it even work?

alpha/apps/kaltura/lib/model/objectfilters/UserEntryFilter.php
```diff
@@ -28,6 +28,7 @@ protected function init()
 			"_eq_extended_status",
 			"_in_extended_status",
 			"_notin_extended_status",
+			"_bitor_extended_status",
```
yossipapi, line 31: Why is it UserEntryFilter.php it should be on a new class called PermissionLevelUserEntryFilter that extends this class.

alpha/apps/kaltura/lib/model/objectfilters/UserEntryFilter.php
```diff
@@ -28,6 +28,7 @@ protected function init()
 			"_eq_extended_status",
```
yossipapi, line 28: Now that extended_status is not part of the basic userEntry and you moved it to  ViewHistoryUserEntryFilter.php it should not be here

PR #13100, opened 2025-01-26
ONPREM-1444 Support ranged indexing for Sphinx and ElasticSearch

GElkayam:


GElkayam:
Had a chat with @yossipapi  and approved this approach.

PR #13112, opened 2025-02-02
VCP-19989: add S3 DF assume role support for customer or Kaltura ARN

amiras89:


amiras89:
// TODO implement doLoginIAMRole in all classes that extends kFileTransferMgr once approved

infra/storage/RefreshableRole.class.php
```diff
@@ -20,9 +20,14 @@ class RefreshableRole
 	const ROLE_SESSION_NAME_PREFIX = "kaltura_s3_access_";
 	const ASSUME_ROLE_CREDENTIALS_EXPIRY_TIME = 43200;
 
-	public static function getCacheCredentialsProvider($roleArn, $s3Region = null)
+	public static function getCacheCredentialsProvider($roleArn, $s3Region = null, $dirnameSuffix = null)
```
yossipapi, line 23: Why not simply set $dirnameSuffix = "" as default And than
$credentialsCacheDir = sys_get_temp_dir() . DIRECTORY_SEPARATOR . 's3_creds_cache_v3' . "_$dirnameSuffix";

A bit more clean

infra/storage/file_transfer_managers/s3Mgr.class.php
```diff
@@ -122,9 +131,12 @@ protected function doLogin($sftp_user, $sftp_pass)
 			return false;
 		}
 
-		if($this->s3Arn && (!isset($sftp_user) || !$sftp_user) && (!isset($sftp_pass) || !$sftp_pass))
+		if($this->s3Arn && $this->s3Region && (!isset($sftp_user) || !$sftp_user) && (!isset($sftp_pass) || !$sftp_pass))
 		{
-			KalturaLog::debug('Found env VAR from config- ' . $this->s3Arn);
+			$s3Arn = $this->s3Arn;
```
yossipapi, line 136: Why do you define these local params? if its only for the log simply use the values stored on $this

plugins/s3DropFolderPlugin/lib/model/S3DropFolderFile.php
```diff
@@ -19,6 +19,7 @@ public function getNameForParsing ()
 
 	public function getFileUrl()
 	{
+		KalturaLog::debug("Generating drop folder file URL");
```
yossipapi, line 22: Do you need this log or was it for debugging?

infra/storage/file_transfer_managers/s3Mgr.class.php
```diff
@@ -79,14 +80,22 @@ protected function __construct(array $options = null)
...
 			$this->storageClass = $options['storageClass'];
 		}
-
-		if (class_exists('KBatchBase'))
+		
+		if($options && isset($options['s3Arn']))
 		{
-			$this->s3Arn = kBatchUtils::getKconfParam('arnRole', true);
+			$this->s3Arn = $options['s3Arn'];
+			$this->hash = true;
```
yossipapi, line 87: Why no simply save the hash here instead of true? 
Also why is this the only case where we hash, what about the other case? I think you should hash there as well to be able to distinguish between creds used by the file transfer manager and the sharedFileSystemWrtappers.

infra/storage/file_transfer_managers/s3Mgr.class.php
```diff
@@ -122,9 +131,12 @@ protected function doLogin($sftp_user, $sftp_pass)
 			return false;
 		}
 
-		if($this->s3Arn && (!isset($sftp_user) || !$sftp_user) && (!isset($sftp_pass) || !$sftp_pass))
+		if($this->s3Arn && $this->s3Region && (!isset($sftp_user) || !$sftp_user) && (!isset($sftp_pass) || !$sftp_pass))
```
yossipapi, line 134: Why did region become mandatory?   it can be null when calling getCacheCredentialsProvider

yossipapi (approved):

plugins/drop_folder/batch/Engine/KDropFolderFileTransferEngine.php
```diff
@@ -280,7 +280,19 @@ public static function getFileTransferManager(KalturaDropFolder $dropFolder)
...
 		{
 			$engineOptions['s3Region'] = $dropFolder->s3Region;
+
+			if ($dropFolder->useS3Arn)
+			{
+				$engineOptions['s3Arn'] = $dropFolder->s3Arn;
+				
+				if (empty($engineOptions['s3Arn']))
+				{
+					$msg = "Drop Folder ID [{$dropFolder->id}] enabled 'Bucket Policy Allow Access' but 's3Arn' value under 's3_drop_folder' in 'runtime_config' map is missing";
```
yossipapi, line 290: Why no fail it in the step of the drop folder creation? Would be much more clean.

plugins/s3DropFolderPlugin/lib/model/S3DropFolderFile.php
```diff
@@ -21,6 +21,14 @@ public function getFileUrl()
 	{
 		$dbDropFolder = DropFolderPeer::retrieveByPK($this->getDropFolderId());
 		$s3Options = $dbDropFolder->getDropFolderParams();
+		
+		if ($s3Options['useS3Arn'] && empty($s3Options['s3Arn']))
```
yossipapi, line 25: Same comment as above:
Why no fail it in the step of the drop folder creation? Would be much more clean.
In validate for insert

yossipapi (approved):

PR #13119, opened 2025-02-04
SUP-44096 Use KS user name in email report instead of admin name

bw-kaltura:


api_v3/services/ReportService.php
```diff
@@ -440,21 +440,13 @@ public function getCsvFromStringParamsAction($id, $params = null, $excludedField
...
-			}
-			else
-			{
-				$partnerId = kCurrentContext::getCurrentPartnerId();
-				$partner = PartnerPeer::retrieveByPK($partnerId);
-				$params->recipientEmail = $partner->getAdminEmail();
-			}
+			$params->recipientEmail = $kuser ? $kuser->getEmail() : $partner->getAdminEmail();
 		}
+		$params->recipientName = $kuser ? $kuser->getFullName() : $partner->getName();
```
MosheMaorKaltura, line 449: If you already exported this param to be set in the api call, then why are you overriding it? why not check also if it was set, do not override it (like in the email)

bw-kaltura: The api doesn't have a field for name, only for email.

bw-kaltura: updated the code.

MosheMaorKaltura (approved):

PR #13150, opened 2025-02-23
PLAT-25152 list past drop folder files by id instead of createdAt

salma2222:


plugins/drop_folder/batch/Engine/KDropFolderEngine.php
```diff
@@ -46,22 +46,67 @@ abstract public function processFolder (KalturaBatchJob $job, KalturaDropFolderC
...
+		$fromCreatedAt = time() - $timeRange;
+		$dropFolderFileId = $this->getSingleDropFolderFileIdByCreatedAt($fromCreatedAt);
+		if(!$dropFolderFileId)
+		{
+			return;
+		}
+
+		do
+		{
+			$dropFolderFiles = $this->loadDropFolderFilesByPage($pager, $dropFolderFileId);
```
drorsou, line 100: Need to advance the pager
$pager->pageIndex++;

plugins/drop_folder/batch/Engine/KDropFolderEngine.php
```diff
@@ -46,22 +46,67 @@ abstract public function processFolder (KalturaBatchJob $job, KalturaDropFolderC
...
+			foreach ($dropFolderFiles as $dropFolderFile)
+			{
+				$this->handleExistingDropFolderFile($dropFolderFile);
```
drorsou, line 103: It's helpful to keep the debug print, we can look for missing files which aren't downloaded
KalturaLog::info("Handle drop folder file: {$dropFolderFile->fileName}");

plugins/drop_folder/batch/Engine/KDropFolderEngine.php
salma2222: no need, we advance the id

danbar0kal (approved):

plugins/drop_folder/batch/Engine/KDropFolderEngine.php
salma2222: added

PR #13155, opened 2025-02-26
SUP-46575 add OTP field and logic to google auth form to appear under…

noahsol30:
… youtube distributions


noahsol30:
As mentioned on https://kaltura.atlassian.net/browse/SUP-46575
When account is setup with OTP and we access youtube distributions -> then access link under "Authorize API Access" -> user tries to authenticate they would get "O
TP is missing" without a way to fill out OTP.

After change
Screenshot 2025-02-26 at 17 36 58

Field validation
Screenshot 2025-02-26 at 17 37 11

Error message for when OTP isn't correct (code: INVALID_OTP, message: OTP provided failed to validate)
Same as when OTP isn't correct in KMC.
Screenshot 2025-02-26 at 17 37 29

Wasn't able to setup OTP with app on local so was hoping to test it out in STG if this is approved.
Thanks

MosheMaorKaltura (approved):

PR #13159, opened 2025-02-28
PLAT-25146: Add applicative groups with no users limit

drorsou:


MosheMaorKaltura (approved):

api_v3/lib/types/filters/KalturaGroupUserFilter.php
```diff
@@ -42,7 +45,7 @@ public function getListResponse(KalturaFilterPager $pager, KalturaDetachedRespon
 			$c = new Criteria();
 			$c->add(kuserPeer::PARTNER_ID, $partnerId);
 			$c->add(kuserPeer::PUSER_ID, $this->groupIdEqual);
-			$c->add(kuserPeer::TYPE, KuserType::GROUP);
+			$c->add(kuserPeer::TYPE, $this->groupType);
```
yossipapi, line 48: Shouldn't we default to GROUP in case the filter was not provided?

api_v3/services/GroupUserService.php
```diff
@@ -431,21 +434,52 @@ protected function validateKuserkGroupCoExistence(kuser $kGroup, $kuserId)
...
-		if (KuserKgroupPeer::doCount($criteria) >= $maxGroupsPerUser)
+		return GroupUserService::DEFAULT_MAX_NUMBER_OF_GROUPS_PER_USER;
+	}
+
+	protected function retrieveActiveGroupsForKuser($kuserId)
+	{
+		$criteria = new Criteria();
+		$criteria->add(KuserKgroupPeer::KUSER_ID, $kuserId);
+		$criteria->add(KuserKgroupPeer::STATUS, KuserKgroupStatus::ACTIVE);
+		return KuserKgroupPeer::doSelect($criteria);
```
yossipapi, line 469: Did you test a user with a lot of groups? 
As this now loads the entire object to run time, we need to make sure we won't get OOO

yossipapi:
Other than the comment I added 
What about other locations that had specific handling for groups, for example 
https://github.com/kaltura/server/blob/4f7c12f5c15141b49f6afeb4b520e0e298878d53/alpha/apps/kaltura/lib/kObjectDeleteHandler.php#L170

api_v3/services/GroupUserService.php
danbar0kal: how much would you say is a lot?

PR #13168, opened 2025-03-09
PLAT-25167 refactor Reach get vendorCatalogItem objects

salma2222:


MosheMaorKaltura (approved):
we must add a description - how to add new catalog item, in confluence and also in the description of this item

PR #13191, opened 2025-03-17
Login Failure attempts in Time Frame send email

danbar0kal:


alpha/lib/model/UserLoginDataPeer.php
```diff
@@ -737,13 +737,45 @@ public static function loginAttemptsLogic($loginData)
...
+					$loginData->save();
+					KalturaLog::notice('User login blocked for login data id ['.$loginData->getId().']');
+					throw new kUserException('', kUserException::LOGIN_RETRIES_EXCEEDED);
+				}
+				else
+				{
+					$loginData->setLoginAttempts(0);
+					$loginData->setFirstLoginFailTime(time());
+					$loginData->save();
+					throw new kUserException('', kUserException::WRONG_PASSWORD);
```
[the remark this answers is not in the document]
danbar0kal: this use case is for when the count of failure reached, but outside of the timeframe, so it doesnt count

MosheMaorKaltura (approved):

alpha/lib/model/UserLoginDataPeer.php
```diff
@@ -737,19 +737,53 @@ public static function loginAttemptsLogic($loginData)
...
+			if ($partner && $partner->getLoginFailTimeframe()) //timeframe is set for this partner
+			{
+				if (time() - $loginData->getFirstLoginFailTime() <= $partner->getLoginFailTimeframe()) //timeframe is not over yet
+				{
+					self::setUserBlockedData($loginData);
+					KalturaLog::notice('User login blocked for login data id ['.$loginData->getId().']');
+					throw new kUserException('', kUserException::LOGIN_RETRIES_EXCEEDED);
+				}
+				else
+				{//timeframe is over, we need to reset the login attempts
```
yossipapi, line 753: We are in a case where we failed to log in but we don't call 		
$loginData->incLoginAttempts();

something in this change seems too cumbersome to me and even less secure than before
IMHO the addition of the login fail time frame should not reset the failed login attempts if the time frame passed but retain a or logic 
i.e did I reach max login attempts in the time frame or reach it regardless. Otherwise, if I have set a max time frame of 1 minute and max attempts to 20 I have 20 chances every minute to guess the password which security-wise is bad.

alpha/lib/model/UserLoginDataPeer.php
```diff
@@ -750,7 +750,7 @@ public static function loginAttemptsLogic($loginData)
 		throw new kUserException('', kUserException::WRONG_PASSWORD);
 	}
 
-	public static function setLastLoginFields($loginData, $kuser)
+		public static function setLastLoginFields($loginData, $kuser)
```
MosheMaorKaltura, line 753: Indentation

MosheMaorKaltura (approved):

PR #13213, opened 2025-03-26
Plat 25210

MosheMaorKaltura:


plugins/admin_console/kaltura_application_links/admin/scripts/plugin/generic-app-admin-action.phtml
```diff
@@ -0,0 +1,24 @@
...
+<style type="text/css">
+    #sub-header { padding: 5px 20px 0px; }
+    #wrapper { padding: 0 0 0 0; width: 100%; }
+</style>
+<div>
+    <h1> Application generic links </h1>
+    <ul id="applicationLinks"></ul>
+</div>
+<script type="text/javascript">
+    var initialized = false;
```
salma2222, line 12: not used

salma2222 (approved):

plugins/admin_console/kaltura_application_links/admin/scripts/plugin/generic-app-admin-action.phtml
salma2222: probably need 

if (initialized) {
            return;
        }
        initialized = true;
        

in case of double clicks and so

PR #13224, opened 2025-04-04
PSVAMB-84354 - new catalog item enabling configurably scheduled live …

hilak:
…captioning


plugins/reach/lib/api/data/vendorCatalogItem/KalturaVendorConfigSchedulingLiveCaptionCatalogItem.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * @package plugins.reach
+ * @subpackage api.objects
+ */
+class KalturaVendorConfigSchedulingLiveCaptionCatalogItem extends KalturaVendorLiveCaptionCatalogItem
```
yossipapi, line 6: I haven't read this entire PR 
But why isn't this a config param on the KalturaVendorLiveCaptionCatalogItem? what is there a need to create an entirely new catalog item type for this?

hilak: @yossipapi  addressed, please review

plugins/schedule/base/lib/model/EntryScheduleEvent.php
```diff
@@ -43,6 +43,14 @@ public function getEntryIds()
 	{
 		return $this->getFromCustomData(self::CUSTOM_DATA_FIELD_ENTRY_IDS);
 	}
+
+	/**
+	 * @return string
```
yossipapi, line 48: Why was this added? I don't see your new code uses it anywhere.

plugins/reach/lib/api/data/vendorCatalogItem/KalturaVendorLiveCaptionCatalogItem.php
```diff
@@ -5,14 +5,40 @@
  */
 class KalturaVendorLiveCaptionCatalogItem extends KalturaVendorLiveCatalogItem
 {
+	/**
+	 * How long before the live stream start should service activate? (in secs)
+	 * @var int
+	 */
+	public $startTimeBuffer;
```
yossipapi, line 12: Is there in your opinion a need to add some upper/lower limits to these values?
I don't see any reason why someone can potentially add a value here that is larger than a few hours
So Maybe we should enforce it system side

yossipapi (changes requested):

yossipapi (approved):

PR #13232, opened 2025-04-14
SUP-47636: Caption tokenize

david-winder-kaltura:


shamamayair (approved):

plugins/content/caption/base/CaptionPlugin.php
```diff
@@ -670,17 +670,21 @@ protected static function getLiveCaptionArray($entryId, $deliveryProfile): array
 		}
 
 		$liveCaptions = array();
+		$baseUrl = $deliveryProfile->getPackagerUrl($liveEntryServerNodes[0]);
+		$url = parse_url($baseUrl, PHP_URL_PATH);
+		$prefix = substr($baseUrl, 0, -strlen($url));
```
MosheMaorKaltura, line 675: getting the prefix out of the request is wrong, assuming that future wise we would send a query param like ?test=1 , this code will break

ImryAshu (approved):

PR #13233, opened 2025-04-15
PLAT-25174 - add default reach profile on partner catalog item

salma2222:


plugins/reach/lib/model/om/BasePartnerCatalogItem.php
```diff
@@ -8,8 +8,9 @@
  * @package plugins.reach
  * @subpackage model.om
  */
-abstract class BasePartnerCatalogItem extends BaseObject  implements Persistent {
-
+abstract class BasePartnerCatalogItem extends BaseObject  implements Persistent
```
danbar0kal, line 11: i think this is an auto generated class that we shouldn't edit

MosheMaorKaltura (approved):

PR #13244, opened 2025-04-20
PLAT-25226: Support api rate limit in ps2 actions

yossipapi:


infra/monitor/KalturaMonitorClient.php
```diff
@@ -805,4 +805,40 @@ public static function sendErrorEvent($errorCode)
...
+			return;
+		}
+		
+		preg_match('#/p/([^/]+)/sp/([^/]+)#', $uri, $matches);
+		if(!$matches || !count($matches))
+		{
+			return;
+		}
+		
+		$p = isset($matches[1]) ? $matches[1] : (isset($matches[2]) ? $matches[2] : null);
```
MosheMaorKaltura, line 823: if $p is for partnerId, I would consider renaming

infra/monitor/KalturaMonitorClient.php
```diff
@@ -805,4 +805,40 @@ public static function sendErrorEvent($errorCode)
...
+		if(!$p)
+		{
```
MosheMaorKaltura, line 825: If partnerId comes from /sp/ - dont we need to divide by 100 to get the correct value?

yossipapi: You're right, fixed.

infra/monitor/KalturaMonitorClient.php
yossipapi: renamed

infra/monitor/KalturaMonitorClient.php
```diff
@@ -805,4 +805,40 @@ public static function sendErrorEvent($errorCode)
...
+		}
+		
+		preg_match('#/p/([^/]+)/sp/([^/]+)#', $uri, $matches);
+		if(!$matches || !count($matches))
+		{
+			return;
+		}
+		
+		$partner_id = isset($matches[1]) ? $matches[1] : (isset($matches[2]) ? $matches[2]/100 : null);
+		if(!$partner_id)
```
MosheMaorKaltura, line 824: Is it possible to get manifest request on partner 0?

MosheMaorKaltura (approved):

infra/monitor/KalturaMonitorClient.php
yossipapi: Should not happen.

PR #13246, opened 2025-04-21
PLAT-25177 add objectType for entry vendor task

salma2222:


plugins/reach/lib/model/EntryVendorTask.php
```diff
@@ -176,7 +182,12 @@ public function getUserId()
 	{
 		return $this->getFromCustomData(self::CUSTOM_DATA_USER_ID, null, null);
 	}
-	
+
+	public function getEntryObjectType()
+	{
+		return $this->getFromCustomData(self::CUSTOM_DATA_ENTRY_OBJECT_TYPE);
```
MosheMaorKaltura, line 188: default value - should be entry

salma2222: done

plugins/reach/lib/kReachManager.php
```diff
@@ -780,60 +772,103 @@ public static function addEntryVendorTask(entry $entry, ReachProfile $reachProfi
...
+				KalturaLog::debug("Entry [{$entry->getId()}] is temporary, entry vendor task object wont be created for it");
+				return false;
+			}
+
+			//Check if static content and the catalog item is excluding static content, if so, dont create the task
+			if(count($vendorCatalogItem->getAdminTagsToExcludeArray()) && array_intersect($vendorCatalogItem->getAdminTagsToExcludeArray(), $entry->getAdminTagsArr()))
+			{
+				KalturaLog::debug("Entry [{$entry->getId()}] has admin tags that are excluded by the catalog item, entry vendor task object wont be created for it");
+				return false;
+			}
```
MosheMaorKaltura, line 817: in case $entryObjectType == KalturaEntryObjectType::ENTRY we should allow, but in other object types we should deny (until new object type is added to the code)

plugins/reach/lib/kReachManager.php
```diff
@@ -780,60 +772,103 @@ public static function addEntryVendorTask(entry $entry, ReachProfile $reachProfi
...
-			$entryVendorTask->setDictionary($dictionary->getData());
 
-		$entryVendorTask->setStatus($status);
-		return $entryVendorTask;
+		return $status;
 	}
 	
 	//For automatic dispatched tasks make sure to set the entry creator user as the entry owner
-	protected static function getTaskKuserId(entry $entry)
+	protected static function getTaskKuserId(entry $entry, $entryObjectType)
```
MosheMaorKaltura, line 854: We should use different approach, since this function is getting only entry objects as input - there is no use for sending entryObjectType

plugins/reach/lib/kReachManager.php
```diff
@@ -780,60 +772,103 @@ public static function addEntryVendorTask(entry $entry, ReachProfile $reachProfi
...
-			$kuserId = $entry->getKuserId();
+			$kuserId = $entryObjectType == KalturaEntryObjectType::ENTRY ? $entry->getKuserId() : null;
 		}
 		
 		return $kuserId;
 	}
 	
 	//For automatic dispatched tasks make sure to set the entry creator user as the entry owner
-	protected static function getTaskPuserId(entry $entry)
+	protected static function getTaskPuserId(entry $entry, $entryObjectType)
```
MosheMaorKaltura, line 866: same here, if input type is entry - whats the point of sending entryObjectType

plugins/reach/lib/model/VendorCatalogItem.php
```diff
@@ -194,8 +194,12 @@ public function calculatePriceForEntry(entry $entry, $taskDuration = null)
 		return call_user_func($this->getPricing()->getPriceFunction(), $durationMsec, $this->getPricing()->getPricePerUnit());
 	}
 	
-	public function getTaskVersion($entryId, $jobData = null)
+	public function getTaskVersion($entryId, $entryObjectType = KalturaEntryObjectType::ENTRY, $jobData = null)
 	{
+		if($entryObjectType != KalturaEntryObjectType::ENTRY)
```
MosheMaorKaltura, line 199: Lets move the old code to be getEntryTaskVersion, and the new function will decide which function to call base on the type, if its entry call getEntryTaskVersion(), otherwise do nothing (for now)

plugins/reach/lib/model/data/VendorAlignmentCatalogItem.php
```diff
@@ -20,18 +20,22 @@ public function applyDefaultValues()
 		$this->setServiceFeature(VendorServiceFeature::ALIGNMENT);
 	}
 	
-	public function getTaskVersion($entryId, $jobData = null)
+	public function getTaskVersion($entryId, $entryObjectType = KalturaEntryObjectType::ENTRY, $jobData = null)
```
MosheMaorKaltura, line 23: Also here this is a dedicated to entry function, lets split

MosheMaorKaltura (approved):

PR #13253, opened 2025-04-23
Fix apcu_fetch call in kApcWrapper

navarro967:
Correct the argument passed to the `apcu_fetch` function to ensure it properly returns the success status.


navarro967:
@yossipapi
@amiras89

PR #13259, opened 2025-04-27
plat-25172: Specify which catalogItem requires profile with overages

salma2222:


plugins/reach/lib/model/VendorCatalogItem.php
```diff
@@ -278,6 +279,16 @@ public function getContract()
...
 	}
 
+	public function setRequiresOverages($v)
+	{
+		$this->putInCustomData(self::CUSTOM_ADMIN_REQUIRES_OVERAGES, $v);
+	}
+
+	public function getRequiresOverages()
+	{
+		return $this->getFromCustomData(self::CUSTOM_ADMIN_REQUIRES_OVERAGES);
```
MosheMaorKaltura, line 289: set default value explicit to false (rather then default null) as it is boolean var

plugins/reach/lib/model/ReachProfile.php
```diff
@@ -564,4 +564,11 @@ public function getCacheInvalidationKeys()
 	{
 		return array("reachProfile:id=".strtolower($this->getId()));
 	}
+
+	public function getAllowsOverages()
+	{
+		// TODO - implement
```
MosheMaorKaltura, line 570: ???

plugins/reach/lib/model/VendorCatalogItem.php
salma2222: done

MosheMaorKaltura (approved):

PR #13278, opened 2025-05-05
SUP-47255 fix getPlaybackContext to return child entry with child fla…

Tova-a:
…vorIDs

### DIFF
--- a/alpha/apps/kaltura/lib/kContextDataHelper.php
+++ b/alpha/apps/kaltura/lib/kContextDataHelper.php
@@ -73,8 +73,8 @@ class kContextDataHelper
 	 *
 	 * @var string
 	 */
-	private $entryId;
-	
+	private $childEntryId;
+
 	private $storageProfilesXML = null;
 	
 	private $streamerType = null;
@@ -85,15 +85,15 @@ class kContextDataHelper
 	 * 
 	 * @param entry $entry
 	 * @param Partner $partner
-	 * @param string $entryId
 	 * @param asset $asset
+	 * @param string $childEntryId
 	 */
-	public function __construct(entry $entry, Partner $partner, string $entryId, asset $asset = null)
+	public function __construct(entry $entry, Partner $partner, asset $asset = null ,string $childEntryId = null)
 	{
 		$this->entry = $entry;
 		$this->partner = $partner;
-		$this->entryId = $entryId;
 		$this->asset = $asset;
+		$this->childEntryId = $childEntryId;
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


alpha/apps/kaltura/lib/kContextDataHelper.php
```diff
@@ -56,7 +56,13 @@ class kContextDataHelper
 	 * @var entry
 	 */
 	private $entry;
-	
+
+	/**
+	 *
+	 * @var entry
+	 */
+	private $originalEntry;
```
MosheMaorKaltura, line 64: No need to hold entire entry object as we are only interested in the entryId

MosheMaorKaltura (approved):

alpha/apps/kaltura/lib/kContextDataHelper.php
```diff
@@ -74,18 +74,19 @@ class kContextDataHelper
...
 	private $mediaProtocol = null;
-	
+
 	/**
 	 * 
 	 * @param entry $entry
 	 * @param Partner $partner
 	 * @param asset $asset
 	 */
-	public function __construct(entry $entry, Partner $partner, asset $asset = null)
```
yossipapi, line 84: Change the order so optional params will always be last

yossipapi (changes requested):

alpha/apps/kaltura/lib/kContextDataHelper.php
yossipapi, line 84: Why do you need the entry id if you already have the entry?

alpha/apps/kaltura/lib/kContextDataHelper.php
```diff
@@ -268,9 +269,9 @@ private function setContextDataFlavorAssets($flavorTags)
 		if (is_null($this->asset))
 		{
 			if($flavorParamsIds && count($flavorParamsIds))
-				$flavorAssets = assetPeer::retrieveReadyByEntryIdAndFlavorParams($mediaEntryId, $flavorParamsIds, $flavorParamsNotIn);
```
yossipapi, line 271: Based on what I see in the code, I am not sure this change is correct in the case of playlist

api_v3/services/BaseEntryService.php
```diff
@@ -1025,7 +1025,7 @@ function getPlaybackContextAction($entryId, KalturaPlaybackContextOptions $conte
 				throw new KalturaAPIException(KalturaErrors::FLAVOR_ASSET_ID_NOT_FOUND, $contextDataParams->flavorAssetId);
 		}
 
-		$contextDataHelper = new kContextDataHelper($dbEntry, $this->getPartner(), $asset);
```
yossipapi, line 1028: Now that you changed the order in the ctor this is not correct.
Also, please note that this class is initialized from other location as well
So if you add a mandatory param you need to add it in all places.

api_v3/services/BaseEntryService.php
```diff
@@ -999,10 +999,12 @@ function getPlaybackContextAction($entryId, KalturaPlaybackContextOptions $conte
 			kApiCache::setExpiry(60);
 		}
 
+		$childEntry = null;
```
MosheMaorKaltura, line 1002: rename, childEntry to childEntryId

Tova-a: Done.

MosheMaorKaltura (approved):

PR #13280, opened 2025-05-05
Ursa 21.15.0 plat 25229

danbar0kal:


plugins/search/providers/elastic_search/lib/api/types/KalturaESearchParams.php
```diff
@@ -46,6 +52,11 @@ public function toObject($object_to_fill = null, $props_to_skip = array())
 			$object_to_fill = new ESearchParams();
 		}
 
+		if ($this->objectId && $this->objectIds)
```
drorsou, line 55:
```suggestion
		if (isset($this->objectId) && isset($this->objectIds))
```
Better to use isset() because it should throw an exception on an empty string as well

drorsou (approved):

MosheMaorKaltura (approved):

PR #13299, opened 2025-05-13
add "opensearch" distribution config to force ES7 syntax

amiras89:


plugins/search/providers/elastic_search/scripts/populateElasticFromLog.php
```diff
@@ -43,12 +43,25 @@ function filter($i)
 		exit(-1);
 	}
 	$config = parse_ini_file($configFile);
+	KalturaLog::debug("Configuration file [$configFile] loaded successfully - values: " . print_r($config, true));
 }
 
 $elasticCluster = $config['elasticCluster'];
 $elasticServer = $config['elasticServer'];
 $elasticPort = (isset($config['elasticPort']) ? $config['elasticPort'] : 9200);
-$elasticVersion = isset($config['elasticVersion']) ? $config['elasticVersion'] : elasticClient::ELASTIC_MAJOR_VERSION_5;
```
yossipapi, line 51: I think in this case instead of duplicating the code maybe it would be better to simply not send the version here

PR #13312, opened 2025-05-19
entry search notIn

danbar0kal:


plugins/search/providers/elastic_search/lib/model/ESearchParams.php
```diff
@@ -98,6 +105,22 @@ public function setObjectIds($objectIds)
 		$this->objectIds = $objectIds;
 	}
 
+	/**
+	 * @return string
```
MosheMaorKaltura, line 109: Fix annotation as it returns boolean

plugins/search/providers/elastic_search/lib/model/ESearchParams.php
```diff
@@ -98,6 +105,22 @@ public function setObjectIds($objectIds)
...
+	 */
+	public function getObjectIdsNotIn()
+	{
+		return $this->objectIdsNotIn;
+	}
+
+	/**
+	 * @param string $objectIdsNotIn
```
MosheMaorKaltura, line 117: same here

plugins/search/providers/elastic_search/lib/model/search/ESearchQueryAttributes.php
```diff
@@ -113,6 +118,22 @@ public function setObjectId($objectId)
 		$this->objectId = $objectId;
 	}
 
+	/**
+	 * @return KalturaNullableBoolean
```
MosheMaorKaltura, line 122: cannot us this type here as you are in core and this type comes from the higher level of the API

MosheMaorKaltura (approved):

PR #13316, opened 2025-05-20
SUP-47031 verify session is not widget session to not play entry in e…

noahsol30:
…mbed


noahsol30:
When embedding an entry from KMS, with embed code that's generated in KMS, when the entry is out of schedule - the user is still able to play the entry. As the KS has sview privilege it sets it as scheduled now and is playable. But KS in such case also has widgets so we check if it is widget session or not.

PR #13321, opened 2025-05-20
SUP-48097 include updated type notation for array as generated from c…

noahsol30:
…lient generator


noahsol30:
Previously the client generator would generate and array with the @var annotation of "array of KalturaIntegerValue".
With recent changes in the client generator (https://github.com/kaltura/clients-generator/pull/994) it's now changed to "@var Kaltura_Client_Type_IntegerValue[]"
This led to an element not to appear in the media repurposing form.
In this code the element is created when we open the form. Adding supporting for older value just in case and added the new notation for array as well with the brackets.

MosheMaorKaltura (approved):

PR #13328, opened 2025-05-25
allowing to set entitlement with a list of ids

danbar0kal:


alpha/lib/model/categoryEntryPeer.php
```diff
@@ -195,6 +195,17 @@ public static function selectByEntryId($entryId)
 		
 		return self::doSelect($c);
 	}
+
+	public static function selectByEntryIds($entryIds)
```
MosheMaorKaltura, line 199: force type to be array, reduce validations

plugins/search/providers/elastic_search/lib/entitlement/kEntryElasticEntitlement.php
```diff
@@ -118,7 +118,18 @@ public static function setFilteredCategoryIds(ESearchOperator $eSearchOperator,
 
         $searchItems = $eSearchOperator->getSearchItems();
         $filteredCategoryIds = array();
-        $filteredEntryId = $objectId ? array($objectId) : array();
+	    $filteredEntryId = array();
+	    if ($objectId)
+	    {
+		    if(strpos($objectId, ',') !== false)
+		    {
+			    $filteredEntryId = explode(',', $objectId);
```
MosheMaorKaltura, line 126: this is the only line needed

alpha/lib/model/categoryEntryPeer.php
```diff
@@ -195,6 +195,15 @@ public static function selectByEntryId($entryId)
 		
 		return self::doSelect($c);
 	}
+
+	public static function selectByEntryIds(array $entryIds, $privacyContext)
+	{
+		$c = new Criteria();
+		$c->add(self::ENTRY_ID, $entryIds, Criteria::IN);
+		$c->add(self::PRIVACY_CONTEXT, $privacyContext, Criteria::EQUAL);
```
drorsou, line 203: If we don't have privacy_context, the query searches only for entries without.

danbar0kal: yes, but nothing to do about it. this is how it worked before with on entryId. KMS will always use privacy context in any case

plugins/search/providers/elastic_search/lib/entitlement/kEntryElasticEntitlement.php
```diff
@@ -111,20 +111,21 @@ private static function initializeUserCategoryEntryEntitlement($ks)
...
+    public static function setFilteredCategoryIds(ESearchOperator $eSearchOperator, $objectIdsCsvStr, $objectIdsNotIn = null)
     {
-        if($eSearchOperator->getOperator() != ESearchOperatorType::AND_OP)
+        if(($eSearchOperator->getOperator() != ESearchOperatorType::AND_OP) || $objectIdsNotIn)
             return;
 
         $searchItems = $eSearchOperator->getSearchItems();
         $filteredCategoryIds = array();
-        $filteredEntryId = $objectId ? array($objectId) : array();
+	    $filteredEntryIds = $objectIdsCsvStr ? explode(',', $objectIdsCsvStr) : array();
```
MosheMaorKaltura, line 121: explode returns empty array for nothing

alpha/lib/model/categoryEntryPeer.php
MosheMaorKaltura, line 203: add this only if privacy_context contains valid value

MosheMaorKaltura (approved):

PR #13334, opened 2025-05-27
PLAT-25308: Migration script to Firebase V2

drorsou:


alpha/scripts/utils/migrateFirebaseToV2.php
```diff
@@ -0,0 +1,114 @@
...
+
+                $template->setStatus(EventNotificationTemplateStatus::DELETED);
+				$template->save();
+            }
+        }
+    }
+
+    kMemoryManager::clearMemory();
+
+    $c = new Criteria();
```
MosheMaorKaltura, line 88: This block is not needed, just need to add $c->setOffset($offset);

MosheMaorKaltura (approved):

PR #13339, opened 2025-05-30
PSVAMB-85314 -> isDuplicateTask() statement been change so anyway we …

TalDubovKaltura:
…can remove it


hilak (commented):
@TalDubovKaltura  I'm not sure this is the correct behavior and would suggest consulting with the architect. 
The point of this function is to prevent an endless string of identical vendor tasks from being added to the Kaltura queue, which could cause many issues. The way it's implemented now, the function is preventing nothing.

TalDubovKaltura:
> @TalDubovKaltura I'm not sure this is the correct behavior and would suggest consulting with the architect. The point of this function is to prevent an endless string of identical vendor tasks from being added to the Kaltura queue, which could cause many issues. The way it's implemented now, the function is preventing nothing.

But the function still in use(Im just not overriding it) and the check of active task is still occurring, same as other  vendor catalog items

PR #13345, opened 2025-06-03
PLAT-25324: add $mapScale to cache key

MosheMaorKaltura:


danbar0kal (approved):

alpha/apps/kaltura/lib/myEntryUtils.class.php
```diff
@@ -2263,7 +2263,7 @@ public static function getVolumeMapContent($flavorAsset, $mapScale = null, $dura
 		$content = null;
 
 		$cacheStore = kCacheManager::getSingleLayerCache(kCacheManager::CACHE_TYPE_VOLUME_MAP);
-		$cacheKey = $entryId . '_volumeMap';
+		$cacheKey = $entryId . '_volumeMap' . $mapScale;
```
yossipapi, line 2266: I think you should also include the version of the asset in your cache key.
In case someone replaces content based on what I see the code will return the old volume map.

yossipapi, line 2266: $cacheKey = $entryId . '_volumeMap' . $mapScale;
to
$cacheKey = $entryId . '_volumeMap**_**' . $mapScale;

PR #13371, opened 2025-06-16
PSVAMB-85314 -> Support SignLanguage ratio display in the partner cat…

TalDubovKaltura:
…alog item list also

### DIFF
--- a/plugins/reach/admin/scripts/plugin/partner-catalog-item-list-action.phtml
+++ b/plugins/reach/admin/scripts/plugin/partner-catalog-item-list-action.phtml
@@ -59,8 +59,16 @@
 				else
 				    echo null;?>
 				</td>
-				<td><?php if (isset($catalogItem->outputFormat))
-				    echo $this->enumTranslate('Kaltura_Client_Reach_Enum_VendorCatalogItemOutputFormat', $catalogItem->outputFormat);
+				<td><?php if (isset($catalogItem->outputFormat)) {
+                        if($catalogItem->serviceFeature === Kaltura_Client_Reach_Enum_VendorServiceFeature::SIGN_LANGUAGE)
+                        {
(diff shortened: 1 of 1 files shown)


plugins/reach/admin/scripts/plugin/partner-catalog-item-list-action.phtml
```diff
@@ -59,8 +59,16 @@
 				else
 				    echo null;?>
 				</td>
-				<td><?php if (isset($catalogItem->outputFormat))
-				    echo $this->enumTranslate('Kaltura_Client_Reach_Enum_VendorCatalogItemOutputFormat', $catalogItem->outputFormat);
+				<td><?php if (isset($catalogItem->outputFormat)) {
```
yossipapi, line 62: Why isn't VendorCatalogItemSignLanguageOutputFormat an extension of VendorCatalogItemOutputFormat
This would remove the need if this if and will make much more sense

TalDubovKaltura: But the values for both VendorCatalogItemSignLanguageOutputFormat and VendorCatalogItemOutputFormat are the same, the only difference is the view in the KAC
image
For example in catalog item list:
image

yossipapi: This implements base enum
image

Whereas it could extend this:
image

hilak: @yossipapi  I think it's better to eliminate KalturaVendorSignLanguageFormat, and add the 2 new values to the base enum, since this is part of the base reach capability anyway. Keeping the enum and making it extend the base class risks overlap of enum values in the future.

PR #13389, opened 2025-06-24
VCP-20936: modify playsViewsElasticConsumer to support multi-cluster setup

amiras89:


amiras89:
Was only temporary solution for test multi-cluster transition
Real solution is https://github.com/kaltura/server/pull/13394/files

PR #13443, opened 2025-07-15
NO-VCP: dynamically calculate sphinx index name (increase entry to 20)

amiras89:


amiras89:
added as part of different pr

PR #13893, opened 2026-04-07
PLAT-25757 log for total amount of to cc bcc recipents sent

noahsol30:
## Pull Request Checklist

Please complete the following before submitting:

General notes - 
- [ ] I have tested the changes locally.
- [ ] I have written unit tests where applicable.
- [ ] I have updated documentation where needed.
- [ ] I have added comments to complex code.
- [ ] This PR follows the coding style guidelines.
- [ ] I have updated release notes with new feature

New Kaltura Types
- [ ] I have created new clients
- [ ] I have notified related apps - KMCNG / KMS / EP .... about new clients

New Kaltura Services / Actions
- [ ] I have added a deployment script

### Questions

1. What is the purpose of this PR?
   - _Enter your answer here_

2. Does this change affect production code or infrastructure?
   - [ ] Yes
   - [ ] No

3. If yes, what is the rollback plan?
   - _Enter your answer here_

### DIFF
--- a/plugins/event_notification/providers/email/lib/batch/KDispatchEmailNotificationEngine.class.php
+++ b/plugins/event_notification/providers/email/lib/batch/KDispatchEmailNotificationEngine.class.php
@@ -159,7 +159,8 @@ class KDispatchEmailNotificationEngine extends KDispatchEventNotificationEngine
 			$this::$mailer->MessageID = $data->messageID;
 
 		$contentParameters = $this->getContentParameters($data);
-			
+		$validToCount = 0;
+		$validCcCount = 0;
 		if($data->to)
 		{
 			$recipients = $this->getRecipientArray($data->to, $contentParameters);
@@ -170,6 +171,7 @@ class KDispatchEmailNotificationEngine extends KDispatchEventNotificationEngine
 					continue;
 				}
 				KalturaLog::info("Adding recipient to TO recipients $name<$email>");
+				$validToCount++;
 				self::$mailer->AddAddress($email, $name);
 			}
 		}
@@ -184,6 +186,7 @@ class KDispatchEmailNotificationEngine extends KDispatchEventNotificationEngine
 					continue;
 				}
 				KalturaLog::info("Adding recipient to CC recipients $name<$email>");
+				$validCcCount++;
 				self::$mailer->AddCC($email, $name);
 			}
 		}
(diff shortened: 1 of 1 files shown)


noahsol30:
There seems to be an issue here that we send to "to" and "cc" again if "bcc" is sent in bulks. So the logging for "to" and "cc" I can alter so it's printed only once. But avoided it to check if we want to possibly fix the logic itself instead.
wdyt?

plugins/event_notification/providers/email/lib/batch/KDispatchEmailNotificationEngine.class.php
```diff
@@ -257,11 +257,28 @@ protected function sendEmail(KalturaEmailNotificationTemplate $emailNotification
 
 		$recipientsBccHandledCounter = 0;
 		$recipientsBccBulk = 500;
+		$recipientsTo = $data->to ? $this->getRecipientArray($data->to, $contentParameters) : null;
```
inbal-ben-david, line 260: why doing such a heavy operation like getRecipientArray here when you either way did it in an earlier phase and can just get the number of items in the array?

plugins/event_notification/providers/email/lib/batch/KDispatchEmailNotificationEngine.class.php
```diff
@@ -257,11 +257,28 @@ protected function sendEmail(KalturaEmailNotificationTemplate $emailNotification
...
+		$recipientsCc = $data->cc ? $this->getRecipientArray($data->cc, $contentParameters) : null;
+
+		if($recipientsTo)
```
inbal-ben-david, line 263: why having redundent log prints?
just have one log with something like number of recipients: to[] cc[] bcc

plugins/event_notification/providers/email/lib/batch/KDispatchEmailNotificationEngine.class.php
```diff
@@ -277,6 +294,9 @@ protected function sendEmail(KalturaEmailNotificationTemplate $emailNotification
 		    try 
 		    {
 				KalturaLog::info('Sending Bulk');
+				$batchNum = ceil($recipientsBccHandledCounter / $recipientsBccBulk);
```
inbal-ben-david, line 297: i think you should have +1 to start the index at 1

plugins/event_notification/providers/email/lib/batch/KDispatchEmailNotificationEngine.class.php
```diff
@@ -250,18 +253,27 @@ protected function sendEmail(KalturaEmailNotificationTemplate $emailNotification
 		
 		$recipientsBcc = array();
 
+		$validBccCount = 0;
```
noahsol30, line 256: added valid counts for to, cc and bcc as we filter out invalid emails. so the log line will only print valid emails that are sent as they are the ones added to mailer

plugins/event_notification/providers/email/lib/batch/KDispatchEmailNotificationEngine.class.php
```diff
@@ -277,6 +289,12 @@ protected function sendEmail(KalturaEmailNotificationTemplate $emailNotification
 		    try 
 		    {
 				KalturaLog::info('Sending Bulk');
+
+				if($recipientsBcc)
+				{
+					$batchNum = ceil($recipientsBccHandledCounter / $recipientsBccBulk);
```
noahsol30, line 295: no need to change logic here as it will only be applied in case bcc got values. then it would numerate properly 1, 2 etc

